PR #1, opened 2015-07-06
ECC Sign and Verify tests, at crypto_ops level

pascal-brand-st-dev:


pascal-brand-st-dev:
Used to test https://github.com/OP-TEE/optee_os/pull/371

PR #24, opened 2015-09-04
GP11 : ECDSA NIST test vectors

cedric-chaumont-st-dev:
Signed-off-by: Cedric Chaumont <cedric.chaumont@st.com>

PR #26, opened 2015-09-07
GP11 : DSA-SHA224 DSA-SHA256 algorithms

cedric-chaumont-st-dev:


host/xtest/Makefile
```diff
@@ -10,6 +10,10 @@ endif
 # be used instead.
 OPTEE_CLIENT_EXPORT ?= ../../../optee_client/out/export
 
+# If CFG_XTEST_LITE=n : compile full xtest suite (default)
+# If CFG_XTEST_LITE=y : compile sanity xtest suite
+CFG_XTEST_LITE ?= n
```
jenswi-linaro: Runtime configuration would be more useful.

cedric-chaumont-st-dev: Do you mean something like CFG_RUNTIME_LITE ?

cedric-chaumont-st-dev: Do you mean something like "xtest lite" ?  Could it be done in another PR ?

jforissier: +1 with @jenswi-linaro. IMO compile-time options are inconvenient, so they should not be used unless there is a reason (like reducing file size, or incompatible options...).
It is just a matter of replacing `#ifndef ... #endif` with `if (...) { ... }`, no?

pascal-brand-st-dev: +1 with @jenswi-linaro 
But this is not just a `if...` as the current `ifndef` is in the definition of the testcases to take into account. One solution is to add a boolen in `XTEST_AC_DSA_CASE` to indicate if the test case should be run in the light suite, and then add the `if` (same for #24)

So I would vote to have it in another PR (soon hopefully?) and to remove `CFG_XTEST_LITE` from the current PR

jforissier: It can be a simple `if` I think: what about adding  global variable (`option_light` for instance) that would be set when the command line is parsed, and could be tested from anywhere? Personally I don't like the idea of adding things when we know we want to remove them ASAP.

jenswi-linaro: I agree with Pascal, let's do the "lite" stuff in a separate PR.

cedric-chaumont-st-dev:
Updated. Squashed. Verified.
37980 subtests of which 1 failed
687 test cases of which 1 failed
0 test case was skipped
TEE test application done!

NOTE: GP test 7671 failing (patch on going)

pascal-brand-st-dev:

cedric-chaumont-st-dev:
Update git comment.

pascal-brand-st:
Merged manually

PR #29, opened 2015-09-10
ta os_test: increase size beyond 1 MB

jenswi-linaro:
Increase size of os_test TA beyond 1 MB of code and data.

Tests that the fix for a large va space for TAs still works.


jforissier:
You should move the S-o-b at the end of the commit message.

`Tested-by: Jerome Forissier <jerome.forissier@linaro.org> (HiKey 32-bit)`

For the record... with this patch, `xtest 1006` passes on the latest optee_os master (commit 07a169c) but fails if I revert optee_os commit e852b51 ("core: correct TA virtual address mapping calculation").

jenswi-linaro:
Moved S-o-b line, applied tags.

PR #31, opened 2015-09-14
Negative tests

jenswi-linaro:
Adds two negative test to 1008


jenswi-linaro:
Replaced by https://github.com/OP-TEE/optee_test/pull/46

PR #35, opened 2015-09-18
Cleanup the way GP tests are added

pascal-brand-st-dev:
- scripts to enable the GP tests are replaced by the Makefile
- Update documentation


jforissier:
Two minor comments:
- It would be nice to avoid patching `xtest_7000.c` which is then reported as modified by `git status`. Since we are basically just adding tests, it should be possible to add a new file to the build.
- We could get rid of `CFG_GP_TESTSUITE_ENABLE` and keep `CFG_GP_PACKAGE_PATH` only (if defined, it means we want to run the GP tests).

This can be done separately, I can propose patches if you wish. Anyway:

`Tested-by: Jerome Forissier <jerome.forissier@linaro.org (HiKey)`

PR #38, opened 2015-09-24
Fix openssl copy in case of 32bits distro

pascal-brand-st-dev:


jforissier:
I don't like the fact that we are linking against a SSL library that is in the git repo (`libcrypto.a`), but using headers that come from the host platform. But since this is not the point of this PR: 

cedric-chaumont-st-dev:
Reviewed-by: Cedric Chaumont <cedric.chaumont@linaro.org>

cedric-chaumont-st-dev:
Tested-by: Cedric Chaumont <cedric.chaumont@linaro.org> (STM Boards)

PR #42, opened 2015-09-28
Add xtest filtering level (xtest lite)

cedric-chaumont-st-dev:
RSA/DSA/ECDSA/ECDH
default: sanity test suite (level=0)
usage: xtest -h

Signed-off-by: Cedric Chaumont <cedric.chaumont@st.com>


cedric-chaumont-st-dev:
@pascal-brand-st-dev : sanity/full should be taken into account because of 'goto run;'

pascal-brand-st-dev:
@cedric-chaumont-st-dev You're right, I missed it. However, I do not find it clean in the way "if -t option is there, then -l option is forgiven". I would drop sanity/full, and only keep -l option. @jforissier what do you think?

jforissier:
@pascal-brand-st-dev I agree. We should add `-t` later if/when we integrate benchmark (performance) tests, in which case `-t` could combine nicely with `-l`. But I don't think it is a good idea to use `-t something` as a synonym for `-l x`.

cedric-chaumont-st-dev:
okidoki.

cedric-chaumont-st-dev:
Update.

pascal-brand-st-dev:

jforissier:

cedric-chaumont-st-dev:
Rebase. Squash. Verify (STM, AR Juno)
44893 subtests of which 0 failed
688 test cases of which 0 failed
0 test case was skipped
TEE test application done!

PR #44, opened 2015-10-13
Fix clean in case TA_DEV_KIT_DIR is wrongly defined

pascal-brand-st-dev:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ else
 	out-dir := $(CURDIR)/out
 endif
 
-include $(TA_DEV_KIT_DIR)/host_include/conf.mk
+-include $(TA_DEV_KIT_DIR)/host_include/conf.mk
 
 ifneq ($V,1)
 	q := @
@@ -15,7 +15,12 @@ else
 endif
 
 .PHONY: all
+ifneq ($(wildcard $(TA_DEV_KIT_DIR)/host_include/conf.mk),)
 all: xtest ta
+else
+all:
+	$(q)echo "TA_DEV_KIT_DIR is not correctly defined" && false
+endif
 
 .PHONY: xtest
 xtest:
@@ -32,9 +37,15 @@ ta:
 			  $@
 
 .PHONY: clean
+ifneq ($(wildcard $(TA_DEV_KIT_DIR)/host_include/conf.mk),)
 clean:
 	$(q)$(MAKE) -C host/xtest O=$(out-dir)/xtest q=$(q) $@
 	$(q)$(MAKE) -C ta O=$(out-dir)/ta q=$(q) $@
+else
+clean:
+	$(q)echo "TA_DEV_KIT_DIR is not correctly defined"
+	$(q)echo "You can remove manually $(out-dir)"
+endif
 
 .PHONY: patch
 patch:
@@ -78,6 +89,7 @@ endef
 
 # openssl .h file installation
 forgpdir=${CURDIR}/host/xtest/for_gp
+.PHONY: patch-openssl
 patch-openssl:
 	$(q)mkdir -p ${forgpdir}/include/openssl ${forgpdir}/lib
 	$(q)if [ -d /usr/include/x86_64-linux-gnu/openssl ]; then \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -6,7 +6,9 @@ else
 	out-dir := $(CURDIR)/out
 endif
 
+ifneq ($(wildcard $(TA_DEV_KIT_DIR)/host_include/conf.mk),)
 include $(TA_DEV_KIT_DIR)/host_include/conf.mk
```
jenswi-linaro: Replacing these three lines with
`-include $(TA_DEV_KIT_DIR)/host_include/conf.mk`
would give the same result

Makefile
```diff
@@ -32,9 +39,15 @@ ta:
...
 
 .PHONY: clean
+ifneq ($(wildcard $(TA_DEV_KIT_DIR)/host_include/conf.mk),)
 clean:
 	$(q)$(MAKE) -C host/xtest O=$(out-dir)/xtest q=$(q) $@
 	$(q)$(MAKE) -C ta O=$(out-dir)/ta q=$(q) $@
+else
+clean:
+	$(q)echo "TA_DEV_KIT_DIR is not correctly defined - Remove $(out-dir)"
+	$(q)rm -rf $(out-dir)
```
jenswi-linaro: What if `$(out-dir)` has been specified as `/` by misstake?

Makefile
pascal-brand-st-dev: I'll fix

Makefile
pascal-brand-st-dev: Right, it's too dangerous.
I'll check if the clean can work when the dev_kit is not there (it may be possible). Or if not possible, we'll just have an echo || false, as in the `all:` target

jforissier: Would it be OK to do `rm -rf $(out-dir)/ta $(out-dir)/xtest` maybe?

pascal-brand-st-dev: It is not possible to clean without the devkit.  So either "nothing" or Jerome's proposal is fine with me.

jenswi-linaro: Why do we want to do a `rm -rf`?

pascal-brand-st-dev: what would you prefer?

jenswi-linaro: Not using `rm -rf`, as a rule of thumb never use `rm -rf`.

jforissier: @jenswi-linaro wise words ;)
I'm fine with simply printing a warning and doing nothing else.

pascal-brand-st-dev: Another question: should I raise an error or not:
* if no error, then this is not dangerous because we may think it has been cleaned, but it has not
* if an error is raised, then, from build, the first `make all-clean` will work, but not the second one

(So long thread for something short)

jforissier: Well I'm changing my mind ;o
All things considered, I still prefer `rm -rf $(out-dir)/ta $(out-dir)/xtest` which will always clean stuff properly and does not bring a big risk of accidental deletion since paths end with `ta` and `xtest`.
See, there are even `rm -rf`'s in the Linux makefiles: [here](https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/arch/x86/Makefile?id=refs/tags/v4.3-rc5#n258) ;-)

jenswi-linaro: :-)

Well, I'd much rather do an accidental `rm -rf /ta` than `rm -rf /`, but who can predict what people may store in `/ta` on their desktop machine?

I'm still opposed to this `rm -rf` especially since it's a makefile I can't avoid.

jforissier: You're not supposed to build as root, and chances are that anything possibly named /ta would be read-only to a regular user. Anyway, then I suggest the option: do nothing (well just print a warning message actually) and do not return an error status. That may be the most reasonable choice, what do you think?

pascal-brand-st-dev:
Updated according to Jerome's last comment.

jforissier:

Makefile
jenswi-linaro: Sounds good.

jenswi-linaro:
```
```

PR #47, opened 2015-10-27
Add concurrency test 1013

jenswi-linaro:
Adds ta_concurrent and test case 1013 to test concurrency.


host/xtest/xtest_1000.c
```diff
@@ -1017,3 +1030,96 @@ static void xtest_tee_test_1012(ADBG_Case_t *c)
...
+	TEEC_ReleaseSharedMemory(&shm);
+
+	Do_ADBG_Log("Max concurrency %zu", max_concurrency);
+	/*
+	 * Concurrency can be limited by several factors, for instance in a
+	 * single CPU system it's dependent on the Preemtion Model used by
+	 * the kernel (Preemptible Kernel (Low-Latency Desktop) gives the
+	 * best result there).
+	 */
+	(void)ADBG_EXPECT_COMPARE_UNSIGNED(c, max_concurrency, >, 0);
```
jforissier: Should be `>, 1` otherwise you're not checking that at least some level of concurrency was possible.
Maybe this test should be skipped when optee_os was built with `CFG_NUM_THREADS = 1`?

jforissier:
What about replacing the SHA code by a simple busy loop? It should be enough to validate concurrency I suppose. Something like:
```
while (num_rounds--)
        asm(""); /* Prevents GCC optimization at -O3 */
```

host/xtest/xtest_1000.c
jenswi-linaro: Then the test could fail sometimes when running on a single CPU system depending on how the Linux kernel is configured.

jenswi-linaro:
Yes we could do that, if num_rounds is declared volatile it will not be optimized.

host/xtest/xtest_1000.c
jforissier: True, single CPU (maxcpus=1) *and* `PREEMPT` disabled will make `> 1` fail (I tried on HiKey). But single CPU and `PREEMPT` enabled is OK, as well as multiple CPUs (with or without `PREEMPT`).
Since `PREEMPT` is enabled by default in the upstream kernel, I think it's safe to tighten the test a bit with `max_concurrency > 1`. Otherwise, this whole test 1013 loses a large part of its value IMO.
As for checking that `CFG_NUM_THREADS >= 2` and `CFG_DISABLE_CONCURRENT_EXEC != y`, it is easy enough since we have the `conf.mk` generated at OP-TEE build time.

jenswi-linaro: `PREEMPT_NONE` is default in the upstream kernel (v4.3). Leaving that as is will break xtest on qemu_virt and any other single CPU system. Prio 1 is no false positives.

Even if the test isn't used as it is right now to assert the amount of concurrency, it still tests that secure world doesn't crash or something.

jforissier: On Mon, Nov 2, 2015 at 9:25 AM, Jens Wiklander <notifications@github.com>
wrote:
>
> PREEMPT_NONE is default in the upstream kernel (v4.3). Leaving that as is
> will break xtest on qemu_virt and any other single CPU system. Prio 1 is no
> false positives.
>
OK. Well, PREEMPT *is* the default for arm64, due to CONFIG_PREEMPT=y
in arch/arm64/configs/defconfig. But you're right, many arm defconfig's do
not override this, so it is the default PREEMPT_NONE set in
kernel/Kconfig.preempt that applies.

Let's keep the assertion: `> 0`, then.

> Even if the test isn't used as it is right now to assert the amount of
> concurrency, it still tests that secure world doesn't crash or something.
>

True.

jenswi-linaro:
Updated with busy loop instead of a big SHA-256 implementation.
Added another TA function for SHA-256 using TEE Internal API.

jforissier:
Doing TEE Internal API calls is a good idea.

Unfortunately, the test fails on HiKey due to `num_threads = 5`. Currently you can't do more than 3, or `TEEC_OpenSession()` fails with `TEEC_ERROR_OUT_OF_MEMORY`. I can probably adjust the HiKey memory layout so that 5 threads will run OK (15MiB are reserved by UEFI and not mapped by OP-TEE atm). But, wouldn't it be better to set `num_threads = 3`? Other platforms might have the same limitation.

jenswi-linaro:
I'll set it to 3 instead.

jenswi-linaro:
Updated number of parallel threads to 3 instead.

jforissier:

jenswi-linaro:
Tags applied

jenswi-linaro:
Rebased

PR #50, opened 2015-11-04
Make cross compile variables configurable via the environment

b49020:
Required to override cross compile variables with cross compile
variables set by yocto environment.
Also LDFLAGS should be empty to properly compile trusted application
with yocto.

Signed-off-by: Sumit Garg <b49020@freescale.com>

### DIFF
--- a/ta/Makefile
+++ b/ta/Makefile
@@ -5,6 +5,10 @@ ifeq ($O,)
 $(error output path should be specified when calling this makefile)
 endif
 
+# Prevent use of LDFLAGS from the environment. For example, yocto exports
+# LDFLAGS that are suitable for the client applications, not for TAs
+LDFLAGS=
+
 TA_DIRS := create_fail_test \
 	   crypt \
 	   os_test \
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,7 @@ endif
 .PHONY: xtest
 xtest:
 	$(q)$(MAKE) -C host/xtest CROSS_COMPILE="$(CROSS_COMPILE_HOST)" \
+			     --no-builtin-variables \
 			     q=$(q) \
 			     O=$(out-dir)/xtest \
 			     $@
--- a/host/xtest/Makefile
+++ b/host/xtest/Makefile
@@ -12,14 +12,14 @@ include $(TA_DEV_KIT_DIR)/host_include/conf.mk
 # be used instead.
 OPTEE_CLIENT_EXPORT ?= ../../../optee_client/out/export
 
-CC		:= $(CROSS_COMPILE)gcc
-CPP		:= $(CROSS_COMPILE)cpp
-LD		:= $(CROSS_COMPILE)ld
-AR		:= $(CROSS_COMPILE)ar
-NM		:= $(CROSS_COMPILE)nm
-OBJCOPY		:= $(CROSS_COMPILE)objcopy
-OBJDUMP		:= $(CROSS_COMPILE)objdump
-READELF		:= $(CROSS_COMPILE)readelf
+CC		?= $(CROSS_COMPILE)gcc
+CPP		?= $(CROSS_COMPILE)cpp
+LD		?= $(CROSS_COMPILE)ld
+AR		?= $(CROSS_COMPILE)ar
+NM		?= $(CROSS_COMPILE)nm
+OBJCOPY		?= $(CROSS_COMPILE)objcopy
+OBJDUMP		?= $(CROSS_COMPILE)objdump
+READELF		?= $(CROSS_COMPILE)readelf
 
 ifdef CFG_GP_PACKAGE_PATH
 GP := _gp


jenswi-linaro:
```
```

ta/Makefile
```diff
@@ -5,6 +5,8 @@ ifeq ($O,)
 $(error output path should be specified when calling this makefile)
 endif
 
+LDFLAGS=
```
jforissier: This is not consistent with the other changes in this patch: you want to be able to override `CC`, `CPP`, `LD` etc. via the environment, but you don't want the environment to influence `LDFLAGS`. This does not sound good.
Also, this introduces a different behavior between environment and command line (`LDFLAGS=foo make` and `make LDFLAGS=foo` will not do the same), which I don't really like.
Can't you just clear `LDFLAGS` before running this Makefile?

jbech-linaro: Agree, this also makes our current xtest builds fail (just tried this patch on QEMU).

jbech-linaro: In fact, it's not the LDFLAGS that makes the build fail, it is the other changes in `xtest/Makefile` that is the culprit. The reason is that there are a few of the variables which already are implicitly set from the start:
```
CC: cc
CPP: cc -E
LD: ld
AR: ar
```
So, those will always be set when this file is entered and therefore the right side after `?=` will __never__ be used.

https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html

b49020: This was particularly required to compile trusted application with yocto. For client application you can override LDFLAGS as well. From my understanding to compile trusted application we don't require external LDFLAGS as set by yocto environment. If I compile using make LDFLAGS=, client application fails to compile. Can you suggest proper way of doing this?

jforissier: @jbech-linaro ah, true. `$(origin ...)` can help here.
```
$ cat Makefile 
define set-default
$(if $(filter $(origin $(1)),default),$(eval $(1):=$(2)))
endef

$(call set-default,CC,makefile-default-cc)

all:
	@echo CC=$(CC)
$
$ make
CC=makefile-default-cc
$ CC=mycc make
$ CC=mycc make
CC=mycc
$ make CC=mycc
CC=mycc
$ CC=ignored make CC=mycc
CC=mycc

```

b49020: Can we use ‘-R’ or ‘--no-builtin-variables’ option to cancel all variables used by implicit rules?
If not, can you suggest a way how can I override CC, CPP, LD, AR etc by cross compile variables set by yocto?

jbech-linaro: Sumit, your suggestion is working, i.e, the following change on top of your patch (and I think this is more clean than the suggestion from @jforissier above).
```
diff --git a/Makefile b/Makefile
index e199ff9..f4d8622 100644
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,7 @@ endif
 .PHONY: xtest
 xtest:
        $(q)$(MAKE) -C host/xtest CROSS_COMPILE="$(CROSS_COMPILE_HOST)" \
+                            --no-builtin-variables \
                             q=$(q) \
                             O=$(out-dir)/xtest \
                             $@
```

So, then that would be OK I think. But then we still have the remaining question about the LDFLAGS. How did you call when you tried Jeromes suggestion?
  `make LDFLAGS=`
or
  `LDFLAGS= make`
?

b49020: As I have replied before also make LDFLAGS= or LDFLAGS= make fails to compile client application as it fails to load standard headers needed to compile it.

jforissier: we could add `LDFLAGS ?= $(TA_LDFLAGS)` to `ta/Makefile` and `LDFLAGS ?= $(XTEST_LDFLAGS)` to `host/xtest/Makefile`. Then you can choose which flags (for which binary) you're setting.

b49020: Yeah this would be more clean.

b49020: But this would again not work in yocto environment as LDFLAGS is set by the environment so it will override LDFLAGS for TA too. Based on my understanding why do we need to configure LDFLAGS by external environment to compile trusted application as its headers and libraries would be provided by Trusted OS itself? For client application it seems right to override LDFLAGS. Kindly correct me if I am missing something.

jforissier: @b49020 yeah I should have written `=` instead of `?=`. OK, let's not over-engineer things, now it seems to me your first proposal is valid for the uses cases we want to deal with. May I suggest that you add a comment, such as the following:
```
# Prevent use of LDFLAGS from the environment. For example, yocto exports
# LDFLAGS that are suitable for the client applications, not for TAs
LDFLAGS=
```
@jbech-linaro what do you think? I am OK with the `--no-builtin-variables` flag BTW.

b49020: Ok, will add this comment.

jbech-linaro: OK, I'm also fine with this if you @b49020 :
a) Add the `--no-builtin-variables` as I suggested in my comment above.
b) Add the clarifying comment proposed by Jerome just above.

Please also rebase and add our tags and I'll merge that patch.

jforissier: With @jbech-linaro comments incorporated, this is also:

b49020:
I have done the rebase, taken care of all the comments and put the tags on it.

PR #51, opened 2015-11-13
ta: 64-bit fixes

jenswi-linaro:
64-bit fixes for TAs to work and compile for Aarch64.

### DIFF
--- a/ta/crypt/cryp_taf.c
+++ b/ta/crypt/cryp_taf.c
@@ -34,6 +34,8 @@ do {                                             \
 		return TEE_ERROR_BAD_PARAMETERS; \
 } while (0)
 
+#define VAL2HANDLE(v) (void *)(uintptr_t)(v)
+
 TEE_Result ta_entry_allocate_operation(uint32_t param_type, TEE_Param params[4])
 {
 	TEE_Result res;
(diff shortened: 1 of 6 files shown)


ta/crypt/cryp_taf.c
```diff
@@ -47,104 +49,121 @@ TEE_Result ta_entry_allocate_operation(uint32_t param_type, TEE_Param params[4])
 	res = TEE_AllocateOperation(&op,
 				    params[0].value.b, params[1].value.a,
 				    params[1].value.b);
-	params[0].value.a = (uint32_t) op;
+	params[0].value.a = (uintptr_t)op;
```
jforissier: `a` is still `uint32_t`, isn't it? Then what is the guarantee that `op` isn't larger?

jenswi-linaro: A pointer passed as handles to TAs are converted with `tee_svc_kaddr_to_uref()` to make it a low number.

jforissier: OK, got it, it will work. But it does not make the TA really 64-bit clean. I mean, ideally the TA should not need to assume anything about the value of a `TEE_OperationHandle`. What about reserving both `params[0].value.a` and `params[0].value.b` for the handle? Would it be too intrusive a change?

jenswi-linaro: This is only for testing our implementation. It works now, if we'd like to make it more portable to be able to test other TEEs we can improve that later.

jbech-linaro:

jforissier:

jenswi-linaro:
Rebased and tags applied

jenswi-linaro:
Rebased

PR #56, opened 2015-12-03
xtest: name of default test suite is regression

jenswi-linaro:
xtest without arguments tries to run a test suite regression, but the
suite isn't found as the rest of the code thinks the name is supposed to
be sanity. This patch fixes that.


jforissier:

jenswi-linaro:
Tag applied

PR #57, opened 2015-12-04
64-bit TA: fix tests 8009 and 8062

jforissier:
These tests (8009: GP a7-2b-1f, 8062: GP a7-c0-a0) incorrectly use
sizeof(ptr) when the length of the string pointed to by ptr is meant.
Since the size of the string ('TEE\0') is 4, this would happen to work
in 32-bit mode, but fails when TAs are built as 64-bit binaries.

This commit also fixes a buffer overflow on the host side, caused by
the same mistake.


pascal-brand-st-dev:

jforissier:
Review tag applied.

jforissier:
And rebased ;-)

PR #59, opened 2015-12-17
Run xtest 1013 test 50 times

pascal-brand-st-dev:
xtest 1013 tests the concurrent TA feature. It lauches several threads
in parallel. Hence, the running behavior is quite sensible, and an
issue in concurrent TA could be raised or not depending on the
runtime.

Running the tests several times allow to run the test in slightly
different configuration, and has a higher probability to raise
potential issues.

### DIFF
--- a/host/xtest/xtest_1000.c
+++ b/host/xtest/xtest_1000.c
@@ -1080,9 +1080,11 @@ static void *test_1013_thread(void *arg)
 	return NULL;
 }
 
-static void xtest_tee_test_1013(ADBG_Case_t *c)
+#define NUM_THREADS 3
+
+static void xtest_tee_test_1013_single(ADBG_Case_t *c, double *mean_concurrency)
 {
-	size_t num_threads = 3;
+	size_t num_threads = NUM_THREADS;
 	size_t nt;
 	size_t n;
 	pthread_t thr[num_threads];
@@ -1098,6 +1100,7 @@ static void xtest_tee_test_1013(ADBG_Case_t *c)
 	};
 	uint8_t out[32] = { 0 };
 
+	*mean_concurrency = 0;
 
 	memset(&shm, 0, sizeof(shm));
 	shm.size = sizeof(struct ta_concurrent_shm);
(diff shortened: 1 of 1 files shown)


host/xtest/xtest_1000.c
```diff
@@ -1182,3 +1182,11 @@ static void xtest_tee_test_1013(ADBG_Case_t *c)
...
 	TEEC_ReleaseSharedMemory(&shm);
 }
+
+static void xtest_tee_test_1013(ADBG_Case_t *c)
+{
+	int i;
+	for (i = 0; i < 50; i++)
+		xtest_tee_test_1013_single(c);
+}
+
```
jforissier: The problem I see with this: it will cause 50 times more text output for the test, and it will count 50 times more sub-cases. IMO it would be better to loop inside `xtest_tee_test_1013()` -- you could extract the two scenarios (busy loop/hashing) and iterate over them.

jenswi-linaro: +1, with less text output more pressure will be applied by the test also.

pascal-brand-st-dev:
Updated

host/xtest/xtest_1000.c
```diff
@@ -1106,9 +1106,6 @@ static void xtest_tee_test_1013(ADBG_Case_t *c)
 		TEEC_AllocateSharedMemory(&xtest_teec_ctx, &shm)))
 		return;
 
-	Do_ADBG_BeginSubCase(c, "Busy loop with %zu parallel threads",
```
jenswi-linaro: This information would be interesting to see, perhaps print it before the loop calling this function.

host/xtest/xtest_1000.c
```diff
@@ -1130,8 +1127,6 @@ static void xtest_tee_test_1013(ADBG_Case_t *c)
 			max_concurrency = arg[n].max_concurrency;
 	}
 
-	Do_ADBG_Log("Max concurrency %zu", max_concurrency);
```
jenswi-linaro: This information is interesting to see, perhaps print it after the loop calling this function is ended. I guess we're interested in the average max concurrency.

pascal-brand-st-dev:
updated

jforissier:
Why did you drop `Do_ADBG_BeginSubCase` / `Do_ADBG_EndSubCase`? I think having them once for the busy loop, and once for the hash test would be good. They are useful because they are parsed by the Expect script during the Travis build, that's what makes the `####` progress and the total number is reported in the end.

jenswi-linaro:
@jforissier the blame is on me.
The top case (1013) is still printed. Does it matter that the number of sub-cases aren't counted here?

jforissier:
Well that's OK, the start of the top case is indeed counted by the Expect script. The new xtest 1013 will count for 1 instead of 2 in the Expect report, not a problem really.

`Reviewed-by: Jerome.Forissier <jerome.forissier@linaro.org>`

jenswi-linaro:
```
```

PR #60, opened 2015-12-17
Enable 9000 tests related Invoke_Crypto_AsymmetricDecrypt

pascal-brand-st-dev:
3 issues are fixed:
- RSA crypto key (TEE_ATTR_RSA_MODULUS_VALUE01 /
  TEE_ATTR_RSA_PRIVATE_EXPONENT_VALUE01) results in an encrypted equal to
  0 which is wrong.
  I think these are not prime numbers.
  When replacing with the ones of ac_rsassa_vect2_xxx[], then the
  encryption is correct
- The number to encrypt, DATA_FOR_CRYPTO1, starts with a 0, and is a 96
  bytes number. That means it is the same number as the one of 95 bytes
  without the first 0
  The first bytes has been replaced by '1' instead to remove
  any ambiguity
- Invoke_Crypto_AsymmetricDecrypt(): it expects that the encrypted has
  the same length as the initial message. In the case of RSA-2048 bits,
  the encrypted of a 64bytes number is 256bytes long.
  This is now taken into account

Enabled tests are:
	9053 9074 9081 9087 9092 9104 9111
	9145 9146 9147 9148 9149
	9181 9182 9183 9184 9186 9187
	9195 9196


pascal-brand-st-dev:
@jforissier ping? (no hurry)

host/xtest/xml/include/xml_crypto_api.h
```diff
@@ -182,37 +182,28 @@ static const uint8_t TEE_ATTR_DH_PRIVATE_VALUE_VALUE01[] = {
...
-	0x47, 0x6b, 0x61, 0x18,
-	0x4d, 0xfe, 0xfd, 0x2a, 0x7e, 0x77, 0x40, 0x44, 0x43, 0xc6, 0x33, 0x6c,
-	0xe5, 0x1b, 0x8d, 0x80,
-	0xf9, 0x97, 0xa2, 0xe4, 0xb9, 0x34, 0x3e, 0x28, 0x94, 0x9f, 0xbd, 0xa8,
-	0x2b, 0x0a, 0x4d, 0x1a,
-	0xa8, 0x06, 0xe5, 0x99, 0x4e, 0xb9, 0x13, 0x45, 0xc8, 0xf6, 0x0f, 0xd0,
-	0x4d, 0xbf, 0xe7, 0x8f,
-	0xed, 0xca, 0x8e, 0xf8, 0x8d, 0x87, 0x5f, 0xd4, 0xb4, 0x1a, 0x2c, 0xc9,
-	0xa7, 0x67, 0x7e, 0xb2,
-	0x1b, 0xc
```
jforissier: No need to completely change the value here. The original modulus is simply truncated, see TEE_Initial_Configuration-Test_Suite_v1_1_0_4-2014_11_07/packages/Crypto/html/TEE_Crypto_API_operationsDescriptions.html.
Remove the last `0xc` and append the following:
`0xc1, 0xce, 0xb6, 0x83, 0x7c, 0xce, 0xb4, 0x3d, 0x85, 0xc7, 0x53, 0x30, 0x7c, 0xfe, 0x85`

host/xtest/xml/include/xml_crypto_api.h
```diff
@@ -275,38 +266,28 @@ static const uint8_t TEE_ATTR_DES3_192_VALUE01[] = {
 	0x89, 0x5D, 0xAD, 0xF2, 0x9D, 0x62, 0xEF, 0x25
 };
 static const uint8_t TEE_ATTR_RSA_PRIVATE_EXPONENT_VALUE01[] = {
```
jforissier: Then this will not be needed

host/xtest/xml/include/xml_crypto_api.h
```diff
@@ -374,7 +355,11 @@ static const uint8_t NONCE1_VALUE_AES_CCM[] = {
 
 /*ALL_CRYPTO_DATA_VALUE*/
 static const uint8_t DATA_FOR_CRYPTO1[] = {
-	0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0A, 0x0B,
+	/*
+	 * do not start we a '0' as this is the representation of a number, so
+	 * the most significant '0' are removed after encrypt / decrypt
+	 */
+	0x01, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0A, 0x0B,
```
jforissier: I don't think it's OK. We should be able to encrypt any data buffer! At least I see no restriction in the spec for `TEE_AsymmetricEncrypt()`. I think this needs further investigation.

pascal-brand-st-dev:
Updated

jforissier:
I think I finally understood the RSA size issue (and solution) ;)

PR #67, opened 2016-01-22
Disable xtest 20000 (FS corruption tests) when OP-TEE is build for RPMB

jforissier:
When the filesystem is implemented over RPMB (CFG_RPMB_FS=y), there is
not point trying to corrupt files in the REE.


jenswi-linaro:
Would it be possible to do something similar when using emulated RPMB?

jforissier:
@jenswi-linaro yes it should be possible to do something similar, but it would not be straightforward because:
1. The tests we have currently look very specific to the REE implementation, so we would need to rewrite them
2. File corruption cannot be done with a simple fopen/fwrite. We would need a communication channel to the tee-supplicant, which does not exist currently.

jenswi-linaro:
@jforissier OK

pascal-brand-st-dev:
I would prefer to keep coherency between what we have in conf.mk, and what is defined. This means leave CFG_ENC_FS defined if it is 'y' in conf.mk. I would prefer to fix .c, from

    #ifdef CFG_ENC_FS

into

    #if defined(CFG_ENC_FS) && !defined(CFG_RPMB_FS)

jforissier:
@pascal-brand-st-dev you're right, it's cleaner. Updated.

host/xtest/xtest_20000.c
```diff
@@ -11,7 +11,7 @@
  * GNU General Public License for more details.
  */
 
-#ifdef CFG_ENC_FS
+#if defined(CFG_ENC_FS) && !defined(CFG_RPMB_FS)
```
pascal-brand-st-dev: Not needed in `xtest_main.c` and `xtest_test.h`?

jforissier: Sorry forgot to add them. Fixed.

pascal-brand-st-dev:

jforissier:
Rebased and tags applied.

PR #72, opened 2016-03-02
xtest client: compilation mode (32bits / 64bits) fix

pascal-brand-st-dev:
Compilation of xtest requires linking to the precompiled library
libcrypto.a. Hence it needs to know in which mode (32bits / 64bits) xtest
is compiled.

Before this patch, it was believed that the compilation mode of the client
part is the same as the optee_os core part. This is not true and fixed
using optional environment variable COMPILE_NS_USER, which is defined
by default the same as the optee_os core compilation mode


jforissier:

PR #73, opened 2016-03-03
Misc

jenswi-linaro:


jforissier:
About the first patch: I suppose optee_os has to be built with `CFG_TEE_CORE_EMBED_INTERNAL_TESTS=y`, which is not the default for all platforms currently. So, either optee_os should be updated or test 1001 should not fail when the static TA is not available.

jenswi-linaro:
Update

jforissier:

jenswi-linaro:
Tags applied

pascal-brand-st-dev:
@jenswi-linaro Could rebase on master? Thanks.

jenswi-linaro:
Rebased

PR #78, opened 2016-04-08
Android makefiles for trusted applications

kuscsik:
Signed-off-by: Zoltan Kuscsik <zoltan.kuscsik@linaro.org>


jbech-linaro:
So this is for putting the TA's on the file system and not actually compile them, since that must be done using the ta-dev-kit from optee_os. I think this is a good start. Later on it would be nice to automatically compile the TA's also if it doesn't make thing too complicated and clutter down to much. But again, this is a good start.

kuscsik:
I'm playing with adding a script that generates the prebuilt TA files for Android. See the README:

https://github.com/kuscsik/aosp_manifest

pascal-brand-st-dev:
* Does it hurt to have 1 Android.mk per TA, instead of a global one?
* enc_fs TA is missing

kuscsik:
Added separate Android.mk - files per TA. The enc_fs is just I header? Where is the TA?

ta/concurrent/Android.mk
```diff
@@ -0,0 +1,7 @@
+include $(CLEAR_VARS)
```
pascal-brand-st-dev: I understood they should be named `Android.mk`, but something like `ta_android.mk` Am I wrong.

pascal-brand-st-dev:
I was wrong about enc_fs. Just forget about it.

ta/concurrent/Android.mk
kuscsik: I checked some of the upstream android projects and it is apparently fine to do this.
See for example:
https://android.googlesource.com/platform/external/zxing/+/android-5.1.1_r37

Where the Android.mk includes Android.mk from the core/ subfolder.
I hope to avoid confusion between GNU and Android make files.

kuscsik: The all-makefiles-under macro is doing the trick:

https://android.googlesource.com/platform/build/+/android-5.1.1_r37/core/definitions.mk#141

pascal-brand-st-dev:
`Reviewed-by: Pascal Brand <pascal.brand@linaro.org>`

kuscsik:
I have a missing "\" I noticed it at rebuild. Fixing.

kuscsik:
Fixed

PR #81, opened 2016-04-14
Experimental, work around for Android build

d3zd3z:
This contain two of @kuscsik unmerged change, plus a hack, since I can't figure out what is supposed to be setting `TA_DEV_KIT_DIR`.

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -3,6 +3,9 @@ LOCAL_PATH := $(call my-dir)
 VERSION = $(shell git describe --always --dirty=-dev 2>/dev/null || echo Unknown)
 OPTEE_CLIENT_PATH ?= $(LOCAL_PATH)/../optee_client
 
+# Not sure why this isn't being set for me.
+TA_DEV_KIT_DIR := optee/optee_os/out/arm-plat-hikey/export-ta_arm64
+
 -include $(TA_DEV_KIT_DIR)/host_include/conf.mk
 
 include $(CLEAR_VARS)


d3zd3z:
Updated this as "experimental" rather than "hack".  At this point, I think we should go ahead and review this for merge, since it is needed to build Android.

jbech-linaro:
I don't think this looks like a hack, you need to point to the TA dev kit in some way. In the future if/when we need to support more than HiKey, then we might have to fine tune this a bit, but for now I think this is good.

Android.mk
```diff
@@ -3,6 +3,9 @@ LOCAL_PATH := $(call my-dir)
 VERSION = $(shell git describe --always --dirty=-dev 2>/dev/null || echo Unknown)
 OPTEE_CLIENT_PATH ?= $(LOCAL_PATH)/../optee_client
 
+# Not sure why this isn't being set for me.
+TA_DEV_KIT_DIR := optee/optee_os/out/arm-plat-hikey/export-ta_arm64
```
kuscsik: Path will be different for 32bit and 64bit OP TEE os builds.

jforissier: +1, this has nothing to do here. Can't you just set `TA_DEV_KIT_DIR` as a make argument or an environment variable?

d3zd3z: The proper fix is probably going to depend on having the rest of optee built by android makefiles. Meanwhile, without this, it won't build. Maybe it can be set in a device file?

kuscsik:
This will work fine:

$ make TA_DEV_KIT_DIR= optee/optee_os/out/arm-plat-hikey/export-ta_arm64

Android.mk
kuscsik: This will work fine:

$ make TA_DEV_KIT_DIR= optee/optee_os/out/arm-plat-hikey/export-ta_arm64

d3zd3z: I'm not sure I'd describe that as fine, though. This path is an internal detail of the build system and doesn't really belong in the users command line.

jforissier: Agreed, so what about that device file you mentioned? Would it be an option?

d3zd3z: I'll try it next week, when I return from vacation.

d3zd3z:
It does work putting this value in the device.mk file, which is a much better place for it. No need to pull this.

PR #88, opened 2016-05-06
ta Android.mk files: add LOCAL_PATH for each file

liuyq:
so that we can have LOCAL_SRC_FILES defined with LOCAL_MODULE,
and possible to create the dependency for LOCAL_SRC_FILES
in the device configuration repository


ta/storage/Android.mk
```diff
@@ -1,6 +1,7 @@
+LOCAL_PATH := $(call my-dir)
+
 include $(CLEAR_VARS)
 LOCAL_MODULE := b689f2a7-8adf-477a-9f9932e90c0ad0a2.ta
```
jbech-linaro: Shouldn't it be `+LOCAL_SRC_FILES := $(LOCAL_MODULE)` below here also?

liuyq: Yes, should have that here as well.
Will send a new request later.

jbech-linaro: You're aware of that you can push push new changes to the same branch to update it? I.e., you don't have to close a pull request and create another one just to make changes.

liuyq: Ah, sorry, I did not know how to do that with git hub:(

PR #102, opened 2016-06-21
Fix shm allocation flags in 8051 to 8054

pascal-brand-st-dev:
Allocation flags, in test 8051 to 8054, were TEEC_MEMREF_PARTIAL_INPUT,...
instead of TEEC_MEM_INPUT,...


jbech-linaro:

jforissier:
So this means the GP test is wrong? `TEE_Internal_API.xml` uses `TEEC_MEMREF_PARTIAL_INPUT` but it should be using `TEEC_MEM_INPUT` etc.?
While I think it is correct, I'm wondering: what is the value of the GP test suite actually? It seems we are using a heavily patched version already. We are more or less fixing things without feedback from GP, are we doing the right thing? If so, those fixes should go "upstream", how can we do that?

Anyway:

pascal-brand-st-dev:
tags applied

PR #103, opened 2016-06-29
Add install target

jforissier:
It is pretty common to have an install target that can be used like
this [1]:

  make install DESTDIR=/some/dir

[1] https://www.gnu.org/prep/standards/html_node/DESTDIR.html

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,8 +10,10 @@ endif
 
 ifneq ($V,1)
 	q := @
+	echo := @echo
 else
 	q :=
+	echo := @:
 endif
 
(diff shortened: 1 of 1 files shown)


jbech-linaro:
I'm fine with the rule, but I wonder when we are going to use it? `make install` is typically something you run when building, install and running on the same machine. But, no matter what, this is good to have.

jforissier:
Right, I should maybe have added more context to my commit message ;)
This comes in handy when you want to install into a "root FS" local directory exported via NFS.

PR #106, opened 2016-07-21
os_test: cleanup undef instruction test

jenswi-linaro:
Cleans up undefined instruction test by using a separate
assembly implementation for the undefined instruction.

Tested-by: Jens Wiklander <jens.wiklander@linaro.org> (QEMU, FVP)

### DIFF
--- a/ta/os_test/os_test.c
+++ b/ta/os_test/os_test.c
@@ -832,6 +832,17 @@ TEE_Result ta_entry_wait(uint32_t param_types, TEE_Param params[4])
 	return res;
 }
 
+static void undef_instr(void)
+{
+#if defined(ARM64)
+	__asm__(".word 0x0");
+#elif defined(ARM32)
+	__asm__(".word 0xe7ffffff");
+#else
+#error "Unsupported architecture"
+#endif
+}
+
 TEE_Result ta_entry_bad_mem_access(uint32_t param_types, TEE_Param params[4])
 {
 	long stack;
@@ -854,10 +865,7 @@ TEE_Result ta_entry_bad_mem_access(uint32_t param_types, TEE_Param params[4])
 		((void (*)(void))(stack_addr + 0x40000000)) ();
 		break;
 	case 5:
-		{
-			static const uint32_t my_undef = 0xffffffff;
-			((void (*)(void))(uintptr_t) &my_undef) ();
-		}
+		undef_instr();
 		break;
 	default:
 		break;


ta/os_test/os_test.c
```diff
@@ -35,6 +35,7 @@
 #include "os_test.h"
 #include "testframework.h"
 #include "test_float_subj.h"
+#include "undef_instr.h"
```
jforissier: Why not just write a local function and use inline assembly?
```
static void undef_instr()
{
#ifdef ARM64
__asm__(".word 0x0");
#else
__asm__(".word 0xe7ffffff");
#endif
}
```

jenswi-linaro: What if we run on some other arch?

jbech-linaro: Is that very likely as of now? Jeromes proposal doesn't clutter down that much. Also, if we in the future would get other arch's, then I guess we could just expand the `#if ... #else`?

jforissier: OK, that should be:
```
static void undef_instr()
{
#if defined(ARM64)
__asm__(".word 0x0");
#elif defined(ARM32)
__asm__(".word 0xe7ffffff");
#else
#error "Unsupported architecture"
#endif
}
```

jenswi-linaro: Generally I try to avoid inline assembly unless there's some benefit. In this case there isn't except that we can avoid a couple of files.

If you'd rather have it as inline assembly I'll change it, I hope we're not starting a new trend though.

jforissier: OK for generally preferring `.S` over inline assembly. Still, creating 3 source files for one instruction seems a bit excessive. Actually, the Linux kernel CodingStyle discourages the use of inline assembly for complex functions only:

> Chapter 19:  Inline assembly
>
> In architecture-specific code, you may need to use inline assembly to interface with CPU or platform functionality. Don't hesitate to do so when necessary. [...]
> Large, non-trivial assembly functions should go in .S files, with corresponding C prototypes defined in C header files.

If inline assembly is a concern, why not keep the current code? Simply adjust the value of the undefined instruction for A32 and A64, add a comment stating the intent or rename `my_undef` to `undef_instr` and voilà.

jenswi-linaro:
Update

jbech-linaro:

jforissier:

jenswi-linaro:
Tags applied

d3zd3z:
BTW, I did hear back from ARM, and there indeed is no reserved undefined instruction on Aarch64. Their suggestion was to use BRK, but I'm not sure this will cause the same fault, and may interfere with debugging.

PR #109, opened 2016-08-02
Avoid writing to the TA directory

d3zd3z:
Presuming that the tee-supplicant is configured to support loading TAs
from an alternate path, use this alternate path to write test TAs,
instead of trying to rename and overwrite the existing TA images.
This gives several benefits:

- The tests no longer assume that the TA directory is writable.  This
  allows this test to pass on platforms where it is not.

- There is no longer a race with pressing Ctrl-C at an inopportune
  time will leave a corrupted TA file in place instead of the correct
  one.

This change affects test 1008.4 and 1008.5, "Load fake uuid TA", and
"Load corrupt TA".

Signed-off-by: David Brown <david.brown@linaro.org>

### DIFF
--- a/host/xtest/xtest_1000.c
+++ b/host/xtest/xtest_1000.c
@@ -18,6 +18,8 @@
 #ifdef USER_SPACE
 #include <pthread.h>
 #include <unistd.h>
+#include <sys/stat.h>
+#include <sys/types.h>
 #endif
 
 #include "xtest_test.h"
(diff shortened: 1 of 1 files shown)


d3zd3z:
This depends on [optee_client#54|https://github.com/OP-TEE/optee_client/pull/54] in order to load the TAs. Without it, the tests will fail because the supplicant will continue loading the non-corrupt TA.

jforissier:

d3zd3z:
Squashed, and applied review tags.

jforissier:
I was faster than you ;) could you please rebase and re-push?

d3zd3z:
We prefer rebasing to merges?

jforissier:
Yes, to keep the commit history as clean and linear as possible.

PR #111, opened 2016-08-12
Add xtest 6015 (TA storage separation)

jforissier:
xtest 6015 is added to test storage isolation. It invokes a first TA to
create a persistent object. Then it uses a second TA to try and open an
object with the same name, which is expected to fail.

The test also attempts to simulate a file manipulation in the REE file
system (renaming the storage path of TA1 to match the TA2 directory).
This should fail as well.

To obtain two different TAs capable of manipulating persistent objects,
the "storage" TA is built twice with a different UUID. The TA generated
under $(O)/ta/storage is unchanged, and a copy of this TA (but with a
new UUID) is created under $(O)/ta/storage2. This has the advantage of
reducing the amount of new code needed, and also may be useful to
implement concurrency tests in the future.

The new test is not included in the Android build currently. Patch
welcome :)


host/xtest/xtest_6000.c
```diff
@@ -1352,6 +1352,117 @@ static void xtest_tee_test_6014_single(ADBG_Case_t *c, uint32_t storage_id)
 
 DEFINE_TEST_MULTIPLE_STORAGE_IDS(xtest_tee_test_6014)
 
+#ifndef __ANDROID__
```
d3zd3z: I think this will actually just work on Android. I can try it a bit later. The `/data/tee` path is an Androidism anyway, and on non-Android, we really should be using something like `/var/lib/tee`. But that's a different issue to fix. At a minimum, it'd be nice to figure out how to define the data path in a way that could be shared between the supplicant and the tests, however.

host/xtest/xtest_helpers.c
```diff
@@ -381,3 +381,31 @@ TEEC_Result ta_crypt_cmd_free_operation(ADBG_Case_t *c, TEEC_Session *s,
...
+
+int ree_fs_get_ta_dirname(TEEC_UUID *p_uuid, char *buffer, uint32_t len)
+{
+
+	if (p_uuid == NULL || buffer == NULL)
+		return 0;
+
+	return snprintf(buffer, len,
+			"%08X%04X%04X%02X%02X%02X%02X%02X%02X%02X%02X",
+			SWAP_BYTES_32(p_uuid->timeLow),
```
d3zd3z: I find it curious that the filesystem gets the endinaness right on the directory name, but the filenames of the TAs themselves don't. I guess that means there is less to fix than I thought.

host/xtest/xtest_6000.c
jforissier: Yes I think it should work, the only reason why I added this guard is because I didn't know how to hack `Android.mk` to build twice but with a different name and compiler flags ;) And, I have no Android environment available immediately.

d3zd3z:
Well, not only doesn't it yet work, applying this patch breaks many of the 601x tests on Android. I'm investigating now.

jforissier:
@d3zd3z if you get `TEE_ERROR_SECURITY` it's probably because of the UUID not being used consistently (I had a hard time getting `user_ta_header.c` to compile with `-DSTORAGE2` for instance).

d3zd3z:
Building it twice like this is kind of fighting against make, since you're building the same object with two different #defines, make will have no problem using the wrong one.

d3zd3z:
One possibility is somehow having different object directories, but I'm still thinking about how to do that with both regular and Android builds.

Also, after reverting and reapplying the patch, it now just works for me on Android (without the 6015 test). I'm not sure what happened the first time.

d3zd3z:
What about two separate TA directories, where the source files are just `#include "../storage-code/storage.c"` and likewise for the other file? That way, the build system thinks everything is still separate, but the code will remain shared.

jforissier:
Strictly speaking, we do have different object directories due to `O=...`. But I think your suggestion is better than my solution because it will likely integrate better with both build systems (regular and Android).

d3zd3z:
I'm trying an approach now that has two dirs with symlinks. I'm not sure if symlinks is better or worse than #includes. Right now, it isn't doing the right thing in Android. Feel free to upload one if you want me to try it.

d3zd3z:
Another point, adding another TA does require adding an entry to `device/linaro/hikey/device.mk`. It's easy to add, but requires coordination between the two repos.

jforissier:
I prefer #includes over symlinks, just because there are more things that can go wrong with symlinks than with plain files when copying/archiving/moving to a different SCCS etc.

Update pushed, can you please give try it on Android?

d3zd3z:
```
Reviewed-by: David Brown <david.brown@linaro.org>
Tested-by: David Brown <david.brown@linaro.org> (Hikey android)
```

PR #114, opened 2016-08-17
Build xtest with "-DDEBUG -O0 -g3" when "make DEBUG=1"

jforissier:

### DIFF
--- a/host/xtest/Makefile
+++ b/host/xtest/Makefile
@@ -133,6 +133,10 @@ CFLAGS += -Wall -Wcast-align -Werror \
 	  -Wno-missing-field-initializers -Wno-format-zero-length
 endif
 
+ifeq ($(DEBUG),1)
+CFLAGS += -DDEBUG -O0 -g3
+endif
+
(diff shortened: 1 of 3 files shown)


host/xtest/Makefile
```diff
@@ -133,6 +133,10 @@ CFLAGS += -Wall -Wcast-align -Werror \
 	  -Wno-missing-field-initializers -Wno-format-zero-length
 endif
 
+ifeq ($(DEBUG),1)
+CFLAGS += -DDEBUG -O0 -g3
```
jenswi-linaro: What about always using `-g3` regardless of `DEBUG`?

jforissier: You mean just add `CFLAGS += -g3`? I'm OK with that.

jenswi-linaro: Yes

jforissier:
Updated to always build with `-g3` (Android too!).

jenswi-linaro:
The `DEBUG` flag was quite useful too, but doing only the `-g3` right now is OK.

jforissier:
OK let's close this PR and open a new one for `-g3` only.

PR #116, opened 2016-08-19
Fix the invalid free in xtest 9089

prime-zeng:
when the ALLOCATE_SHARED_MEMORY get failed, it will directly go to the
exit, and do an invalid free in CRYPTO_FREE(mac)

Signed-off-by: Zeng Tao <prime.zeng@hisilicon.com>

### DIFF
--- a/host/xtest/xml/include/xml_crypto_api.h
+++ b/host/xtest/xml/include/xml_crypto_api.h
@@ -2046,8 +2046,8 @@ static TEEC_Result Invoke_Crypto_MACCompareFinal(
 
 	res = TEEC_InvokeCommand(s, cmd_id, &op, &ret_orig);
 
-exit:
 	CRYPTO_FREE(mac);
+exit:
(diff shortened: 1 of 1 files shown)


jforissier:

d3zd3z:
This isn't quite right either. Either of the first two allocations could fail as well, and we will try freeing them.

This appears to be incorrect in every instance in this entire file. Perhaps the macro needs to take an exit label parameter so that it can branch to different points depending on what has been allocated.

jforissier:
@d3zd3z I agree, this just fixes the problem of `CRYPTO_FREE(mac)`.

d3zd3z:
I think we should go ahead and apply this change, and I've submitted [pr#117](https://github.com/OP-TEE/optee_test/issues/117) to track the fixing of the rest.

d3zd3z:
```
Reviewed-by: David Brown <david.brown@linaro.org>
```

jforissier:
I have added David's R-b and thus merged the commit manually. Thanks @prime-zeng.

PR #123, opened 2016-09-01
xtest: fix compiler warnings (re-redux)

jforissier:
When CFG_RPMB_FS is not set, the parameter of function
xtest_tee_test_6016() is not used, and other static functions aren't
either. Move the #ifdef sections to exclude as much code as possible
and mark the unused parameter with `(void)c` to squelch the compiler
warnings.


jbech-linaro:

PR #125, opened 2016-09-13
Inclusion of sha-perf into xtest (for details see SWG-185)

igoropaniuk:
Please, read comments before reading code

1. Removed all global variables from sha-perf.c, now all these params
are passed as function arguments
 58 static size_t size = 1024;  /* Buffer size (-s) */
 59 static unsigned int n = 5000;   /* Number of measurements (-n) */
 60 static unsigned int l = 1;  /* Inner loops (-l) */
 61 static int verbosity = 0;   /* Verbosity (-v) */
 62 static int algo = TA_SHA_SHA1;  /* Algorithm (-a) */
 63 static int random_in = 0;   /* Get input data from /dev/urandom (-r)
*/
 64 static int warmup = 2;      /* Start with a 2-second busy loop (-w)
*/
 65 static int offset = 0;          /* Buffer offset wrt. alloc'ed
address (-u) */
2. Because of strict rules (-Werror params are set almost for everything),
added some minor fixes to get sha-perf.c compiled
3. Now we have two entry points for running sha-perf tests:
"sha_perf_runner_cmd_parser" - where we bypass all params from cmdline to
support old-school running of sha-perf test (for the backward
compability).
This function just parses cmdline params and calls sha_perf_run_test.
"sha_perf_run_test" - entry point for starting any sha-based test. We can
use this function in any defined ADBG testsuite in xtest
4. Now we can run sha-perf tests like this. I can also add analyzing of
argv[0], so we could call it like "busybox-way"
$ xtest sha-perf -a SHA1 -s 131072
5. I haven't divided yet sha-perf TA on two pieces (ta_entry.c and
sha-perf.c) to have the same file hierarchy as we have in other TAs.

It's just a POC, I'm going to add additional checks + change some code
in aes-perf. I just try be sure that I'm moving in the right direction

=================== Update =================== 
Added b108783fea5c72125635c0712bf2df75c6f83219 commit and removed RFC tag

1. Performed division of TA on ta_entry.c/ta_sha_perf.c to fit TAs file
hierarhy in xtest
2. Fixed all issues from previous RFC (except verbosity levels and
op.params[0].memref.offset = offset, because decided not to change
internal logic for sha-perf host/TA apps at this stage)
3. Added two testcases just as an example to show how to define your own
testcases for SHA performance tests
4. Minor changes


host/xtest/sha_perf.h
```diff
@@ -0,0 +1,31 @@
...
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+extern int sha_perf_runner_cmd_parser(int argc, char *argv[]);
+extern void sha_perf_run_test(int algo, size_t size, unsigned int n,
```
jforissier: `extern` not needed

host/xtest/xtest_main.c
```diff
@@ -150,6 +154,12 @@ int main(int argc, char *argv[])
 
 	opterr = 0;
 
+	if (argc > 1 && !strcmp(argv[1], "sha-perf")){
+		/* Run sha-perf */
+		sha_perf_runner_cmd_parser(argc-1, &argv[1]);
```
jforissier: `return sha_perf_runner_cmd_parser(...);`

host/xtest/sha_perf.c
```diff
@@ -0,0 +1,480 @@
...
+	vverbose("\n");
+	printf("min=%gμs max=%gμs mean=%gμs stddev=%gμs (%gMiB/s)\n",
+	       stats.min/1000, stats.max/1000, stats.m/1000,
+	       stddev(&stats)/1000, mb_per_sec(size, stats.m));
+	free_shm();
+}
+
+static void usage(const char *progname)
+{
+	fprintf(stderr, "SHA performance testing tool for OP-TEE (%s)\n\n",
```
igoropaniuk: BTW, I've removed default values from usage printing

Do we really need them here?

host/xtest/xtest_main.c
```diff
@@ -150,6 +154,12 @@ int main(int argc, char *argv[])
 
 	opterr = 0;
 
+	if (argc > 1 && !strcmp(argv[1], "sha-perf")){
```
jenswi-linaro: `s/sha-perf/--sha-perf/`

host/xtest/xtest_main.c
jforissier: And for convenience, compare also `argv[0].` Something like this maybe? (untested).
```
char *c;

c = strrchr(argv[0], '/');
if (c)
        c++;
else
        c = &argv[0];
if (!strcmp(c, "sha-perf")
        return sha_perf_runner_cmd_parser(argc, argv);
```

host/xtest/xtest_main.c
igoropaniuk: BTW, what you do think if  also add checking of argv[0] ? We will able to create symlinks like ("busybox" way) 
ln -s aes-perf xtest
and just run it as $ aes-perf [params ]

jforissier:
This looks pretty good IMO.

host/xtest/sha_perf.c
jforissier: Ah, I did not notice that. Yes I think it's important to print the default values. Otherwise how do you know which argument you need to provide?

igoropaniuk: Ok, I'll add it :)

host/xtest/xtest_main.c
jenswi-linaro: The busybox way is an interesting hack to save memory by giving different personalities to the same binary depending on with which name it's started. As we don't really need it I'd rather avoid it to keep things simple.

host/xtest/sha_perf.c
```diff
@@ -0,0 +1,480 @@
...
+				unsigned int l, int random_in, int offset, 
+				int warmup, int verbosity)
+{
+	uint64_t t;
+	struct statistics stats;
+	TEEC_Operation op;
+	int n0 = n;
+	struct timespec ts;
+
+	vverbose("sha-perf version %s\n", TO_STR(VERSION));
```
igoropaniuk: and what about these vverbose macros? I think it will be good to change it to something, that we already have in xtest framework

For example Do_ADBG_Log

host/xtest/xtest_main.c
jforissier: Well that would not add great complexity, and it would save some typing. But I can also create wrapper scripts instead of symlinks so not a big deal.

host/xtest/sha_perf.c
jforissier: Yeah, and also having two levels of verbosity is not really useful so you can keep only one if you wish.

host/xtest/xtest_main.c
jenswi-linaro: It's not a big issue, if you want two ways of doing the same thing it's fine with me. :-)

host/xtest/sha_perf.c
```diff
@@ -0,0 +1,480 @@
...
+					 TEEC_MEMREF_PARTIAL_OUTPUT,
+					 TEEC_VALUE_INPUT, TEEC_NONE);
+	op.params[0].memref.parent = &in_shm;
+	op.params[0].memref.offset = 0;
+	op.params[0].memref.size = size + offset;
+	op.params[1].memref.parent = &out_shm;
+	op.params[1].memref.offset = 0;
+	op.params[1].memref.size = hash_size(algo);
+	op.params[2].value.a = l;
+	op.params[2].value.b = offset;
```
etienne-lms: for clarity, i think you can get rid of `param[2].b` and simply use:
```
op.params[0].memref.offset = offset;
op.params[0].memref.size = size;
```
TA should be change the same: simply using buffer address/size.

jforissier:
How do you invoke the SHA benchmark suite? `xtest -h` says nothing about it. `xtest -t benchmark` runs only the file benchmark.

jforissier:
[Replying to myself...]
I think there is little value in adding a new test suite name, i.e., just add the SHA tests to the "benchmark" suite like this:

```
diff --git a/host/xtest/xtest_main.c b/host/xtest/xtest_main.c
index fc7895c..6f8e064 100644
--- a/host/xtest/xtest_main.c
+++ b/host/xtest/xtest_main.c
@@ -121,10 +121,7 @@ ADBG_SUITE_DEFINE_BEGIN(XTEST_TEE_BENCHMARK, NULL)
 ADBG_SUITE_ENTRY(XTEST_TEE_BENCHMARK_1001, NULL)
 ADBG_SUITE_ENTRY(XTEST_TEE_BENCHMARK_1002, NULL)
 ADBG_SUITE_ENTRY(XTEST_TEE_BENCHMARK_1003, NULL)
-ADBG_SUITE_DEFINE_END()
 
-ADBG_SUITE_DECLARE(XTEST_TEE_BENCHMARK_SHA)
-ADBG_SUITE_DEFINE_BEGIN(XTEST_TEE_BENCHMARK_SHA, NULL)
 ADBG_SUITE_ENTRY(XTEST_TEE_BENCHMARK_2001, NULL)
 ADBG_SUITE_ENTRY(XTEST_TEE_BENCHMARK_2002, NULL)
 ADBG_SUITE_DEFINE_END()
```

(maybe it was the initial intent judging by the test numbers)

igoropaniuk:
I was going to ask about this. I've created different testsuite for sha tests called benchmark_sha, but wasn't sure if it was right decision 

Running testsuite:
`else if (strcmp(test_suite, "benchmark_sha") == 0)  {                                                                                                                        
        ret = Do_ADBG_RunSuite(&ADBG_Suite_XTEST_TEE_BENCHMARK_SHA, argc - optind, (argv + optind));   }`

jforissier:
```
root@HiKey:/ xtest -t benchmark_sha
Run test suite with level=0

No test suite found: benchmark_sha
TEE test application done!
```

.gitignore
```diff
@@ -17,3 +17,4 @@ package/testsuite/global_platform/api_1.0/TEE*
 /libs
 /obj
 /TEE_Initial_Configuration-Test_Suite*
+*.swp
```
jforissier: Should be a separate commit

jforissier:

etienne-lms:
nice work :)

PR #129, opened 2016-09-16
Inclusion of aes-perf benchmark tool into xtest

igoropaniuk:
1. The way of invoking aes-perf tool is just the same as sha-perf 
$ xtest --aes-perf
2. Fixed issues with only-whitespace-lines and >80character-lines
3. Added two examples of test cases for aes-perf


host/xtest/aes_perf.c
```diff
@@ -0,0 +1,474 @@
...
+#include <string.h>
+#include <strings.h>
+#include <time.h>
+#include <unistd.h>
+
+#include <tee_client_api.h>
+#include "ta_aes_perf.h"
+#include "aes_perf.h"
+
+#define _verbose(lvl, ...)			\
```
jbech-linaro: This is also in sha_perf.c, could we move things common to both to a single file?

jforissier:

host/xtest/aes_perf.c
```diff
@@ -0,0 +1,474 @@
...
+					 TEEC_NONE, TEEC_NONE);
+	op.params[0].value.a = decrypt;
+	op.params[0].value.b = keysize;
+	op.params[1].value.a = mode;
+	res = TEEC_InvokeCommand(&sess, TA_AES_PERF_CMD_PREPARE_KEY, &op,
+				 &ret_origin);
+	check_res(res, "TEEC_InvokeCommand");
+}
+
+static void do_warmup(int warmup)
```
jbech-linaro: This is probably more a comment/question to Jerome. I've read the comment from the help above. When do we need to run this and how does it affect the overall test?

host/xtest/aes_perf.c
```diff
@@ -0,0 +1,474 @@
...
+	/*
+	* Command line parameters
+	*/
+
+	size_t size = 1024;	/* Buffer size (-s) */
+	unsigned int n = 5000;	/* Number of measurements (-n) */
+	unsigned int l = 1;	/* Inner loops (-l) */
+	int verbosity = 0;	/* Verbosity (-v) */
+	int decrypt = 0;		/* Encrypt by default, -d to decrypt */
+	int keysize = 128;	/* AES key size (-k) */
```
jbech-linaro: What about defining that in `ta_aes_perf.h` also, just like the TA_AES_ECB etc?

host/xtest/aes_perf.c
```diff
@@ -0,0 +1,474 @@
...
+	}
+	for (i = 1; i < argc; i++) {
+		if (!strcmp(argv[i], "-d")) {
+			decrypt = 1;
+		} else if (!strcmp(argv[i], "-i")) {
+			in_place = 1;
+		} else if (!strcmp(argv[i], "-k")) {
+			NEXT_ARG(i);
+			keysize = atoi(argv[i]);
+			if (keysize != 128 && keysize != 192 &&
```
jbech-linaro: Likewise here with the key sizes (`ta_aes_perf.h`).

host/xtest/aes_perf.c
jforissier: Refactoring can be done in another PR/commit.

jbech-linaro: Yepps, sounds like a plan.

host/xtest/aes_perf.c
igoropaniuk: Like in host/xtest/xtest_benchmark_2000.c ? (see lines 31-36)

host/xtest/aes_perf.c
jforissier: When I was first testing on HiKey I found this "warm-up" would reduce the value of "max" in the report (min/max/mean/stddev), and also slightly reduce stddev. The impact on the reported average time and throughput was not huge IIRC, which is expected given a sufficient number of loops (default value = 5000 seems more than enough to eliminate the effect of a few bogus values on start-up). But still, I was getting more stable results with warm-up enabled than without.'

Now I checked again and the effect is pretty much zero actually. Perhaps something changed in the kernel, like no frequency scaling? (I'm using our optee branch). Indeed, I don't see any `cpufreq` entry in the `/sys` filesystem:
```
root@HiKey:/ find /sys -name \*freq\*
root@HiKey:/
```

jbech-linaro:
Code vise, I like the update, but in the change affecting also sha-perf, think that should go into a single should be in a single commit. I.e, when everything has been squashed there will be two patches ready for merge, one containing the aes-perf stuff and another one that rips out and merge common functionality. That shouldn't be to hard to split up with (if you are familiar with `git add -p`).

host/xtest/aes_perf.c
jforissier: That's totally useless IMO. `TA_AES_ECB` etc. are defined because the client and the secure side have to agree on a way to define the AES modes. I could as well have chosen to transmit a string such as "ECB" or "ecb", but I found it slightly simpler to use numbers. There is no such need for the key size obviously...

jforissier:
@jbech-linaro agreed. Like I said in https://github.com/OP-TEE/optee_test/pull/129#r79130911. 
Thanks for reminding me about `git add -p` ;) that can be handy indeed. The 'edit' command in `git rebase -i` is what I usually do when I need to rewrite history to split commits.

igoropaniuk:
There is only one change that relates to aes-perf stuff - defines for AES key sizes (in ta_aes_perf.h), other changes relate to both aes/sha-perf (verbosity, defines for default values etc.). I can split it in the such way, and add to this PR only AES key sizes? Ok?

For the other stuff I will create another PR

jbech-linaro:
Try to do something like this Igor:
```
============================================
Patch 1: Inclusion of aes-perf benchmark tool into xtest 
  Squash: 
    * Existing stuff
    * Review comments addressed
    * AES defines
============================================
Patch 2: Merge common functionality for xyz-perf
* Things like `_verbose` etc from both files, rename and/or creating of/to crypto_common.h
============================================
```

Having that said, if you think it's more trouble for you to create two commits for a single pull request, then please go a head do as you proposed (and create another PR).

jforissier:
Yes for the split. One logical change, one commit ; we follow the same rules than the Linux kernel (https://github.com/torvalds/linux/blob/92e963f50fc7/Documentation/SubmittingPatches#L201).
OK for a new PR if you don't want to do it in this one, I don't care either way.

igoropaniuk:
Done

host/xtest/xtest_benchmark_2000.c
```diff
@@ -78,3 +79,61 @@ ADBG_CASE_DEFINE(XTEST_TEE_BENCHMARK_2002, xtest_tee_benchmark_2002,
...
+/* ----------------------------------------------------------------------- */
+
+static void xtest_tee_benchmark_2011(ADBG_Case_t *c)
+{
+	UNUSED(c);
+
+	int mode = TA_AES_ECB;	/* AES mode */
+	int decrypt = 0; /* Encrypt */
+	int keysize = AES_128;
+	size_t size = 1024;	/* Buffer size */
```
jforissier: Variables not needed

host/xtest/xtest_benchmark_2000.c
```diff
@@ -78,3 +79,61 @@ ADBG_CASE_DEFINE(XTEST_TEE_BENCHMARK_2002, xtest_tee_benchmark_2002,
...
+
+	aes_perf_run_test(mode, keysize, decrypt, size, CRYPTO_DEF_COUNT,
+		CRYPTO_DEF_LOOPS, CRYPTO_USE_RANDOM, AES_PERF_INPLACE,
+		CRYPTO_DEF_WARMUP, CRYPTO_DEF_VERBOSITY);
+
+
```
jforissier: No blank line needed here

host/xtest/xtest_benchmark_2000.c
```diff
@@ -78,3 +79,61 @@ ADBG_CASE_DEFINE(XTEST_TEE_BENCHMARK_2002, xtest_tee_benchmark_2002,
...
+}
+
+static void xtest_tee_benchmark_2012(ADBG_Case_t *c)
```
jforissier: Same test as 2011?

host/xtest/xtest_benchmark_2000.c
igoropaniuk: fixed

host/xtest/xtest_benchmark_2000.c
igoropaniuk: fixed. Thanks :)

host/xtest/xtest_benchmark_2000.c
igoropaniuk: I know, but what about readability?
It's more convenient (for both reading and changing code), when you see each value per line (with readable and clear variable names). If we put values 1024, 0 etc. directly into the function call, I think people will every time refer to prototype of function to understand what 1024 and 0 means

jforissier: I'm not convinced. You can always add a comment if needed:
```
aes_perf_run_test(TA_AES_ECB, 128 /* keysize */, 0 /* decrypt */,
                  1024 /* bufsize */, ...);
```

Actually I don't like the prototype of `aes_perf_run_test` much (too many parameters and will have to change if we add/remove options). Passing a struct would have been better I suppose, and I should have commented on that in you SHA PR. Anyway, I think we're bikeshedding, and I'm perfectly fine leaving the code as it is.

jforissier:

igoropaniuk:
Tags are applied

PR #130, opened 2016-09-16
Fix 64-bit compile error in case xtest_6000.c

jenswi-linaro:
host/xtest/xtest_6000.c: In function ‘xtest_tee_test_6016_loop’:
host/xtest/xtest_6000.c:1629:4: error: format ‘%d’ expects argument of type ‘int’, but argument 4 has type ‘size_t’ [-Werror=format=]
    "file_%d", n);
    ^
host/xtest/xtest_6000.c:1629:4: error: format ‘%d’ expects argument of type ‘int’, but argument 4 has type ‘size_t’ [-Werror=format=]
cc1: all warnings being treated as errors


jforissier:

jenswi-linaro:
Tag applied

PR #132, opened 2016-09-29
Added info about available benchmarks within xtest tool

igoropaniuk:
What additional info it will be good to add about benchmarks?

FYI, I've used the term "modules" to describe aes-perf/sha-perf test tools, that are included into xtest. We can use them by running: 
$ xterm --sha-perf
If it's OK, I will also change "applets" to "modules" in the usage output for xtest utility


igoropaniuk:
tags applied

jforissier:
Note for next time: you should use the imperative mood for your commit subject, i.e., "Add info about..." rather than "Added info about...".

igoropaniuk:
got it

PR #135, opened 2016-10-20
Android.mk: set TA_DEV_KIT_DIR default value

kuscsik:
TA_DEV_KIT_DIR must be set to non-empty value to
avoid the Android build scripts complaining about
includes pointing outside the Android source tree.

This change fixes the AOSP builds which include
optee_test in their source tree, but OPTEE is not
added to the PRODUCT_PACKAGES.


kuscsik:
Updated.

jenswi-linaro:

kuscsik:
"Reviewed-by" line added.

jforissier:
@kuscsik, could you please rebase on top of optee_test/master? Thanks.

kuscsik:
Sure, sorry for not doing that earlier. It is on master now.

PR #142, opened 2016-10-28
update xtest on new "TA private memory definition"

etienne-lms:
Related to [optee_os/pull#119](https://github.com/OP-TEE/optee_os/pull/1119).

TA private memory definition has changed. Memory reference received
as parameters by an TA when invoked are never "private TA memory",
they are at least shared with the client(s) in some way.

As a side effect, when a TA invokes a TA relaying a memref parameter,
the invoked TA sees memory as ANY_OWNER: it is non secure shared memory.
When a TA invoke a TA with its private memory (stack, heap, code/data)
as memref parameter, the invoked TA sees ANY_OWNER memory, mapped secure,
inside TEE reserved memory (TA RAM).

Since this change, xtest 1006 is broken.
Rename PRIVATE_PARAMS into PARAM_ACCESS tests.
Run TEE_CheckMemoryAccessRights() on several areas of the memory space.
Fix TA that see memref parameters as ANY_OWNER only memory.

### DIFF
--- a/ta/os_test/os_test.c
+++ b/ta/os_test/os_test.c
@@ -409,21 +409,63 @@ static TEE_Result test_mem_access_right(uint32_t param_types,
 	TEE_Param l_params[4] = { { {0} } };
 	uint8_t buf[32];
 	TEE_TASessionHandle sess = TEE_HANDLE_NULL;
+	TEE_UUID *uuid;
 
 	if (param_types !=
 	    TEE_PARAM_TYPES(TEE_PARAM_TYPE_MEMREF_INPUT, 0, 0, 0))
 		return TEE_ERROR_GENERIC;
-	res =
-	    TEE_CheckMemoryAccessRights(TEE_MEMORY_ACCESS_READ |
-					TEE_MEMORY_ACCESS_ANY_OWNER,
-					params[0].memref.buffer,
-					params[0].memref.size);
+
+	/* test access rights on memref parameter */
+	res = TEE_CheckMemoryAccessRights(TEE_MEMORY_ACCESS_READ |
+					  TEE_MEMORY_ACCESS_ANY_OWNER,
+					  params[0].memref.buffer,
+					  params[0].memref.size);
 	if (res != TEE_SUCCESS)
 		return res;
+
 	res = TEE_CheckMemoryAccessRights(TEE_MEMORY_ACCESS_READ,
 					  params[0].memref.buffer,
 					  params[0].memref.size);
 	if (res != TEE_ERROR_ACCESS_DENIED)
+		return TEE_ERROR_GENERIC;
+
+	/* test access rights on private read-only and read-write memory */
+	res = TEE_CheckMemoryAccessRights(TEE_MEMORY_ACCESS_READ,
+					  (void *)&test_uuid, sizeof(test_uuid));
+	if (res != TEE_SUCCESS)
+		return res;
+
+	res = TEE_CheckMemoryAccessRights(TEE_MEMORY_ACCESS_WRITE,
+					  (void *)&test_uuid, sizeof(test_uuid));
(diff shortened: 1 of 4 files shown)


ta/os_test/os_test.c
```diff
@@ -409,23 +409,65 @@ static TEE_Result test_mem_access_right(uint32_t param_types,
...
+
 	res = TEE_CheckMemoryAccessRights(TEE_MEMORY_ACCESS_READ,
 					  params[0].memref.buffer,
 					  params[0].memref.size);
 	if (res != TEE_ERROR_ACCESS_DENIED)
 		return TEE_ERROR_GENERIC;
 
+	/* test access rights on private read-only and read-write memory */
+	res = TEE_CheckMemoryAccessRights(TEE_MEMORY_ACCESS_READ,
+					  (void *)&test_uuid, sizeof(test_uuid));
```
jenswi-linaro: casting shouldn't be needed, please fix the other places too.

ta/os_test/os_test.c
```diff
@@ -409,23 +409,65 @@ static TEE_Result test_mem_access_right(uint32_t param_types,
...
+		return res;
+
+	/* test access rights on invalid memory (at least lower 256kB) */
+	res = TEE_CheckMemoryAccessRights(TEE_MEMORY_ACCESS_READ,
+					  NULL, 1);
+	if (res == TEE_SUCCESS)
+		return TEE_ERROR_GENERIC;
+
+	res = TEE_CheckMemoryAccessRights(TEE_MEMORY_ACCESS_READ,
+					  (void*)(256 * 1024), 1);
```
jenswi-linaro: better cast to a `uintptr_t` first to avoid warnings when compiled for 64-bit

ta/os_test/os_test.c
etienne-lms: i needed to cast as loosing here `const`.
i must crosscheck for few other.

ta/os_test/os_test.c
etienne-lms: `TEE_CheckMemoryAccessRights()` proto declares `void *`. why `uintptr_t` ?

ta/os_test/os_test.c
jenswi-linaro: I see

ta/os_test/os_test.c
etienne-lms: ok... i add this.

jenswi-linaro: I've seen the warning a few times in the past, maybe some warning flag is missing.

jbech-linaro: Why s(w)hould it warn if the declaration says `void *`? It shouldn't matter regardless if your're running 32 or 64bit environment when casting to void pointer, or am I wrong? Wouldn't that be a compiler bug if that happened?

jenswi-linaro: From gcc man page:
```
       -Wno-int-to-pointer-cast
           Suppress warnings from casts to pointer type of an integer of a
           different size. In C++, casting to a pointer type of smaller size
           is an error. Wint-to-pointer-cast is enabled by default.
```

It's no big deal, let's wait until we get the warning and fix it then.

etienne-lms: ok, let's leave it as it is.

jbech-linaro: For a pointer of a defined type I agree, but for void pointers, my experience is that the compiler swallows anything. I also think it's just for us to wait until we face the warning and fix it then.

etienne-lms:
explicit ref to https://github.com/OP-TEE/optee_os/pull/1119

jenswi-linaro:

etienne-lms:
rebased, tags applied.

PR #150, opened 2016-12-08
Format UUID strings correctly

jenswi-linaro:
Prior to this patch although GlobalPlatform specifies that TAs and such
are identified with UUIDs, we don't format them quite right when turning
them into strings. Per https://www.ietf.org/rfc/rfc4122.txt, there
should be another hyphen after the first two bytes of clockSeqAndNode.

Unfortunately, fixing this breaks compatibility between how TAs are
built, and when the OS loads them.

With this patch UUID string are formated with the additional hyphen as
for instance: f81d4fae-7dec-11d0-a765-00a0c91e6bf6

Fixes: https://github.com/OP-TEE/optee_os/issues/857

Depends on https://github.com/OP-TEE/optee_os/pull/1227 and https://github.com/OP-TEE/optee_client/pull/71

### DIFF
--- a/host/xtest/xtest_1000.c
+++ b/host/xtest/xtest_1000.c
@@ -529,7 +529,7 @@ static void uuid_to_full_name(char *buf, size_t blen, const TEEC_UUID *uuid,
 			bool for_write)
 {
 	snprintf(buf, blen,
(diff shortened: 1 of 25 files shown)


jenswi-linaro:
Tag applied

Issue #154, opened 2017-01-02
Docs issues


lws-team:
Hi -

The README.md has some small problems it seems to me.

1) There is no more any `optee_armtz` module referenced in there.  It looks like it would be called "optee" now.

2) There is a missing step in the instructions, that the `*.ta` must be copied into the rootfs, at /lib/optee_armtz (this name seems inconsistent since the module name changed, but it works) for non-Android anyway.  `tee-supplicant` will go fish them out of there and autoload them when the client side asks for the UUID.

Do we agree that these things should be fixed / added to README.md, in which case I can make a PR?  Or we think it's super as it is?

jenswi-linaro:
Thanks, please go ahead and make a PR. :-)

PR #155, opened 2017-01-03
docs: update optee module name and add how to copy to rootfs

lws-team:
Signed-off-by: Andy Green <andy@warmcat.com>

As discussed here

https://github.com/OP-TEE/optee_test/issues/154

### DIFF
--- a/host/xtest/xtest_10000.c
+++ b/host/xtest/xtest_10000.c
@@ -280,11 +280,6 @@ static const uint8_t pbkdf2_3_dkm[] = {
 #define pbkdf2_4_password pbkdf2_1_password
 #define pbkdf2_4_salt pbkdf2_1_salt
 #define pbkdf2_4_iteration_count 16777216
-static const uint8_t pbkdf2_4_dkm[] = {
-	0xee, 0xfe, 0x3d, 0x61, 0xcd, 0x4d, 0xa4, 0xe4,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -54,36 +54,55 @@ This will:
...
+ $ scp ./host/xtest/build/xtest/xtest root@$MY_IP:/bin
+ $ scp `find . -name *.ta` root@$MY_IP:/lib/optee_armtz
+```
 
 ### HOWTO run xtest
 
 	# all xtest
 	boot and execute on your target
-	$ modprobe optee_armtz
+	$ modprobe optee
```
jenswi-linaro: It could be worth mentioning that this step is only needed if the driver is compiled as a loadable module. For easy testing etc I guess it's more or less assumed that the driver is compiled into the kernel.

jbech-linaro: I agree, the default is that it's built into kernel (I cannot recall the last time we built it as a loadable module).

lws-team: Well, I just worked with what was there.  In my case too, it's a built-in.

I'll update the PR later to remove all modprobings.

lws-team:
It's updated per the comments.

etienne-lms:
We can close this. Issue was fixed in commit ba271a6e3eb2bbb6996b5de73918216f64a9e077 (see https://github.com/OP-TEE/optee_test/pull/169)

jforissier:
@etienne-lms I agree. Moreover, the subject of this PR has become totally unrelated to the patch that's still here.

PR #160, opened 2017-01-04
Update Makefile rule for filtering out GP tests

jforissier:
Forgot to adjust the sed rule in parent commit. As a result, unwanted GP tests
would run.

Fixes: b6bea047b1d4 ("Update GP tests after simplification of test declarations")


etienne-lms:

Issue #162, opened 2017-01-05
make clean doesn't seem effective


lws-team:
To allow `make clean` to be effective, I have a local patch

```
diff --git a/Makefile b/Makefile
index 9e263dd..1c2321f 100644
--- a/Makefile
+++ b/Makefile
@@ -46,6 +46,7 @@ ta:
 .PHONY: clean
 ifneq ($(wildcard $(TA_DEV_KIT_DIR)/host_include/conf.mk),)
 clean:
+       @rm -rf `find . -name build`
        $(q)$(MAKE) -C host/xtest O=$(out-dir)/xtest q=$(q) $@
        $(q)$(MAKE) -C ta O=$(out-dir)/ta q=$(q) $@
 else
```

This is based on the observation

```
$ find . -name build
./ta/crypt/build
./ta/storage2/build
./ta/create_fail_test/build
./ta/concurrent_large/build
./ta/os_test/build
./ta/storage_benchmark/build
./ta/storage/build
./ta/sha_perf/build
./ta/aes_perf/build
./ta/rpc_test/build
./ta/sims/build
./ta/concurrent/build
./host/xtest/build
```

contain the generated files.  However looking at the `Makefile`, it seems to be under the impression everything is going in `./out`.

It needs some adaptation, or I am simply driving it wrong?

What is the correct way to make `make clean` do something useful?

etienne-lms:
@lws-team, optee_test builds default go to the `./out` but this path can be define from env (`O=...`) so in the end, it's up to the build env to define where things are built.

I can't see any occurrence of a `build` directory in the optee_test. Could it be from your build setup ? Did you really go through the optee_test root `Makefile` and its `ta/Makefile` to build the optee_test TAs ? These rather use pathes like `<some_build_path>/ta/<ta_dir_name>/`.

lws-team:
I coordinate my meta-build from a toplevel project where the pieces are git submodules, and a toplevel makefile that does a `make -C subdir ...` to do stuff.

The actual build command in my toplevel Makefile looks like

```
make -C optee_test CROSS_COMPILE_HOST=aarch64-linux-gnu- \
                CROSS_COMPILE_TA=aarch64-linux-gnu- \
                TA_DEV_KIT_DIR=`pwd`/optee_os/out/optee_os_out/arm-plat-hikey/export-ta_arm64 O=build
```

So that is where "build" is coming from.

[...]

PR #163, opened 2017-01-06
fix "make clean" against directories generated from "make"

etienne-lms:
This fixes the cleaning of build directories for both absolute and relative pathes in directive `O`.

This is based on on-going work for https://github.com/OP-TEE/optee_os/pull/1275.

This attempts to fix https://github.com/OP-TEE/optee_test/issues/162.


ta/Makefile
```diff
@@ -44,8 +44,12 @@ all: ta
...
-	$(q)$(foreach dir,$(TA_DIRS), $(MAKE) -C $(dir) O=$(O)/$(dir) &&) true
+	$(q)$(foreach dir,$(TA_DIRS), $(MAKE) -C $(dir) O=$(O)/ta/$(dir) &&) true
 
 .PHONY: clean
 clean:
-	$(q)$(foreach dir,$(TA_DIRS), $(MAKE) -C $(dir) O=$(O)/$(dir) $@ &&) true
+	$(q)$(foreach dir,$(TA_DIRS),$(MAKE) -C $(dir) O=$(O)/ta/$(dir) $@ &&) true
+	$(q)$(foreach dir,$(TA_DIRS), bash -c 'cd $(dir) && \
+						[ ! -d $(O)/ta ] || rmdir $(O)/ta' &&) true
+	$(q)$(foreach dir,$(TA_DIRS), bash -c 'cd $(dir) && \
```
etienne-lms: This does not work when `O` is a relative path with subdirectory/ies.

Makefile
```diff
@@ -1,4 +1,7 @@
 ifneq ($O,)
+ifneq ($(shell echo $(O) | grep -- "/\.$$"),)
+$(error Cannot support output directory 'O' path ending with terminal subpath info "./")
+endif
```
jforissier: Your complaining about `./` but you're testing against `/.`, what is the correct one?
Also the test could be simplified to:
```
ifeq ($(patsubst %/.,,$(O)),)
```

Makefile
```diff
@@ -40,14 +43,19 @@ xtest:
...
 
 .PHONY: clean
 ifneq ($(wildcard $(TA_DEV_KIT_DIR)/host_include/conf.mk),)
 clean:
 	$(q)$(MAKE) -C host/xtest O=$(out-dir)/xtest q=$(q) $@
-	$(q)$(MAKE) -C ta O=$(out-dir)/ta q=$(q) $@
+ifneq ($O,)
+	$(q)bash -c 'cd host/xtest && [ ! -d $(O) ] || \
+			[ `stat --format=%i $(O)` == `stat --format=%i .` ] || \
+			rmdir --ignore-fail-on-non-empty  $(O)'
```
jforissier: Since we basically want to ignore any error, the whole `ifneq` block could be simplified to:
```
        rmdir $(O) 2>/dev/null; true
```

Makefile
etienne-lms: ok, i must fix this. 
the thing is: rmdir does not like pathes as some/path/. or /some/path/./.
from your proposal below, we can even discard the restriction above.

Makefile
etienne-lms: ok for `rmdir $(O) 2> /dev/null`, but we must keep the ugly `cd host/xtest` as $(O) might be a relative part. agree ? any nicer proposal (could check if $(O) is relative, or hide complexity in a macro, but i don't feel it would be useful).

Makefile
etienne-lms: i'm wrong: `rmdir` still needs a path string not ending with `/.` or `/./`

Makefile
```diff
@@ -1,8 +1,14 @@
...
+# Remove trailing / and /. in output path, rmdir does not like them
+out-dir := $(shell s=$O; while [ true ]; do \
+            [ "$$s" == "/" ] && break; \
+            s=`echo "$$s" | sed s\/\\\/$$\/\/ --`; \
+            [ "$$s" == "/" ] && break; \
+            s=`echo "$$s" | sed s\/\\\/\.$$\/\/ --`; \
+            echo $$s | grep --quiet /$$ && continue; \
+            echo $$s | grep --quiet /\.$$ && continue; \
+	    break; \
+        done; echo $$s)
```
etienne-lms: this is a bit ugly, indeed.

host/xtest/Makefile
```diff
@@ -4,6 +4,17 @@
...
+out-dir := $(shell s=$O; while [ true ]; do \
+            [ "$$s" == "/" ] && break; \
+            s=`echo "$$s" | sed s\/\\\/$$\/\/ --`; \
+            [ "$$s" == "/" ] && break; \
+            s=`echo "$$s" | sed s\/\\\/\.$$\/\/ --`; \
+            echo $$s | grep --quiet /$$ && continue; \
+            echo $$s | grep --quiet /\.$$ && continue; \
+	    break; \
+        done; echo $$s)
+
```
jforissier: Put the macro in a common place instead?

host/xtest/Makefile
```diff
@@ -136,16 +147,19 @@ LDFLAGS += -lpthread -lm
...
 .PHONY: clean
 clean:
-	@echo '  CLEAN $(O)'
-	$(q)rm -f $(O)/xtest
+	@echo '  CLEAN $(out-dir)'
+	$(q)rm -f $(out-dir)/xtest
 	$(q)$(foreach obj,$(objs), rm -f $(obj))
+	$(q)rmdir $(out-dir)/adbg/src 2> /dev/null; true
+	$(q)rmdir $(out-dir)/adbg 2> /dev/null; true
+	$(q)rmdir $(out-dir) 2> /dev/null; true
```
jforissier: Define a macro (`rmdir`) then just `$(call rmdir, $(out-dir)/adbg/src $(out-dir)/adbg $(out-dir))`

ta/Makefile
```diff
@@ -44,8 +54,10 @@ all: ta
...
 ta: 
-	$(q)$(foreach dir,$(TA_DIRS), $(MAKE) -C $(dir) O=$(O)/$(dir) &&) true
+	$(q)$(foreach dir,$(TA_DIRS), $(MAKE) -C $(dir) O=$(out-dir)/ta/$(dir) &&) true
 
 .PHONY: clean
 clean:
-	$(q)$(foreach dir,$(TA_DIRS), $(MAKE) -C $(dir) O=$(O)/$(dir) $@ &&) true
+	$(q)$(foreach dir,$(TA_DIRS), $(MAKE) -C $(dir) O=$(out-dir)/ta/$(dir) $@ &&) true
+	$(q)$(foreach dir,$(TA_DIRS), bash -c 'cd $(dir) && rmdir $(out-dir)/ta 2> /dev/null';) true
+	$(q)$(foreach dir,$(TA_DIRS), bash -c 'cd $(dir) && rmdir $(out-dir) 2> /dev/null';) true
```
jforissier: You can probably replace `bash -c 'foo'` with `(foo)`.

jenswi-linaro: Just make sure that `$(SHELL)` is initialized properly, everyone doesn't use bash as their normal shell.

Makefile
etienne-lms: nice macro.

> (note: if O=/ or O=, $(out-dir) will be empty, do we need to handle those cases?)

if O is empty, out-dir is set through another sequence, so it's ok.
if O=/ (or fancy /././///./.), then yes, we must handle the case. I would propose to simply reject empty `$out-dir` contain. I think we don't expect to build straight in OS root directory.

Makefile
```diff
@@ -40,14 +49,15 @@ xtest:
...
+			  O=$(out-dir) \
 			  $@
 
 .PHONY: clean
 ifneq ($(wildcard $(TA_DEV_KIT_DIR)/host_include/conf.mk),)
 clean:
 	$(q)$(MAKE) -C host/xtest O=$(out-dir)/xtest q=$(q) $@
-	$(q)$(MAKE) -C ta O=$(out-dir)/ta q=$(q) $@
+	cd host/xtest && rmdir $(O) 2> /dev/null; true
+	$(q)$(MAKE) -C ta O=$(out-dir) q=$(q) $@
```
etienne-lms: this was not needed for xtest so i discarded the equivalent change.
If you feel it is more coherent, then ok.

host/xtest/Makefile
etienne-lms: you mean adding a output_dir.mk in optee_xtest and including from each concerned Makefile in optee_test ? why not.

host/xtest/Makefile
etienne-lms: macro are nice, but will it simply maintenance here ?

ta/Makefile
etienne-lms: ok, i'll fix that.

etienne-lms:
Squashed and rebased (all comments seemed addressed).
Added a new change to fix 'clean' rule when cleaning/building with/without GP test suite.

ta/Makefile
```diff
@@ -44,8 +47,15 @@ all: ta
 
 .PHONY: ta
 ta: 
-	$(q)$(foreach dir,$(TA_DIRS), $(MAKE) -C $(dir) O=$(O)/$(dir) &&) true
+	$(q)$(foreach dir,$(TA_DIRS), $(MAKE) -C $(dir) O=$(out-dir)/ta/$(dir) &&) true
+
+# remove build directories including ta/<ta-name>/ directories.
+# Note: $(out-dir) may be is relative path.
+define rm-build-subdir
+echo `[ -d $1 ] && { cd $1; rmdir $(out-dir)/ta $(out-dir) 2> /dev/null; true; }` > /dev/null
```
etienne-lms: this `rmdir` should be replaced with a `rmdir --ignore-fail-on-non-empty `

host/xtest/Makefile
```diff
@@ -134,16 +137,23 @@ LDFLAGS += -lpthread -lm
...
-	$(q)mkdir -p $(O)/adbg/src
+$(out-dir)/xtest/%.o: $(CURDIR)/%.c
+	$(q)mkdir -p $(out-dir)/xtest/adbg/src
 	@echo '  CC      $<'
 	$(q)$(CC) $(CFLAGS) -c $< -o $@
 
+define rm-build-dirs
+	$(q)for d in $1; do rmdir $(out-dir)/xtest/$$d 2> /dev/null; true; done
+	$(q)rmdir $(out-dir)/xtest 2> /dev/null; true
+	$(q)rmdir $(out-dir) 2> /dev/null; true
```
etienne-lms: `rmdir --ignore-fail-on-non-empty`

ta/Makefile
```diff
@@ -44,8 +47,15 @@ all: ta
 
 .PHONY: ta
 ta: 
-	$(q)$(foreach dir,$(TA_DIRS), $(MAKE) -C $(dir) O=$(O)/$(dir) &&) true
+	$(q)$(foreach dir,$(TA_DIRS), $(MAKE) -C $(dir) O=$(out-dir)/ta/$(dir) &&) true
+
+# remove build directories including ta/<ta-name>/ directories.
+# Note: $(out-dir) may be is relative path.
```
etienne-lms: s/is/a/

etienne-lms:
fixed, rebased. all comment addressed.
review opened.

etienne-lms:
Are we interested in these changes ?
If not, I will discard the 1st commit and push back the second as a specific PR.

scripts/common.mk
```diff
@@ -0,0 +1,9 @@
+# Makefile script utilisies used in various OP-TEE test components
```
etienne-lms: done

etienne-lms:
comments addressed, rebased, tags applied.

PR #164, opened 2017-01-10
xtest: remove deprecated USER_SPACE

etienne-lms:
Legacy USER_SPACE enables thread support in xtest. Withdraw as always
set.


jforissier:

etienne-lms:
tag applied

PR #166, opened 2017-01-20
Add AES CTS test case

jforissier:
Add test case for issue https://github.com/OP-TEE/optee_os/issues/1203.


jforissier:
Depends on https://github.com/OP-TEE/optee_os/pull/1301.

host/xtest/xtest_4000.c
```diff
@@ -2000,6 +2000,34 @@ static const uint8_t ciph_data_aes_cts_vect6_ctx[] = {
 	0x3b, 0xc1, 0x03, 0xe1, 0xa1, 0x94, 0xbb, 0xd8,
 };
 
+/* Test case for https://github.com/OP-TEE/optee_os/issues/1203 */
+
+#define ciph_data_aes_cts_issue1203_key ciph_data_aes_cts_vect1_key
+#define ciph_data_aes_cts_issue1203_iv ciph_data_aes_cts_vect1_iv
+
+static const uint8_t ciph_data_aes_cts_issue1203_ptx[80] = {
+	0x00
```
jenswi-linaro: The plaintext could be be a little more complicated to be able to detect if bytes are swapped or similar.

jforissier: Other test cases should detect that kind of problem, but I can just use random bytes as well. Will update.

jenswi-linaro: Please do, because in this case we're obviously triggering a different code path compared to the other test cases.

jenswi-linaro:

PR #161, opened 2017-01-05
host: xtest_10000: remove unused static array

lws-team:
Newer (Linaro) compilers complain

```
.../optee_test/host/xtest/xtest_10000.c:300:22: error: 'pbkdf2_4_dkm' defined but not used [-Werror=unused-const-variable=]
```

and break the build.

Signed-off-by: Andy Green <andy@warmcat.com>

PR #168, opened 2017-02-03
Fix issue with pbkdf2_4_dkm not defined

jbech-linaro:
Resolves the same issue as #161, but keeping the variable.

### DIFF
--- a/host/xtest/xtest_10000.c
+++ b/host/xtest/xtest_10000.c
@@ -280,7 +280,7 @@ static const uint8_t pbkdf2_3_dkm[] = {
 #define pbkdf2_4_password pbkdf2_1_password
 #define pbkdf2_4_salt pbkdf2_1_salt
 #define pbkdf2_4_iteration_count 16777216
-static const uint8_t pbkdf2_4_dkm[] = {
+static const uint8_t pbkdf2_4_dkm[] __maybe_unused = {
 	0xee, 0xfe, 0x3d, 0x61, 0xcd, 0x4d, 0xa4, 0xe4,
 	0xe9, 0x94, 0x5b, 0x3d, 0x6b, 0xa2, 0x15, 0x8c,
 	0x26, 0x34, 0xe9, 0x84


jenswi-linaro:

jbech-linaro:
Squashed, rebased and tag(s) applied (should be) ready for merge! Thanks!

host/xtest/xtest_10000.c
```diff
@@ -280,7 +280,7 @@ static const uint8_t pbkdf2_3_dkm[] = {
 #define pbkdf2_4_password pbkdf2_1_password
 #define pbkdf2_4_salt pbkdf2_1_salt
 #define pbkdf2_4_iteration_count 16777216
-static const uint8_t pbkdf2_4_dkm[] = {
+static const uint8_t pbkdf2_4_dkm[] __maybe_unused = {
```
jforissier: Does this build proberly? Where is this `__maybe_unused` defined?

jbech-linaro: Builds, yes, absolutely, both on gcc5.x and [gcc6.x](https://travis-ci.org/jbech-linaro/build). But, now when you're saying it, I'm also a bit puzzled where it comes from. I took for granted that we had that ready to be used. When I compile the file and look at the pre-processor output (`-E -dM`). I see
```c
...
#define __maybe_unused __attribute__((unused)) 
...
```
It's used in [other places](https://github.com/OP-TEE/optee_test/search?utf8=%E2%9C%93&q=__maybe_unused) to under xtest. But no define and I cannot find anything about it when grepping in the toolchain folder. Weird! Maybe we should define it in optee_test somewhere? `xtest_test.h`? Or some new common h-file shared with both the TA's and the host.

jenswi-linaro: `compiler.h` is included somewhere.

jbech-linaro: Actually leaving out `-dM` reveals (gcc5.x here, same on 6.x)
```c
# 1 "/mnt/sshd/devel/optee_projects/qemu/build/../optee_os/out/arm/export-ta_arm32/host_include/compiler.h" 1                                                                                
# 29 "/mnt/sshd/devel/optee_projects/qemu/build/../optee_os/out/arm/export-ta_arm32/host_include/compiler.h"                                                                                 
#define COMPILER_H                                                                                           
...
#define __maybe_unused __attribute__((unused))
```

So it comes from:
`./optee_os/out/arm/export-ta_arm32/host_include/compiler.h`

jforissier: Ah, it's OK. `gcc -E` shows that the include sequence is `xtest_helpers.h` -> `tee_api_types.h` -> `compiler.h`. So it comes by default in the Linux side of `xtest` thanks to the `host_include` folder of the TA devkit. A "regular" client application accessing only the `libteec` API would not have it however.
As for the TA side, it can include `<compiler.h>` or `<tee_api_types.h>`, that's included in the TA devkit.

jforissier: ;)

Yep, basically the "host" (non-secure) part of `xtest` has full access to the TA API, which I don't think is a good thing (i.e., `#include <tee_internal_api.h>` makes no sense). Some trimming of what we're putting in `host_include` would be needed I suppose.

PR #171, opened 2017-02-17
Update 20xxx test for hash tree

jenswi-linaro:
Needed by https://github.com/OP-TEE/optee_os/pull/1360

### DIFF
--- a/host/xtest/xtest_20000.c
+++ b/host/xtest/xtest_20000.c
@@ -26,6 +26,7 @@
 #include <xtest_helpers.h>
 
 #include <tee_fs_key_manager.h>
+#include <fs_htree.h>
 #include <tee_client_api.h>
 #include <tee_api_defines_extensions.h>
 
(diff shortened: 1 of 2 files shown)


jforissier:
It looks like the "very large secure storage" test is a bit too much for RPMB FS (at least, with the size we're emulating in tee-supplicant).

jenswi-linaro:
I'm making a special case for RPMB FS.

host/xtest/xtest_6000.c
```diff
@@ -1731,9 +1731,19 @@ static void xtest_tee_test_6018_single(ADBG_Case_t *c, uint32_t storage_id)
 	uint32_t obj;
 	uint32_t orig;
 	uint8_t block[32 * 1024];
-	size_t num_blocks = 50;
+	size_t num_blocks;
+	size_t block_size;
 	size_t n;
 
+	if (storage_id == TEE_STORAGE_PRIVATE_RPMB) {
```
jforissier: The test has to be a bit more complicated, to handle the case when `storage_id == TEE_STORAGE_PRIVATE` and the REE FS is not compiled in (hence, the second FS in order or priority is selected by the TEE core, and that is RPMB). Here is what I mean:
```C
static uint32_t fs_id_for_tee_storage_private()
{
#ifdef CFG_REE_FS
        return TEE_STORAGE_PRIVATE_REE;
#else
#ifdef CFG_RPMB_FS
        return TEE_STORAGE_PRIVATE_RPMB;
#else
#ifdef CFG_SQL_FS
        return TEE_STORAGE_PRIVATE_SQL;
#endif
#endif
#endif
}

static uint32_t real_id_for(uint32_t id)
{
        if (id == TEE_STORAGE_PRIVATE)
                return fs_id_for_tee_storage_private();
        return id;
}

static bool storage_is(uint32_t id1, uint32_t id2)
{
        return (real_id_for(id1) == real_id_for(id2));
}

static void xtest_tee_test_6018_single(ADBG_Case_t *c, uint32_t storage_id)
{
        /* ... */

        if (storage_is(storage_id, TEE_STORAGE_PRIVATE_RPMB)) {
                /* RPMB FS is a bit resource constrained */
                /* ... */
        }
```

jenswi-linaro:
Updated, tags applied.

host/xtest/xtest_20000.c
```diff
@@ -460,93 +529,99 @@ static void storage_corrupt(ADBG_Case_t *c,
...
+			if (!ADBG_EXPECT(c, TEE_SUCCESS,
+				obj_corrupt(&uuid, filename,
+					    ARRAY_SIZE(filename),
+					    offset, file_type,
+					    tv->block_num, vers)))
+				goto exit;
+
+			/*
+			 * All nodes are currently verified when opening so
+			 * any corrupt node will be depected. This will
```
jforissier: detected

jenswi-linaro: Thanks, updating

jenswi-linaro:
Update

jforissier:

jenswi-linaro:
Tag applied

PR #172, opened 2017-02-21
aes-perf, sha-perf: display more statistics

jforissier:
Show more statistics to help assess the test repeatability and compare
test runs:
- Print the coefficient of variation (cv = stddev/mean)
- When the -v option is given: also show the 2-sigma interval


jbech-linaro:

PR #176, opened 2017-03-07
xtest_20000: Fix uninitialized variable warning

glneo:
When compiled with -Wuninitialized, GCC identifies a false positive
uninitialized variable 'real_offset'. Although we know the default
path will not be taken, GCC does not, for good measure lets just pass
back an error to fix the warning.

Signed-off-by: Andrew F. Davis <afd@ti.com>

### DIFF
--- a/host/xtest/xtest_20000.c
+++ b/host/xtest/xtest_20000.c
@@ -344,6 +344,8 @@ static void get_offs_size(enum tee_fs_htree_type type, size_t idx,
 		sz = BLOCK_SIZE;
 		break;
 	default:
+		*offs = -1UL;
+		sz = -1UL;
 		break;
 	}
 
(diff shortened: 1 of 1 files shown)


jenswi-linaro:

host/xtest/xtest_20000.c
```diff
@@ -384,6 +386,11 @@ static TEEC_Result obj_corrupt(TEEC_UUID *p_uuid, void *file_id,
 			 ta_dirname, obj_filename);
 
 		get_offs_size(type, block_num, version, &real_offset, NULL);
+		if (real_offset == -1UL) {
```
etienne-lms: minor: what about adding a return value to `get_offs_size()` (i.e: 0=success, non-0=error). It would be more easy to read/maintain. this is the most common error case detection scheme.

glneo: Works for me, I made a new pull request for this here: #179 

Pull whichever version you all like better, then I'll close the other.

PR #185, opened 2017-03-21
xtest_1000: Make fake and corrupt TAs readable by others

vchong:

### DIFF
--- a/host/xtest/xtest_1000.c
+++ b/host/xtest/xtest_1000.c
@@ -11,6 +11,7 @@
  * GNU General Public License for more details.
  */
 
+#include <errno.h>
 #include <limits.h>
 #include <pthread.h>
 #include <stdio.h>
@@ -440,9 +441,19 @@ static FILE *open_ta_file(const TEEC_UUID *uuid, const char *mode,
 			  bool for_write)
 {
 	char buf[PATH_MAX];
+	FILE *ret;
 
 	uuid_to_full_name(buf, sizeof(buf), uuid, for_write);
-	return fopen(buf, mode);
+	ret = fopen(buf, mode);
+	if (ret && for_write)
+		if (chmod(buf, 0666)) {
+			fprintf(stderr, "chmod(\"%s\"): %s",
+					buf, strerror(errno));
+			printf("chmod error! This test will fail if " \
+				"tee-supplicant does not have permission " \
+				"to load this TA file!\n");
+		}
+	return ret;
 }
 
 static bool rm_file(const TEEC_UUID *uuid)


vchong:
If for any (stupid?) reason xtest is run with some higher permission that tee-supp doesn't have, then supp will not have permission to access the fake and corrupted TAs so 1008.4 and 1008.5 will fail. This patch will avoid this problem, but admittedly, this is probably a very rare case. If we deem that this is too rare and can be ignored, then we can just abandon this patch.

host/xtest/xtest_1000.c
```diff
@@ -440,9 +441,19 @@ static FILE *open_ta_file(const TEEC_UUID *uuid, const char *mode,
 			  bool for_write)
 {
 	char buf[PATH_MAX];
+	FILE *ret;
 
 	uuid_to_full_name(buf, sizeof(buf), uuid, for_write);
-	return fopen(buf, mode);
+	ret = fopen(buf, mode);
+	if (ret && for_write)
+		if (chmod(buf, 0666)) {
```
jforissier: Why not set mode to 0666 in `fopen()` instead?

host/xtest/xtest_1000.c
```diff
@@ -440,9 +441,19 @@ static FILE *open_ta_file(const TEEC_UUID *uuid, const char *mode,
...
+			fprintf(stderr, "chmod(\"%s\"): %s",
+					buf, strerror(errno));
+			printf("chmod error! This test will fail if " \
+				"tee-supplicant does not have permission " \
+				"to load this TA file!\n");
```
jforissier: You're mentioning read permission by `tee-supplicant` but the subject of the commit is "Make xxx writable". I don't understand.

jforissier:
In my opinion `xtest` should not try to fix an invalid setup. There are so many other things that can be setup wrong. Just report errors in a useful manner as they occur and it's enough.

vchong:
> What's the problem? Why does this change fix it?

If say xtest is run as root, it will create the fake and corrupted TA with mode 0600. If tee-supplicant is not root, then it will not be able to load the TAs and fail.

> In my opinion xtest should not try to fix an invalid setup. There are so many other things that can be setup wrong. Just report errors in a useful manner as they occur and it's enough.

Yes, if we deem running xtest as root and tee-supplicant as non-root as an invalid setup, then we can just abandon this.

host/xtest/xtest_1000.c
vchong: That was my first thought too but looking it up, it seems like `fopen()` can only do the `r, w, r+`, etc stuffs?

host/xtest/xtest_1000.c
vchong: You're right. Strictly speaking we only need 0644, not 0666. Was maybe overdoing the future-proofing. Can change `writable` to `readable` too if necessary.

jenswi-linaro:
It sounds like umask is wrong for xtest, with a normal umask of 022 it should create file that are 0644.

jenswi-linaro:
Before you start xtest at the prompt, what does umask say?

vchong:
> Before you start xtest at the prompt, what does umask say?

077

jenswi-linaro:
There's the problem. I'm not that happy about to hard code an umask 022 in xtest since it's a matter of the environment. Perhaps it's acceptable to just update the instructions for how xtest should be launched in Android.

vchong:
Thanks! Sounds like something like `umask 22 && xtest && umask 77` would work.

jenswi-linaro:
Or even `(umask 022 && xtest)`, the `()` are important.

PR #186, opened 2017-03-23
ta/Makefile: include host_include/conf.mk

jforissier:
The compile-time options of OP-TEE OS (CFG_*) are currently not
available from the main TA makefile. Fix this, since we already
have a need (the sdp_basic TA should automatically be selected
when OP-TEE supports the Secure Data Path).

### DIFF
--- a/ta/Makefile
+++ b/ta/Makefile
@@ -8,6 +8,8 @@ ifeq ($(out-dir),)
 $(error invalid output directory (O=$(O)))
 endif
 
+include $(TA_DEV_KIT_DIR)/host_include/conf.mk
+
 # Prevent use of LDFLAGS from the environment. For example, yocto exports
 # LDFLAGS that are suitable for the client applications, not for TAs
 LDFLAGS=


ta/Makefile
```diff
@@ -8,6 +8,8 @@ ifeq ($(out-dir),)
 $(error invalid output directory (O=$(O)))
 endif
 
+-include $(TA_DEV_KIT_DIR)/host_include/conf.mk
```
jenswi-linaro: Does it make sense to silently ignore this file if it's missing? The rest of TA devkit is needed in order to be able to compile any way.

jforissier: Well, it makes *some* sense, but not that much actually ;)
If `conf.mk` is missing, one can still pass the config options on the command line for instance or export them from a higher-level Makefile (I suppose that's why @etienne-lms could build the SDP app without this patch!). But I see no circumstance where `host_include` or `conf.mk` would not be available in the devkit when one is building `optee_test`, so yeah, better remove the dash so that any issue is detected early.

jenswi-linaro:

etienne-lms:

PR #190, opened 2017-03-30
GP: do not expect to find non-standard macros in tee_internal_api.h

jforissier:
With [1], SLogTrace(), SLogError(), SLogWarning() and S_VAR_NOT_USED()
are not defined in tee_internal_api.h anymore. Define them locally.

[1] https://github.com/OP-TEE/optee_os/pull/1449


jenswi-linaro:

PR #191, opened 2017-03-30
GP: clean up the compilation messages

jforissier:
When the GlobalPlatform test suite is enabled, the build produces
*tons* of warnings. The build also prints some useless messages and
is silent on many operations.

This patch improves the situation by doing the following:
- Fix all the compiler warnings I found with GCC 4.9 (32- and 64-bit
  builds, for both TAs and xtest). Unused parameters, use of size_t
  when uint32_t is expected, constness of the TEE_GetInstanceData()
  parameter, copying a (64-bit) handle into a 32-bit value
  (params[x].value.a) and possibly a few others I don't remember.
- Remove "INFO:" messages
- Use the Kbuild format ("  OPER    path/to/target/file") to trace
  most build steps
Some more cleanup is required to avoid rebuilding stuff each time,
but that can be addressed later.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,18 +81,9 @@ GP_XTEST_IN_DIR=${GP_XTEST_OUT_DIR}/global_platform/${CFG_GP_API}
 GP_USERTA_DIR=$(CURDIR)/ta
 
 define patch-file
-	@if [ ! -e ${1} ]; then \
-		echo "Error: File to patch is unknown: $1"; \
-		return 1; \
-	fi
-	@if [ ! -e ${2} ]; then \
-		echo "Error: Patch to apply is unknown: $2"; \
-		return 1; \
-	fi
 	@if [ ! -e ${1}.orig ]; then \
-		patch -N -b ${1} < ${2}; \
-	else \
-		echo "Warning: Patch already applied on `basename $1`"; \
+		echo "  PATCH   ${1}"; \
+		patch -s -N -b ${1} < ${2}; \
 	fi
 endef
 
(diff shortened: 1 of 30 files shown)


jforissier:
Here is a sample build log with this patch applied: https://pastebin.linaro.org/view/raw/bc1cd02e
Without this patch, the build log is 1.2M due to all the warnings (!).

jenswi-linaro:
I see how it's useful to clean up the .xml related code. But patching the actual TA source could cause problems when updating but also when compliance testing.

jforissier:
@jenswi-linaro I totally agree but this is nothing new... See the [patch-package](https://github.com/OP-TEE/optee_test/blob/master/Makefile#L167) target, several .c and .h source files are patched already. I'm just patching a bit more...

jenswi-linaro:
If the objective is to reduce warnings from TA, wouldn't it be more efficient to just turn of as many warnings as possible? The more the TAs are patched the further they divert from the originals.

jforissier:
True. It is certainly possible to turn off warnings for the GP TAs only. I'll try that.

etienne-lms:
Aside from the discussion on whether this change is to be merged or not, here's my feedback:
- nice xtest+gp build without spurious warnings: fully clean trace in my b2260 build  env.
- `Tested-by: Etienne Carriere <etienne.carriere@linaro.org> (b2260)`

jforissier:
@etienne-lms thanks for the feedback. I still think I can achieve the same result in a simpler way, as suggested by Jens. Stay tuned ;)

etienne-lms:
Removing wraning is nice, but preventing them is nicer I think.
As we (you?) master the patches we apply, I think we can still get such changes under control.

jenswi-linaro:
If you're doing compliance testing with changed TAs, what are you testing? If you want to prevent the warnings I guess you need to talk to Nicolas Lavabre.

etienne-lms:
@jenswi-linaro: nice hit !

PR #192, opened 2017-03-30
[RFC] Introduce test name wildcards

jforissier:
When a test ID is preceded with the -w switch, parse it as a shell
wildcard using the fnmatch() function. This matching is more flexible
than the default substring match, because it allows for instance to run
the tests in the range 2000-2999 without running the tests in the
range 20000-29999.

  xtest -w "*2???"

### DIFF
--- a/host/xtest/adbg/src/adbg_run.c
+++ b/host/xtest/adbg/src/adbg_run.c
@@ -15,6 +15,7 @@
  * 1. Includes
  ************************************************************************/
 #include "adbg_int.h"
+#include <fnmatch.h>
 
 /*************************************************************************
  * 2. Definition of external constants and variables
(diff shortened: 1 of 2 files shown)


jenswi-linaro:
I'd rather have unique prefixes to the test cases instead to keep it simple. It's unfortunate that the 10xxx and 20xxx case where introduced, instead they should be renamed as we discussed in Budapest.

jforissier:
This PR addresses an issue we have with tests in the 2k and 20k ranges. It is currently impossible to run only the 2k tests.

@jenswi-linaro I know this is not the option you preferred when we discussed this in Budapest, but I still think it is the less intrusive and most backward-compatible.
The other option we discussed was moving the GP tests to their own test suite (`xtest -t gp`) to free some ranges and re-number the non-GP tests larger than 10000.

jforissier:
;) we posted at the same time

jenswi-linaro:
Yes I see that. :-)
Well, I still disagree too...

jenswi-linaro:
My main objection against wildcarding is that I would need to type more to match whatever I'd like to match compared to unique prefixes even if they are a bit less powerful. Either I'm interested in running one test case, a family of test cases or all of them.

But I think that the GP test belongs in their own test suite also now that we have several suites.

etienne-lms:
I think both changes are compatible. We can still move the GP tests to a "gp" test suite, and
Remains the decision whereas 10xxx and 20xxx should be renamed to other IDs.

Jerome's change is not intrusive as he says, why not keeping it.

jenswi-linaro:
@etienne-lms, true we can do both. In that case I have no objections. :-)

jforissier:
OK I'll create another PR to move the GP tests to their own suite because that makes sense in any case. Then, to address the 2xxx vs. 20xxx issue, either we rename the tests in the slots freed by the GP prefix, or we use this patch, or both as you said ;)

jenswi-linaro:

jforissier:
> OK I'll create another PR to move the GP tests to their own suite because that makes sense in any case.

https://github.com/OP-TEE/optee_test/pull/193

PR #193, opened 2017-03-31
Move GlobalPlatform tests to their own test suite

jforissier:
Introduce a new test suite, selected by "-t gp", and move all the
GlobalPlatform tests there.
In order to maintain some backward compatibility, xtest will run both
the "regression" and "gp" suites in a row by default. To implement this
behavior, accept a "+" as a separator for test suite names, and set the
default test suite name to "regression+gp".


host/xtest/xtest_main.c
```diff
@@ -117,15 +129,29 @@ int main(int argc, char *argv[])
...
-	else {
-		fprintf(stderr, "No test suite found: %s\n", test_suite);
-		ret = -1;
+	for (token = test_suite; ; token = NULL) {
+
+		token = strtok(token, "+");
+		if (!token)
+			break;
+
+		if (strcmp(token, "regression") == 0)
```
etienne-lms: minor style: could use `!strcmp()` for consistency with the use of `strcmp()` above (around line 90).

host/xtest/xtest_main.c
```diff
@@ -44,7 +51,11 @@ void usage(char *program)
 	printf("options:\n");
 	printf("\t-d <device-type>   default not set, use any\n");
 	printf("\t-l <level>         test suite level: [0-15]\n");
-	printf("\t-t <test_suite>    available test suite: regression, benchmark\n");
+	printf("\t-t <test_suite>    available test suites: regression benchmark");
+#ifdef WITH_GP_TESTS
+	printf(" gp");
+#endif
+	printf("\n");
```
etienne-lms: Could you add a word about the magic `+` ?

jforissier: Sure

host/xtest/xtest_main.c
jforissier: OK

jenswi-linaro:
What about changed file names and .gitignores?

jforissier:
> What about changed file names and .gitignores?

Huh?

jenswi-linaro:
The generated files, will they still use the old file names, I'm mainly thinking of xtest_8000.c and xtest_9000.c

jforissier:
Yes, why change since the GP test numbers are not changed by this commit?
Would you like me to rename all the regression test files like `reg_1000.c` etc. and the GP tests `gp_7000.c` etc? I can add a commit for that.

jenswi-linaro:
Yes, please, that would be nice.

host/xtest/Makefile
```diff
@@ -28,33 +28,32 @@ ifdef CFG_GP_PACKAGE_PATH
...
 	adbg/src/adbg_run.c \
-	adbg/src/security_utils_hex.c
+	adbg/src/security_utils_hex.c \
+	aes_perf.c \
+	benchmark_1000.c \
+	benchmark_2000.c \
+	regression_4000.c \
+	regression_5000.c \
+	regression_6000.c \
+	regression_10000.c \
```
jenswi-linaro: `regression_7000.c` is missing.

host/xtest/Makefile
```diff
@@ -28,33 +28,32 @@ ifdef CFG_GP_PACKAGE_PATH
 GP := _gp
```
jenswi-linaro: This is probably not needed any more.

jforissier:
Updated.

.gitignore
```diff
@@ -6,11 +6,11 @@ GTAGS
 out/*
 ta/GP_*
 host/xtest/global_platform/
-host/xtest/xtest_7000_gp.c
-host/xtest/xtest_7500.c
-host/xtest/xtest_8000.c
-host/xtest/xtest_8500.c
-host/xtest/xtest_9000.c
+host/xtest/gp_7000_gp.c
```
jenswi-linaro: The new name is `host/xtest/gp_7000.c`, isn't it?

jenswi-linaro:

etienne-lms:
@jforissier, can you check the test. Running `xtest` without arguments ran only the GP suite on my setup. As I don't master `strtok()` , I cannot tell whether you implemented it right or not.

etienne-lms:
note: i ran  `xtest -t gp+gp` and i launched the gp test twice, as expected.

etienne-lms:
my fault: running `xtest` alone, when built with GP, works nice: both suites are launched.
Your use of `strtok()` works well.
The issue is that one must dig into the log history to get the early suite result summary...  I must say that it is not very convenient.

jforissier:
@etienne-lms you're right, I'm not really happy about this either, but I could not find a reasonable solution (I mean one that would not involve changing too much code).

etienne-lms:
Do you think it is possible to plit `Do_ADBG_RunSuite()` into a `Do_ADBG_RunSuite()` and a `Do_ADBG_ReportSuite()` so that the former runs a test suite and the later only prints the result summary.
This way we could first loop on the target suites for `Do_ADBG_RunSuite()`, and once all done loop again on target suites for `Do_ADBG_RunSuite()` to get a full summary at the end.

jforissier:
Yep that would be an improvement but still would not give fully consolidated results (number of failures etc.).
It would probably be better to merge the test suites into a single `ADBG_Suite_Definition_t`, then run `Do_ADBG_RunSuite()` on that. Another advantage is that we could detect invalid suite names before running the first test.

jforissier:
Updated. Here is what it looks like.

```
root@Vexpress:/ xtest _7001
Test ID: _7001
Run test suite with level=0

#
# regression+gp
#
 
* regression_7001 Allocate_In RELEASE_SHARED_MEMORY_WHEN_ALLOCATED
  regression_7001 OK
 
* gp_7001 29-84-6d
  gp_7001 OK
Result of testsuite regression+gp filtered by "_7001":
regression_7001 OK
gp_7001 OK
18 subtests of which 0 failed
2 test cases of which 0 failed
733 test cases was skipped
TEE test application done!
```

I should probably move the function `move_suite()` to the ADBG framework, though.

jenswi-linaro:
Looks good. Yes please move the function into the ADBG framework. May I suggest to rename the function to `Do_ADBG_AppedToSuite()` also to harmonize better with the rest of the API.

jforissier:
Updated. Once reviewed, I'd like to squash all commits into one, because I don't think it makes much sense to have a "rename files" commit. Let me know if you prefer some kind of different split.

jenswi-linaro:
LGTM

etienne-lms:
This is nice and work well: we get a nice test summary.

The only drawback (I could not say 'regression') is that some tests are duplicated.
The old xtest_7000.c used to be replaced with the xtest_7000_gp.c.
Now we have both regression_7000.c and gp_7000.c and the tests from the former are duplicated in the later, hence running 15 tests twice. It is not an issue, it's just a bit... strange (to those that know the tests are strictly the same).

As you have plenty of time these days, maybe I would like to update the patches so that tests from regression_7000.c are disabled/removed in gp_7000.c ? If you don't want to:
`Tested-by: Etienne Carriere <etienne.carriere@linaro.org> (qemu_virt)`

By the way, ok you squash all the commits.

jforissier:
@etienne-lms you're absolutely right. I have filtered out the duplicate tests from `gp_7000.c`.

etienne-lms:
Nice patch. I was afraid of something bigger. Your change seems ok. My r-b still applies. I'll test that and tell you if something goes wrong.

Makefile
```diff
@@ -199,60 +199,20 @@ patch-package:
...
-	$(call patch-filter-one,9174,xtest_9000.c)
-	$(call patch-filter-one,9195,xtest_9000.c)
-	$(call patch-filter-one,9196,xtest_9000.c)
-	$(call patch-filter-one,9204,xtest_9000.c)
-	$(call patch-filter-one,9239,xtest_9000.c)
+	@echo "INFO: Filter out some tests"
+	@# 7001-7010, 7013, 7016-7019 are in regression_7000.c already
+	$(foreach n,7001 7002 7003 7004 7005 7006 7007 7008 7009 7010 7013 7016 7017 7018 7019,$(call patch-filter-one,$(n),gp_7000.c))
+	$(call patch-filter-one,7038,gp_7000.c)
+	$(foreach n,7538 7540 7546 7557 7559 7577 7641 7642 7643 7644 7686,$(call patch-filter-one,$(n),gp_7500.c))
```
etienne-lms: 7522 must be filtered out.

jforissier: Yes, my bad. Fixed now.

jforissier:
Commits squashed and rebased into a new PR (I've fixed the help text, too, because it was incorrect after merging the test suite structures).

PR #195, opened 2017-04-03
Test hash tree

jenswi-linaro:
Depends on https://github.com/OP-TEE/optee_os/pull/1456


host/xtest/xtest_20000.c
```diff
@@ -112,14 +112,14 @@ static const struct xtest_enc_fs_case xtest_enc_fs_cases[] = {
...
-	XTEST_ENC_FS(1, 2*4096, META1, BLOCK2, VERSION0, VERSION1),
-	XTEST_ENC_FS(1, 3*4096, META1, BLOCK3, VERSION0, VERSION1),
-	XTEST_ENC_FS(1, 4*4096, META1, BLOCK3, VERSION0, VERSION1),
-	XTEST_ENC_FS(1, 4*4096, META1, BLOCK4, VERSION0, VERSION1),
+	XTEST_ENC_FS(1, 4098, META1, BLOCK1, VERSION1, VERSION1),
+	XTEST_ENC_FS(1, 1*4096, META1, BLOCK1, VERSION1, VERSION1),
+	XTEST_ENC_FS(1, 2*4096, META1, BLOCK2, VERSION1, VERSION1),
+	XTEST_ENC_FS(1, 3*4096, META1, BLOCK3, VERSION1, VERSION1),
+	XTEST_ENC_FS(1, 4*4096, META1, BLOCK3, VERSION1, VERSION1),
+	XTEST_ENC_FS(1, 4*4096, META1, BLOCK4, VERSION1, VERSION1),
```
etienne-lms: I don't understand these changes.
Would it be meaningful to keep the old ones **and** add these new cases ?

jenswi-linaro: Each time a node or a block is written after a sync the version for that particular element will toggle. If there's some changes in policy on how often or which version we're starting with, these has to be updated in order to be able to address the currently used element.

When wrong version is corrupted in these test no error will be detected as an already invalid element was corrupted.

These have a very deep dependency to the format of the secure storage files.

jenswi-linaro: To answer your question, no it wouldn't make sense to keep the old ones.
Perhaps we could update the code to parse the secure storage file to be able to determine automatically which elements in the file are active.

jenswi-linaro:
Rebased

PR #202, opened 2017-04-13
Sync with secure storage updates

jenswi-linaro:
Needs https://github.com/OP-TEE/optee_os/pull/1481 for tests to pass


jforissier:
In commit "core: FS: key manager takes supplied UUID", I think you should say in the commit description that the removed test is not applicable since there is no per-TA directory anymore. In any case:

jenswi-linaro:
@jforissier, you mean commit "regression: 6015: remove special case for REE FS"?

jforissier:
Yes, indeed (wrong copy/paste!)

jenswi-linaro:
Commit message updated, tags applied and rebased.

PR #206, opened 2017-05-17
regression 6018: reduce the number of blocks written

jenswi-linaro:
Reduces the number of blocks written when testing RPMB filesystem to
avoid occasional out of memory errors.


jforissier:

jenswi-linaro:
Tag applied

PR #209, opened 2017-05-22
xtest: case TEEC_CONFIG_SHAREDMEM_MAX_SIZE == ULONG_MAX

etienne-lms:
When `TEEC_CONFIG_SHAREDMEM_MAX_SIZE` is `ULONG_MAX`, TEE has no reachable specific limit. Actually implementation may have impose a limit but it is not known.

2nd change propose the "Check that at least the 512kB lower bound of the TEE Client API (v1.0, v2.0) is supported. If agreed, will be squashed with former.

### DIFF
--- a/host/xtest/regression_5000.c
+++ b/host/xtest/regression_5000.c
@@ -68,6 +68,12 @@ struct xtest_session {
 		} \
 	} while (0)
 
+/* Maximal value to be tested on TEEC_CONFIG_SHAREDMEM_MAX_SIZE tests: 512kB */
+#define TEE_CLIENT_API_LOWER_SHM_BUFFER_MAX_SIZE	0x80000
+#ifndef MIN
+#define MIN(a, b)		((a) < (b) ? (a) : (b))
+#endif
+
 /* Initiates the memory and allocates size uint32_t. */
 
 /* Registers the TEEC_SharedMemory to the TEE. */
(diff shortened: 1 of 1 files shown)


etienne-lms:
This change fixes current failures of master branches since merge of https://github.com/OP-TEE/optee_client/pull/89.

jforissier:
IMO it's not the correct way to address the issue. Those "maxsize" tests (and above/below max size) are meaningless because whatever the value of the requested size *and* the value of TEE_CONFIG_SHAREDMEM_MAXSIZE, there is no guarantee the allocation will succeed (or even fail for big values).
I think the best we can do is replace those tests by a single allocation/registration of `min(512KB, TEE_CONFIG_SHAEDMEM_MAX_SIZE)`. This you can expect to succeed if you know you're the only application executing.

etienne-lms:
ok, and simply remove the allocation failure case.

host/xtest/regression_5000.c
```diff
@@ -541,7 +555,7 @@ static void Register_sharedMemory_aboveMaxSize(struct xtest_session *cs)
 	Do_ADBG_BeginSubCase(cs->c, "Register_sharedMemory_aboveMaxSize");
 	{
 		TEEC_Result res;
-		uint32_t size_aboveMax = 0xffffffff;
+		uint32_t size_aboveMax = UINT32_MAX;
```
jenswi-linaro: This should be a `size_t` and the value `SIZE_MAX`.
`RegisterSharedMemory()` will need to be updated to take the size as `size_t` instead of `uint32_t` (that's what we get for using fixed sized integers where they don't belong).

etienne-lms: true. my fault. I looked at the wrong API.

jforissier:
@etienne-lms that will work but to be honest I find it still too complicated. Here is what I had in mind: https://github.com/jforissier/optee_test/commit/alloc-512k

etienne-lms:
I like your change. less conservative.
Can you push as P-R ? I'll close mine.
Note the `uint32_t` to used that should be replaced with `size_t`. Also needs the same changes in `RegisterSharedMemory()` and `AllocateSharedMemory()` prototypes.

jforissier:
@etienne-lms https://github.com/OP-TEE/optee_test/pull/210. This one can be closed now. Thanks for looking into the issue

etienne-lms:
yop! closed.

PR #210, opened 2017-05-23
regression_5006: use 32KB buffers

jforissier:
Fix for the Travis failures we're seing currently (for example https://travis-ci.org/OP-TEE/optee_os/builds/235179728#L4844) caused by the optee_client update: 9faeb1b6b023 ("tee_client_api: Correct max shared memory size define").

Plus some `uint32_t` => `size_t` cleanup.


jforissier:
Forgot to add T-b to first commit:
`Tested-by: Jerome Forissier <jerome.forissier@linaro.org> (HiKey, QEMU)`

etienne-lms:

PR #211, opened 2017-05-30
xtest: fix 32bit/64bit build directive when GP suite is enabled

etienne-lms:
OP-TEE core configuration directive for AArch32 is `CFG_ARM32_core` not
`CFG_ARM32`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,9 +30,6 @@ It makes use of the following environment variables:
 * `TA_DEV_KIT_DIR`: the path to the Trusted Application Dev Kit.
   It can be found in optee_os repository, once optee_os has been compiled.
 * `O`: the output repository
-* When the application is compiled in 32bits mode,
-  `CFG_ARM32=y` must be set.
-
 
 #### Extended test (Global Platform tests)
 Developers can purchase the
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -54,6 +51,12 @@ This will:
 
 Then the tests must be compiled with `CFG_GP_PACKAGE_PATH=<path>`.
 
+It makes use of the following environment variable:
+* `COMPILE_NS_USER`: `32` or `64` if application shall be compiled in AArch32 on in
+  AArch64 mode. If not specificed, build relies on `CFG_ARM32_core=y` from
+  OP-TEE core build to assume application is AArch32. Otherwise, AArch64 is
+  assumed.
```
jforissier: Minor: using Aarch32 and Aarch64 in this context is incorrect, as they are execution states of ARMv8. It would be better to write "32 bits" and "64 bits" which is a bit imprecise but good enough for the purpose.

etienne-lms: ok, thanks.

etienne-lms:
REAMDE.md updated. Tag applied.

PR #214, opened 2017-06-02
Make sure AES CTR does not buffer data

jforissier:
Similar to commit c13fafac91ae ("Make sure AES GCM does not buffer
data"), update regression_4003_NO_XTS to check that TEE_CipherUpdate()
does not not buffer data when algorithm is TEE_ALG_AES_CTR.


jenswi-linaro:

PR #216, opened 2017-06-07
Add regression_1016: TA to TA buffer manipulations

jforissier:
This test case is derived from gp_8040 which was found to be failing on
HiKey with a preliminary version of the 64-bit pager enabled
(CFG_ARM64_core=y CFG_WITH_PAGER=y CFG_PAGED_USER_TA=y). This new
regression_1016 test fails in the same circumstances, but also when
paging is enabled in a 32-bit build of the TEE core.

### DIFF
--- a/ta/os_test/os_test.c
+++ b/ta/os_test/os_test.c
@@ -917,3 +917,125 @@ TEE_Result ta_entry_bad_mem_access(uint32_t param_types, TEE_Param params[4])
 
 	return TEE_SUCCESS;
 }
+
+static void incr_values(size_t bufsize, uint8_t *a, uint8_t *b, uint8_t *c)
+{
+	size_t i;
+
+	for (i = 0; i < bufsize; i++) {
+		a[i]++; b[i]++; c[i]++;
+	}
+}
+
+TEE_Result ta_entry_ta2ta_memref(uint32_t param_types, TEE_Param params[4])
+{
+	static const TEE_UUID test_uuid = TA_OS_TEST_UUID;
+	TEE_TASessionHandle sess = TEE_HANDLE_NULL;
+	TEE_Param l_params[4] = { { {0} } };
+	size_t bufsize = 2 * 1024;
+	uint8_t in[bufsize];
+	uint8_t inout[bufsize];
+	uint8_t out[bufsize];
+	TEE_Result res;
+	uint32_t ret_orig;
+	uint32_t l_pts;
+	size_t i;
+	(void)params;
+
+	if (param_types != TEE_PARAM_TYPES(0, 0, 0, 0))
+		return TEE_ERROR_GENERIC;
+
+	res = TEE_OpenTASession(&test_uuid, 0, 0, NULL, &sess, &ret_orig);
+	if (res != TEE_SUCCESS) {
+		EMSG("TEE_OpenTASession failed");
+		goto cleanup_return;
+	}
+
+	l_pts = TEE_PARAM_TYPES(TEE_PARAM_TYPE_MEMREF_INPUT,
+				TEE_PARAM_TYPE_MEMREF_INOUT,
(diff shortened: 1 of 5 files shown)


ta/os_test/os_test.c
```diff
@@ -917,3 +917,137 @@ TEE_Result ta_entry_bad_mem_access(uint32_t param_types, TEE_Param params[4])
...
+	for (i = 0; i < bufsize; i++) {
+		if (in[i] != 6 || inout[i] != 13 || out[i] != 19) {
+			EMSG("Unexpected value in buffer(s)");
+			DHEXDUMP(in, bufsize);
+			DHEXDUMP(inout, bufsize);
+			DHEXDUMP(out, bufsize);
+			return TEE_ERROR_GENERIC;
+		}
+	}
+
```
etienne-lms: Could this TA also try with buffers from the TA itselft: from the stack, from the heap.

Another nice scenario is the TA to invoke a pTA (some test pTA), also with memrefs from the shm or from the TA memory (stack, heap).

jforissier: > Could this TA also try with buffers from the TA itselft: from the stack, from the heap.

What do you mean? `in`, `out` and `inout` are declared on the stack of the calling TA already.

> Another nice scenario is the TA to invoke a pTA (some test pTA), also with memrefs from the shm or from the TA memory (stack, heap).

Agreed, but let's leave that for another PR. This one focuses on a special case as I said in the description. Since gp_8040 is a GP test, it is not available to Linaro people in general, so it's useful to reproduce the bug in a non-GP test (and the GP test wouldn't trigger the bug with 32-bit paging anyway, which makes this version even more useful).

etienne-lms: ): I should widely open my eyes. It's the nonsecure shm case that is not tested.
But ok, let's get this later.

ta/os_test/os_test.c
```diff
@@ -917,3 +917,137 @@ TEE_Result ta_entry_bad_mem_access(uint32_t param_types, TEE_Param params[4])
...
+	bufsize = params[0].memref.size;
+	if (params[1].memref.size != bufsize ||
+	    params[2].memref.size != bufsize)
+		return TEE_ERROR_GENERIC;
+
+	in = params[0].memref.buffer;
+	inout = params[1].memref.buffer;
+	out = params[2].memref.buffer;
+
+	for (i = 0; i < bufsize / 2; i++) {
```
etienne-lms: why this dual operation? need for speed?

jforissier: Not at all, the idea is to maximize the chance of triggering page faults, since the bug is related to paging. But I'm not sure it is necessary to reproduce the bug. I should try with a simple loop.

jforissier:
I have simplified the loops (the more complex version was not needed to reproduce the pager issue).

About increasing TA_STACK_SIZE: it does not seem to be needed, as the TA does not cause any exception due to accessing past the current 8K stack (tried both 32-bit and 64-bit TAs).

jenswi-linaro:

jforissier:
@etienne-lms are you OK with this?

etienne-lms:
yes sorry, missed the tag: `Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>`.
Maybe could increase a bit the stack: it was 8KB and this change adds 6kB to stack consumption. Otherwise, we'll increase it when experiencing an issue.

jforissier:
> [...]this change adds 6kB to stack consumption. Otherwise, we'll increase it when experiencing an issue.

Yes that's my intent. In fact, I don't know if this change "adds 6kB to stack consumption". Maybe some test needs 8kB, this one needs 6kB, and there is no reason to add the numbers since the first test doesn't call the second one.

Thanks for the R-b.

PR #221, opened 2017-08-05
crypt: Add hack to support 64-bit address space.

noopwafel:
The code is passing pointers around by casting them to 32-bit integers; convert them to/from offsets from ta_head instead. With this patch, all the tests pass even when TAs are loaded at arbitrary 64-bit virtual addresses (i.e. ASLR).

This is pretty ugly, but I thought I'd at least publish this patch just in case it helps anyone else.

### DIFF
--- a/ta/crypt/cryp_taf.c
+++ b/ta/crypt/cryp_taf.c
@@ -34,7 +34,10 @@ do {                                             \
 		return TEE_ERROR_BAD_PARAMETERS; \
 } while (0)
 
-#define VAL2HANDLE(v) (void *)(uintptr_t)(v)
+/* pointers in 32-bit fields passed as an offset to ta_head */
+extern const void *ta_head;
+#define VAL2HANDLE(v) (void *)(uintptr_t)(v == TEE_HANDLE_NULL ? v : v + (uintptr_t)&ta_head)
+#define HANDLE2VAL(h) (uint32_t)(h == TEE_HANDLE_NULL ? (uintptr_t)h : (uintptr_t)((uintptr_t)h - (uintptr_t)&ta_head))
 
 TEE_Result ta_entry_allocate_operation(uint32_t param_type, TEE_Param params[4])
 {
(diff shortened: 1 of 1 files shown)


ta/crypt/cryp_taf.c
```diff
@@ -34,7 +34,10 @@ do {                                             \
 		return TEE_ERROR_BAD_PARAMETERS; \
 } while (0)
 
-#define VAL2HANDLE(v) (void *)(uintptr_t)(v)
+/* pointers in 32-bit fields passed as an offset to ta_head */
+extern const void *ta_head;
+#define VAL2HANDLE(v) (void *)(uintptr_t)(v == TEE_HANDLE_NULL ? v : v + (uintptr_t)&ta_head)
+#define HANDLE2VAL(h) (h == TEE_HANDLE_NULL ? (uintptr_t)h : (uintptr_t)((uintptr_t)h - (uintptr_t)&ta_head))
```
etienne-lms: since returning a `uint32_t`, i think it should be:
`#define HANDLE2VAL(h) (uint32_t)(h == TEE_HANDLE_NULL ? h : (uintptr_t)h - (uintptr_t)&ta_head)`

jforissier:
The patch seems acceptable to me, the problem is: we (Linaro) have no way to test it. That's not good for maintenance. @noopwafel could you share some code to that end? You say it fixes issues when ASLR is enabled for 64-bit TAs. Could you share your ASLR core maybe? Or if you can't, maybe you can propose some simple modification to the address mapping of one of the 64-bit platforms supported upstream (QEMUv8 would be good) so that TA virtual addresses are > 2**32?

noopwafel:
Unfortunately I don't have any nice ASLR code to share, but https://github.com/OP-TEE/optee_os/pull/1746 is hopefully enough to test this.

(Updated the request with the suggested changes, thanks.)

jenswi-linaro:
The code is actually tested regardless if we're using addresses larger than 2^32 or not.

jforissier:
@jenswi-linaro sure, the code is tested against regressions, and that might be considered good enough for merging. What I meant is: this patch is fixing an issue we don't have on our supported platforms, so we can't be certain the fix is correct and comprehensive (does the whole xtest suite run with large user space VAs?).

etienne-lms:
The offset proposal will make it ok for up to 4GB va space. LPAE current limits user vmem to 1GB, while 32bit-mmu limits to 32MB.

jforissier:
Merged, thanks!

PR #223, opened 2017-08-28
xtest: add -DTA_DIR to meet different requirement

airbak:
TA_DIR default value is "/lib/optee_armtz", it should be
able to be changed for different requirement.

Signed-off-by: Jianhui Li <airbak.li@hisilicon.com>


jforissier:

PR #222, opened 2017-08-24
xtest/sdp: sync with ION udpates from linux kernel 4.12

etienne-lms:
This change depends on integration of the optee-next branch of the linux kernel.
Do not merge until optee relies on k4.12 with updated ION drivers.

PR #228, opened 2017-10-17
xtest:sdp: fix a xtest crash if TEE is built without test pTA

etienne-lms:
Xtest crashes in 1014 when TEE core was built without SDP test pTA support (I.e `CFG_TEE_CORE_EMBED_INTERNAL_TESTS=n`).

Note: xtest 1014 is enabled only is `CFG_SECURE_DATA_PATH=y`.


etienne-lms:
Change merged in https://github.com/OP-TEE/optee_test/pull/222 since related to it.

PR #231, opened 2017-10-20
xtest:sdp: support old ION API if new API fails

etienne-lms:
Implement the old ION API using `ion0_`/`ION0_` as prefix.

Routine allocate_ion_buffer() attempts to allocate using the new
API. If the 1st request (query heap types) fails, try the old API.

### DIFF
--- a/host/xtest/sdp_basic.c
+++ b/host/xtest/sdp_basic.c
@@ -41,6 +41,8 @@
 #include "sdp_basic.h"
 #include "xtest_test.h"
 
+#include "include/uapi/linux/ion_old.h"
+
 /*
  * SDP basic test setup overview.
  *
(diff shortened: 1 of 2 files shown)


host/xtest/sdp_basic.c
```diff
@@ -76,6 +78,37 @@ struct tee_ctx {
 	TEEC_Session sess;
 };
 
+/*
+ * Old ION API to allocate and export a buffer
+ */
+static int allocate_ion_old_buffer(size_t size, int heap_type_id, int ion)
```
jforissier: Nitpicking: `allocate_old_ion_buffer()`?

etienne-lms: that's right. i'll fix.

etienne-lms:
squashed, tags applied.

PR #233, opened 2017-10-26
[RFC] sdp: add a searchable test OK message

jbech-linaro:
I want to have something like this merged, but I'm putting this as an RFC, since eventually we should have some kind of standard prefix that makes it easy go "grep" from automated test environment. In this commit I have **no** such thing. But we could for example change/add this to all test cases with something like:
* `[op-tee: sdp: ok]` when **everything** in that suite has passed.
* `[op-tee: sdp: fail]` when **something** in that suite has failed.

Similar we could add the same to standard xtest.
* `[op-tee: xtest standard: ok]` when **everything** in that suite has passed.
* `[op-tee: xtest standard: fail]` when **something** in that suite has failed.
* `[op-tee: xtest l15: ok]` when **everything** in that suite has passed (xtest -l15)
* `[op-tee: xtest l15: fail]` when **something** in that suite has failed (xtest -l15).
etc ... you get the idea.

Something like that should be unique enough to not get mixed with other stuff.

### DIFF
--- a/host/xtest/sdp_basic.c
+++ b/host/xtest/sdp_basic.c
@@ -644,5 +644,11 @@ int sdp_basic_runner_cmd_parser(int argc, char *argv[])
 			     rnd_offset, verbosity);
 	CHECK_RESULT(err, 1, return 1);
 
+	/* 
+	 * Make sure this string is printed after all subtests has been running.
+	 * This string is used by automatic testing. I.e., we are searching for
+	 * this string to figure out whether the SDP tests passed or not.
+	 */
+	fprintf(stdout, "All SDP test cases passed!\n");
 	return 0;
 }


jenswi-linaro:
Why doesn't the SDP tests use the same framework as the regression tests?

etienne-lms:
xtest 1014 is based on sdp-basic.c.
Maybe there is some generic log to add in to get the trace you expect.

Actually, greping 1014 should give the SDP status as a regular xtest case. I expect.
You want to group the xtest cases in families for status reporting?

jbech-linaro:
I'm not after checking sub tests. I want to have a clear answer last whether all went OK or not. Checking individual tests is something that is easy to do just search in the log if the test suite reported an error. As you can see [here](https://github.com/jbech-linaro/hikey_auto_boot/blob/webserver/hikey_job_cfg.yaml#L76-L77) plain xtest is rather easy to search for. But if you check [further down](https://github.com/jbech-linaro/hikey_auto_boot/blob/webserver/hikey_job_cfg.yaml#L80-L113) in the same file, you can see that I need to check for `Test passed` for each test case. It would be much better to have a single message by the end saying all went fine. I.e, the patch here would solve that problem. But, I though that having a similar message wherever we are running tests would be it trivial and also uniform to search for the string.

etienne-lms:
I think I understand your issue.
Your script runs the standard xtest (including non regression 1014) then it runs the `xtest --sdp-basic`. Actually `xtest -t regression 1014` does the same as `xtest --sdp-basic -n 10 -s 17000`.
Maybe hikey does not need to run the "xtest --sdp-basic" with all the specific (useful?) test steps ?

etienne-lms:
Maybe we could add some level of test into 1014, so that -l15 test more various sizes and more loops.

Issue #232, opened 2017-10-26
Not able to Open Persistent Object


sahilnxp:
Hi All,

I am new to TA programming and having some issue with the Persistent Objects creation/opening.
Below is the code snippet and output when i run the code.
In this program first created a Persistent Object then Opened it and then Closed it.
After this I tried creating already created Persistent Object, It gives the TEE_ERROR_ACCESS_CONFLICT, which i fine.

After this when I tried opening that Persistent Object it is again giving the TEE_ERROR_ACCESS_CONFLICT, which I am not able to understand.

I want to know that why Second TEE_OpenPersistentObject is failing ?
Is there any limitation that if you try creating an already created PersistentObject, access to that object is locked somehow ?

Any help on this is really appreciated.

![image](https://user-images.githubusercontent.com/27992285/32050488-9fa2d382-ba6e-11e7-81e4-a4f42e901c56.png)

![image](https://user-images.githubusercontent.com/27992285/32050546-c7a435d8-ba6e-11e7-8f0a-271f43449d01.png)

Thanks
Sahil

jbech-linaro:
Hi @sahilnxp ,

Please have a look at the not yet merged pull request here so see whether that helps understand how to work with persistent storage objects.
https://github.com/linaro-swg/optee_examples/pull/5

// Joakim

sahilnxp:
Hi @jbech-linaro 

Thanks for reply.
I studied the code given in the pull request, but it is not having the case I am trying.

I am trying to create an already created Persistent object without TEE_DATA_FLAG_OVERWRITE which is giving me the TEEC_ERROR_ACCESS_CONFLICT, which is correct according to the specification.
After getting this error since I know that the object is already created, so I tried opening that Persistent Object but it is giving the TEEC_ERROR_ACCESS_CONFLICT.

So I want to understand that if we try creating an already created object without TEE_DATA_FLAG_OVERWRITE, then we will be not be able to open that object any further ?

Thanks,
Sahil

etienne-lms:
This puzzle me and i tried to same from the ote_secure_storage. I think you spotted an issue in the core. The create_object without overwrite fails, but it still create an object instance and does not release it. See https://github.com/OP-TEE/optee_os/pull/1903 (needs storage experts review).

sahilnxp:
Thanks @etienne-lms  for the fix.

[...]

PR #234, opened 2017-10-27
xtest storage: add a test case of creation of already existing object

etienne-lms:
Regression xtest 6012 already tests creation and recreation of a persistent object. This change adds a test in 6012 to re-create the existing object without setting the overwrite attribute. Creation should fail but the object should not be altered and former open should succeed.

### DIFF
--- a/host/xtest/regression_6000.c
+++ b/host/xtest/regression_6000.c
@@ -1406,17 +1406,22 @@ static void xtest_tee_test_6012_single(ADBG_Case_t *c, uint32_t storage_id)
 	uint32_t orig;
 	uint32_t obj;
 
+	/*
+	 * create the object a first time (forced through with overwrite attribute)
+	 */
 	if (!ADBG_EXPECT_TEEC_SUCCESS(c,
 		xtest_teec_open_session(&sess, &storage_ta_uuid, NULL, &orig)))
 		return;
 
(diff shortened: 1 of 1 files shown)


etienne-lms:
This change is related to issue https://github.com/OP-TEE/optee_test/issues/232 and fix https://github.com/OP-TEE/optee_os/pull/1903.

host/xtest/regression_6000.c
```diff
@@ -1406,43 +1406,68 @@ static void xtest_tee_test_6012_single(ADBG_Case_t *c, uint32_t storage_id)
...
 	if (!ADBG_EXPECT_TEEC_SUCCESS(c,
 		fs_create_overwrite(&sess, file_04, sizeof(file_04), storage_id)))
-		goto exit;
+		goto bail1;
 
 	TEEC_CloseSession(&sess);
 
-	/* re-create the same */
+	/*
+	 * re-create the object two times with overwrite attribute
```
jforissier: You mean a second time?

etienne-lms: The sequence below re-creates twice the already existing object. I left it as it was.
Maybe I could say `re-create twice again the object with overwrite attribute`?

jforissier: No it's OK

etienne-lms:
tag applied.

PR #237, opened 2017-11-06
regression: 4005: add nist aes-gcm vectors

jenswi-linaro:
Adds nist AES-GCM test vectors to regression case 4005 with
CFG_GCM_NIST_VECTORS=y. Only the first test in each group is
used if CFG_GCM_NIST_VECTORS_LIMITED=y.

With CFG_GCM_NIST_VECTORS=y the size of the xtest binary grows with more
than 10 MiB, unless CFG_GCM_NIST_VECTORS_LIMITED=y is given, then only
by ~1 MiB.


host/xtest/regression_4000.c
```diff
@@ -2654,8 +2654,8 @@ static const struct xtest_ae_case ae_cases[] = {
 	XTEST_AE_CASE_AES_GCM(vect1, 0, 0, NULL_ARRAY, NULL_ARRAY, NULL_ARRAY),
 	XTEST_AE_CASE_AES_GCM(vect2, 0, 9, NULL_ARRAY, ARRAY, ARRAY),
 	XTEST_AE_CASE_AES_GCM(vect3, 0, 9, NULL_ARRAY, ARRAY, ARRAY),
-	XTEST_AE_CASE_AES_GCM(vect3, 0, 0x20, NULL_ARRAY, ARRAY, ARRAY),
-	XTEST_AE_CASE_AES_GCM(vect4, 5, 9, ARRAY, ARRAY, ARRAY),
+	XTEST_AE_CASE_AES_GCM(vect3, 0, 0x1F, NULL_ARRAY, ARRAY, ARRAY),
+	XTEST_AE_CASE_AES_GCM(vect4, 5, 0x20, ARRAY, ARRAY, ARRAY),
```
jforissier: Why?

scripts/rsp_to_gcm_test.py
```diff
@@ -0,0 +1,112 @@
+#!/usr/bin/env python
```
jforissier: - Does it run with Python 2 & 3?
- Better make this script pass `pycodestyle`/`pep8`

host/xtest/regression_4000.c
jenswi-linaro: It increases test coverage, it should be in a separate commit though. I'll fix.

scripts/rsp_to_gcm_test.py
jenswi-linaro: Yes it seems to work with both Python 2 and 3.
I'll fix the style stuff.

jenswi-linaro:
Updated and rebased

jforissier:

That's a **lot** of test cases indeed :open_mouth: 

```
root@HiKey960:/ time xtest 4005
...
o regression_4005.25290 AE case 25289 algo 0x40000810 line 65897
  regression_4005.25290 OK
  regression_4005 OK
...
real	3m 22.96s
user	0m 2.22s
sys	1m 2.48s
...

jenswi-linaro:
Tags applied

PR #242, opened 2017-11-14
regression: add case 1003 for mutex

jenswi-linaro:
Adds regression case 1003 for read/write mutex testing.

Depends on https://github.com/OP-TEE/optee_os/pull/1942

### DIFF
--- a/host/xtest/regression_1000.c
+++ b/host/xtest/regression_1000.c
@@ -34,6 +34,7 @@
 
 static void xtest_tee_test_1001(ADBG_Case_t *Case_p);
 static void xtest_tee_test_1002(ADBG_Case_t *Case_p);
+static void xtest_tee_test_1003(ADBG_Case_t *Case_p);
 static void xtest_tee_test_1004(ADBG_Case_t *Case_p);
 static void xtest_tee_test_1005(ADBG_Case_t *Case_p);
 static void xtest_tee_test_1006(ADBG_Case_t *Case_p);
@@ -52,6 +53,8 @@ static void xtest_tee_test_1016(ADBG_Case_t *Case_p);
 
 ADBG_CASE_DEFINE(regression, 1001, xtest_tee_test_1001, "Core self tests");
 ADBG_CASE_DEFINE(regression, 1002, xtest_tee_test_1002, "PTA parameters");
+ADBG_CASE_DEFINE(regression, 1003, xtest_tee_test_1003,
+		 "Core internal read/write mutex");
 ADBG_CASE_DEFINE(regression, 1004, xtest_tee_test_1004, "Test User Crypt TA");
 ADBG_CASE_DEFINE(regression, 1005, xtest_tee_test_1005, "Many sessions");
 ADBG_CASE_DEFINE(regression, 1006, xtest_tee_test_1006,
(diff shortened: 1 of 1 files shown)


host/xtest/regression_1000.c
```diff
@@ -324,7 +326,129 @@ static void xtest_tee_test_1002(ADBG_Case_t *c)
 	TEEC_CloseSession(&session);
 }
 
+struct test_1003_arg {
+	uint32_t test_type;
+	size_t repeat;
+	size_t max_before_lockers;
+	size_t max_during_lockers;
```
jforissier: `max_` values are never reported. Print them, or remove them?

host/xtest/regression_1000.c
```diff
@@ -324,7 +326,129 @@ static void xtest_tee_test_1002(ADBG_Case_t *c)
...
+
+			num_concurrent_read_lockers += arg[n].during_lockers;
+			num_concurrent_read_waiters += arg[n].before_lockers;
+		}
+	}
+
+	mean_read_concurrency = (double)num_concurrent_read_lockers /
+				(double)(repeat * num_threads / 2);
+	mean_read_waiters = (double)num_concurrent_read_waiters /
+				(double)(repeat * num_threads / 2);
```
jforissier: Not sure I understand what you are counting here. Why `/ 2`?

host/xtest/regression_1000.c
```diff
@@ -52,6 +53,7 @@ static void xtest_tee_test_1016(ADBG_Case_t *Case_p);
 
 ADBG_CASE_DEFINE(regression, 1001, xtest_tee_test_1001, "Core self tests");
 ADBG_CASE_DEFINE(regression, 1002, xtest_tee_test_1002, "PTA parameters");
+ADBG_CASE_DEFINE(regression, 1003, xtest_tee_test_1003, "Mutex");
```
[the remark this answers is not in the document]
jenswi-linaro: It has nothing to do with PTA, except that it happens to be tested from a PTA. How about "Core internal read/write mutex" instead?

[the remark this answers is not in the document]
jforissier: Sounds good

host/xtest/regression_1000.c
jenswi-linaro: The number of read threads is "num_threads / 2" since every second thread is a read thread.
Concurrency is how many read thread holds the read lock at the same time
Waiters is how many read threads are blocked on the `mutex_read_lock()` call.
I'll explain this in the code.

jforissier: > every second thread is a read thread.

I see `if (n % 3)`, not `if (n % 2)`, so that is 2 read threads for 1 write thread, isn't it?

jenswi-linaro: I'm sorry, I'm slightly out of sync with what's pushed. It should be `/3` in the expression. I'll update.

jenswi-linaro:
Update

host/xtest/regression_1000.c
```diff
@@ -324,7 +327,141 @@ static void xtest_tee_test_1002(ADBG_Case_t *c)
...
+	size_t during_lockers;
+	TEEC_Result res;
+	uint32_t error_orig;
+};
+
+static void *test_1003_thread(void *arg)
+{
+	struct test_1003_arg *a = arg;
+	TEEC_Session session = { 0 };
+	size_t rounds = 64 * 1024;
```
jforissier: FYI, this is much too small to always reach the theoretical maximum number of read concurrency and read waiters (4) on HiKey960. But it doesn't really matter I suppose.

jforissier:

host/xtest/regression_1000.c
jenswi-linaro: Agree

jenswi-linaro:
Squashed and tags applied

PR #244, opened 2017-12-14
Reduce loop count in regression_1012

jforissier:
Reduce the loop count in xtest 1012 from 1000 to 3 in order to minimize
debug traces. There is no point in using a higher value in this
functional test.


jbech-linaro:

PR #247, opened 2017-12-22
[RFC] sks: regression tests for secure key services work base

etienne-lms:
Related to:
* TA from optee_os : https://github.com/OP-TEE/optee_os/pull/2025
* Client lib from optee_client : https://github.com/OP-TEE/optee_client/pull/113
* Build: https://github.com/OP-TEE/build/pull/219

### DIFF
--- a/host/xtest/regression_4000.c
+++ b/host/xtest/regression_4000.c
@@ -1727,6 +1727,352 @@ out:
 	TEEC_CloseSession(&session);
 }
 
+#ifdef CFG_SECURE_KEY_SERVICES
+/*
+ * The test below belongs to the regression 41xx test. As it rely on test
+ * vectors define for the 40xx test, this test sequence in implemented here.
+ * The test below check compliance of crypto algorithms called throug the SKS
+ * PKCS#11 interface.
+ */
+void run_xtest_tee_test_4111(ADBG_Case_t *c, CK_SLOT_ID slot);
+
+/* AES CMAC test resrouces */
+#define CK_MAC_KEY_AES(_key_array) \
+	{								\
+		{ CKA_SIGN, &(CK_BBOOL){CK_TRUE}, sizeof(CK_BBOOL) }, \
+		{ CKA_VERIFY, &(CK_BBOOL){CK_TRUE}, sizeof(CK_BBOOL) }, \
+		{ CKA_KEY_TYPE,	&(CK_KEY_TYPE){CKK_AES},		\
+						sizeof(CK_KEY_TYPE) },	\
+		{ CKA_CLASS, &(CK_OBJECT_CLASS){CKO_SECRET_KEY},	\
+						sizeof(CK_OBJECT_CLASS) }, \
+		{ CKA_VALUE, (void *)_key_array, sizeof(_key_array) }, \
+	}
+
+static CK_ATTRIBUTE cktest_aes_cmac_key1[] =
+	CK_MAC_KEY_AES(mac_cmac_vect1_key);
+
+static CK_ATTRIBUTE cktest_aes_cmac_key2[] =
+	CK_MAC_KEY_AES(mac_cmac_vect5_key);
+
+static CK_ATTRIBUTE cktest_aes_cmac_key3[] =
+	CK_MAC_KEY_AES(mac_cmac_vect9_key);
+
+static CK_MECHANISM cktest_aes_cmac_mechanism = {
+	CKM_AES_CMAC, NULL, 0,
+};
+#if 0
+static CK_MECHANISM cktest_aes_cmac_gen_mechanism = {
+	CKM_AES_CMAC_GENERAL, &(CK_ULONG){12}, sizeof(CK_ULONG),
+};
+#endif
+
+/* HMAC test resrouces */
+
+#define CK_MAC_KEY_HMAC(_type, _key_array) \
+	{								\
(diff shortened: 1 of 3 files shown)


host/xtest/xtest_main.c
```diff
@@ -67,6 +71,9 @@ void usage(char *program)
 	printf("\t--aes-perf [opts]  AES performance testing tool (-h for usage)\n");
 #ifdef CFG_SECURE_DATA_PATH
 	printf("\t--sdp-basic [opts] Basic Secure Data Path test setup ('-h' for usage)\n");
+#endif
+#ifdef CFG_SECURE_KEY_SERVICES
+	printf("\t--sks [opts]       Secure key services test setup ('-h' for usage)\n");
```
jenswi-linaro: Why do we need yet another test suite, couldn't this be part of the regression suite instead?

etienne-lms:
Updated. Test moved to regression 4003 and 4012.

host/xtest/regression_4000.c
```diff
@@ -5136,3 +5389,848 @@ static void xtest_tee_test_4011(ADBG_Case_t *c)
...
+
+	rv = close_lib();
+	ADBG_EXPECT_COMPARE_UNSIGNED(c, rv, ==, CKR_OK);
+
+	Do_ADBG_EndSubCase(c, NULL);
+}
+
+static void xtest_tee_test_4012(ADBG_Case_t *c)
+{
+	xtest_tee_test_4012_token_info(c);
```
jenswi-linaro: The numbers are cheap and we have plenty of them. I think each of these functions deserves its own number and suite entry. This makes it easier when you're only interested in testing a more specific thing, for instance while debugging an error.

We should consider putting all the PKCS#11 test in a 41xx series (and if possible in a separate file, regression_4100.c) since it's a bit different compared to the regular 40xx cases.

host/xtest/regression_4000.c
```diff
@@ -41,6 +45,12 @@ static void xtest_tee_test_4008(ADBG_Case_t *Case_p);
 static void xtest_tee_test_4009(ADBG_Case_t *Case_p);
 static void xtest_tee_test_4010(ADBG_Case_t *Case_p);
 static void xtest_tee_test_4011(ADBG_Case_t *Case_p);
+#ifdef CFG_SECURE_KEY_SERVICES
+static void xtest_tee_test_4003_pkcs11(ADBG_Case_t *Case_p);
```
jenswi-linaro: The 4003_no_xts and 4003_xts exists for historical reason, they could be merged into a single 4003 but none has bothered to do that yet.

Please use a separate number instead of 4003_pkcs11.

host/xtest/regression_4000.c
```diff
@@ -2409,6 +2427,241 @@ static void xtest_tee_test_4003_no_xts(ADBG_Case_t *c)
...
+static CK_OBJECT_CLASS cktest_dataClass = CKO_DATA;
+
+static CK_KEY_TYPE cktest_aes_keyType = CKK_AES;
+static CK_KEY_TYPE cktest_gen_keyType = CKK_GENERIC_SECRET;
+
+static CK_BBOOL cktest_true = CK_TRUE;
+static CK_BBOOL cktest_false = CK_FALSE;
+
+#define	CK_CIPHERING_KEY_AES(_key_array) { \
+		{ CKA_ENCRYPT,	&cktest_true, sizeof(cktest_true) }, \
```
jenswi-linaro: You could consider using compound literals instead, like we doing here https://github.com/OP-TEE/optee_os/blob/master/ta/arch/arm/user_ta_header.c#L91-L105
The line would then become (with no extra dependencies):
```
{ CKA_ENCRYPT, &(CK_BBOOL){CK_TRUE}, sizeof(CK_BBOOL) }
```

host/xtest/regression_4000.c
jforissier: https://github.com/OP-TEE/optee_test/pull/258

host/xtest/regression_4000.c
etienne-lms: Fine, I'll move them the **regression_4100.c**.

host/xtest/regression_4000.c
etienne-lms: i'll rebase.

This test reuses the 4003 data. How to best share the test vectors, if moving the pcsk11 test to a **regression_4100.c** which sound reasonable also to me?

*alt1/* i plan to move the test vectors in **regression_4100_data.h** (where xts vectors are) and have it included in both **regression_4000.c** and **regression_4100.c**  (make each source having a static copy of the vectors). If think it is not very nice.

*alt2/*  Export all test vector in the xtest (remove the `static` attrib). I think it is not nice.

jenswi-linaro: It seems there doesn't exist any nice and easy solution to this so we need to be a bit pragmatic about this. Since you're indeed using `ciph_cases` to drive `xtest_tee_test_4003_pkcs11()` I'm OK with keeping this particular function in this file even if the test number would be in the 41xx series. We can have a comment in regression_4100.c that one of the sequence number are used in regression_4000.c for practical reasons.

etienne-lms: sounds good. ok, thanks for the proposal.

etienne-lms:
Closing this P-R. The feature will be re-submitted once matured.
One can still track on-going developments from https://github.com/etienne-lms/optee_test (branches sks and sks-next).

PR #248, opened 2018-01-19
4003: add aes-ctr corner case

jenswi-linaro:
Adds a AES-CTR corner case to expose a problem in LTC with hardware
accelerated AES encryption.

The fix for the problem is provided in:
https://github.com/OP-TEE/optee_os/pull/2086


jforissier:
With the fix applied, works fine on HiKey960.

jenswi-linaro:
Tags applied

PR #250, opened 2018-01-30
regression: 1006: test passing a NULL reference with 0 size

jforissier:
Test case for OP-TEE issue #2105 [1]: passing a NULL memref.buffer is
acceptable as long as memref.size is zero (GP Internal Core API
specification v1.1.2 section 4.9.4).

Link: [1] https://github.com/OP-TEE/optee_os/issues/2105
Tested-by: Jerome Forissier <jerome.forissier@linaro.org> (QEMU)


jenswi-linaro:

PR #253, opened 2018-02-02
regression 1010: crash TA with SHM memref passed as parameter

etienne-lms:
This adds few test cases in regression test 1010 to check TA crash
when the TA is provided a memref as invocation parameter. These
tests aims at checking the dump of memref parameter state in sane
when core detect TA crash and dumps the TA state for debug purpose.

Refer to https://github.com/OP-TEE/optee_os/pull/2117 which should fix the issue.

The issue can be reproduced on Qemu using this test and the following config:
`make all run CFG_TEE_CORE_LOG_LEVEL=3 CFG_WITH_PAGER=y CFG_TEE_CORE_DEBUG=n CFG_CORE_TZSRAM_EMUL_SIZE=0x40000`

### DIFF
--- a/host/xtest/regression_1000.c
+++ b/host/xtest/regression_1000.c
@@ -23,6 +23,7 @@
 #include "xtest_test.h"
 #include "xtest_helpers.h"
 #include <signed_hdr.h>
+#include <util.h>
 
 #include <pta_invoke_tests.h>
 #include <ta_crypt.h>
@@ -495,7 +496,7 @@ static void xtest_tee_test_1004(ADBG_Case_t *c)
 #define TEEC_ERROR_TARGET_DEAD           0xFFFF3024
 #endif
 
-static void xtest_tee_test_invalid_mem_access(ADBG_Case_t *c, uint32_t n)
+static void xtest_tee_test_invalid_mem_access(ADBG_Case_t *c, unsigned int n)
 {
 	TEEC_Session session = { 0 };
 	TEEC_Operation op = TEEC_OPERATION_INITIALIZER;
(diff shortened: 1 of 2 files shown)


host/xtest/regression_1000.c
```diff
@@ -835,12 +875,26 @@ static void xtest_tee_test_1009(ADBG_Case_t *c)
...
+	int idx;
+	uint32_t memref_sz[] = { 1024, 65536 };
 
 	for (n = 1; n <= 5; n++) {
 		Do_ADBG_BeginSubCase(c, "Invalid memory access %u", n);
 		xtest_tee_test_invalid_mem_access(c, n);
 		Do_ADBG_EndSubCase(c, "Invalid memory access %u", n);
 	}
+
+	for (idx = 0; idx < (sizeof(memref_sz) / sizeof(uint32_t)); idx++) {
```
jenswi-linaro: Please use `ARRAY_SIZE()` instead
Also here you're comparing a signed with an unsigned.
Any particular reason `idx` is an `int` while `n` is an `unsigned`?

etienne-lms: > `int` vs `unsigned`

No specific reason. I can use `unsigned` for both. Maybe `unsigned int` would be even better.

etienne-lms: Note: I use 2 size cases for the memref param: 1 below a page size and above quite far above page size. The latter is mainly to check the behaviour when the memref is a non physically contiguous registered shm buffer. In the later case, I even wonder the accuracy of the generated trace which shows the virtual base address, the physical base address and the size, but is it still relevant when the buffer is not contiguous.

My question here would rather be: is it useful to test several buffer size? even if it's harmless to test several.

jenswi-linaro: +1 to `unsigned int`, please fix `unsigned n` while you're at it.

host/xtest/regression_1000.c
```diff
@@ -834,13 +875,27 @@ static void xtest_tee_test_1009(ADBG_Case_t *c)
 
 static void xtest_tee_test_1010(ADBG_Case_t *c)
 {
-	unsigned n;
+	uint32_t n;
```
jenswi-linaro: We can't seem to get enough of unsigned types. ;-)
I would have used only `size_t` here if nothing else just to avoid the distraction.
For instance, why is `uint32_t` used for `n` instead of `unsigned int` which is used for `idx`?

etienne-lms: I used the native type used in the TA API: `uint32_t` for the operation command. But true that it's a bit silly. I'll move former to `unsigned int`. Sorry bothering.

jenswi-linaro:
With that:

etienne-lms:
Hope that's ok. I applied your tag.

host/xtest/regression_1000.c
```diff
@@ -506,7 +507,7 @@ static void xtest_tee_test_invalid_mem_access(ADBG_Case_t *c, uint32_t n)
 		                        &ret_orig)))
 		return;
 
-	op.params[0].value.a = n;
+	op.params[0].value.a = (uint32_t)n;
```
jenswi-linaro: Minor: Cast is not needed

etienne-lms:
updated. tag applied.

PR #255, opened 2018-02-08
Cmake support for xtest only (not TAs)

jenswi-linaro:
This introduces support for building the host part (what's running in
linux user space) of xtest using CMake. TAs are as before built using TA
dev kit.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,30 @@
+cmake_minimum_required (VERSION 3.2)
+
+# Default cross compile settings
+set (CMAKE_TOOLCHAIN_FILE CMakeToolchain.txt)
+
+################################################################################
+# Compiler flags:
+#   We want to use the same flags in the entire optee_client git
+################################################################################
+add_compile_options (
+	-Wall -Wbad-function-cast -Wcast-align
+	-Werror-implicit-function-declaration -Wextra
+	-Wfloat-equal -Wformat-nonliteral -Wformat-security
+	-Wformat=2 -Winit-self -Wmissing-declarations
+	-Wmissing-format-attribute -Wmissing-include-dirs
+	-Wmissing-noreturn -Wmissing-prototypes -Wnested-externs
+	-Wpointer-arith -Wshadow -Wstrict-prototypes
+	-Wswitch-default -Wunsafe-loop-optimizations
+	-Wwrite-strings -Werror -fPIC
+ 	-Wno-missing-field-initializers
+)
+
+find_program(CCACHE_FOUND ccache)
+if(CCACHE_FOUND)
+	set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE ccache)
+	set_property(GLOBAL PROPERTY RULE_LAUNCH_LINK ccache)
+endif(CCACHE_FOUND)
+
+add_subdirectory (ta)
+add_subdirectory (host/xtest)
--- a/CMakeToolchain.txt
+++ b/CMakeToolchain.txt
@@ -1,0 +1,3 @@
+set (CMAKE_SYSTEM_NAME Linux)
+
+set (CMAKE_SYSTEM_PROCESSOR arm)
--- a/host/xtest/CMakeLists.txt
+++ b/host/xtest/CMakeLists.txt
@@ -1,0 +1,77 @@
+project (xtest C)
+
+include(${OPTEE_TEST_SDK}/host_include/conf.cmake)
+
+################################################################################
+# Packages
+################################################################################
+find_package(Threads REQUIRED)
+if(NOT THREADS_FOUND)
+	message(FATAL_ERROR "Threads not found")
+endif()
+
+include(GNUInstallDirs)
+
+set (SRC
+	adbg/src/adbg_case.c
+	adbg/src/adbg_enum.c
+	adbg/src/adbg_expect.c
+	adbg/src/adbg_log.c
+	adbg/src/adbg_run.c
+	adbg/src/security_utils_hex.c
+	aes_perf.c
+	benchmark_1000.c
+	benchmark_2000.c
+	regression_1000.c
+	regression_4000.c
+	regression_5000.c
+	regression_6000.c
+	regression_7000.c
+	regression_8000.c
+	sha_perf.c
+	xtest_helpers.c
+	xtest_main.c
+	xtest_test.c
+)
+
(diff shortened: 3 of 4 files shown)


jenswi-linaro:
Ping?

etienne-lms:
Do we need to add something in the main build scripts to use CMake for client and host test?
I tried but failed to get the headers when compiling the xtest host part.

CMakeLists.txt
```diff
@@ -0,0 +1,114 @@
...
+	PRIVATE ta/rpc_test/include
+	PRIVATE ta/sdp_basic/include
+	PRIVATE ta/sha_perf/include
+	PRIVATE ta/sims/include
+	PRIVATE ta/socket/include
+	PRIVATE ta/storage_benchmark/include
+)
+
+target_link_libraries (${PROJECT_NAME}
+	PRIVATE pthread
```
igoropaniuk: `${CMAKE_THREAD_LIBS_INIT}` instead of `pthread` (as you're using FindThreads on line 9)

BTW, we can implement the same cmake module-wrapper for `teec` (so we can use just cll `find_package(teec REQUIRED)`, which will automatically provide all paths in the same manner + this `FindTEEC.cmake` can be reused from CMakeLists.txt of other client applications)

CMakeLists.txt
```diff
@@ -0,0 +1,114 @@
+cmake_minimum_required (VERSION 3.2)
+project (xtest C)
+
+include(${OPTEE_TEST_SDK}/host_include/conf.cmake)
```
igoropaniuk: use cache entry for user-settable values?
`set(OPTEE_TEST_SDK "<default value>" CACHE STRING "<description>")`

CMakeLists.txt
```diff
@@ -0,0 +1,114 @@
...
+add_executable (${PROJECT_NAME} ${SRC})
+
+target_compile_options (${PROJECT_NAME} PRIVATE -include conf.h)
+
+target_include_directories(${PROJECT_NAME}
+	PRIVATE host/xtest
+	PRIVATE host/xtest/adbg/include
+	PRIVATE host/xtest/xml/include
+	PRIVATE ${OPTEE_TEST_SDK}/host_include
+	PRIVATE ta/include
```
jbech-linaro: I would have broken out all the TA headers, to a separate file (`CMakeLists.txt` in `optee_test/ta/`), something similar to what I've done [here](https://github.com/jbech-linaro/optee_test/blob/cmake_optee/ta/CMakeLists.txt). Then you can just include them in the `target_link_libraries` further down.

CMakeLists.txt
```diff
@@ -0,0 +1,114 @@
...
+if (CFG_SECURE_DATA_PATH)
+	list (APPEND SRC host/xtest/sdp_basic.c)
+endif()
+
+################################################################################
+# Built binary
+################################################################################
+add_executable (${PROJECT_NAME} ${SRC})
+
+target_compile_options (${PROJECT_NAME} PRIVATE -include conf.h)
```
jbech-linaro: Where do you pick up `conf.h` from? `./out/arm/include/generated/conf.h`? I cannot see that being mentioned in this patch.

jbech-linaro: Ah, you are probably picking it up from `./out/arm/export-ta_arm32/host_include/conf.h` with the `PRIVATE ${OPTEE_TEST_SDK}/host_include`

CMakeLists.txt
jbech-linaro: `find_package(teec REQUIRED)` would be nice, since there is no common (CMake) root for our gits.

jenswi-linaro:
@etienne-lms, you can test build using:
```
repo init -u https://github.com/jenswi-linaro/manifest.git -m default.xml -b br2
```
Then build as usual with:
```
make toolchains
make all -j8
```
To just build the buildroot stuff:
`make buildroot`

CMakeLists.txt
```diff
@@ -0,0 +1,114 @@
+cmake_minimum_required (VERSION 3.2)
```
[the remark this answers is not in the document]
jbech-linaro: I think the reason was when building using Hackbox we noticed that was running 3.2 which is pretty old. We should aim for something newer, but it doesn't seem like we are depending on any of the newer CMake features, so it does not matter that much.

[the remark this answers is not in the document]
jenswi-linaro: We did some test building on hackbox and the Cmake version there was 3.2.
We're obviously not depending on any features from 3.4 so 3.2 is good enough for now.

CMakeLists.txt
```diff
@@ -0,0 +1,114 @@
...
+project (xtest C)
+
+include(${OPTEE_TEST_SDK}/host_include/conf.cmake)
+
```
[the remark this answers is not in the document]
jenswi-linaro: You mean to have some default cross compile setting in case it isn't specified already?

CMakeLists.txt
jenswi-linaro: I get the principle of what you'd like and agree with that, but I'm not certain how to implement this.
How about taking that after everything is merged?

CMakeLists.txt
jenswi-linaro: I'll replace `phtread` as you describe.
Can the rest wait until we've merged everything?

jbech-linaro: > Can the rest wait until we've merged everything?

I think so.

jenswi-linaro:
Update

jbech-linaro:

CMakeLists.txt
igoropaniuk: @jenswi-linaro sure

CMakeLists.txt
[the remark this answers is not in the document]
igoropaniuk: @jenswi-linaro right, the same as we have for `optee_client`

igoropaniuk:
`Reviewed-by: Igor Opaniuk <igor.opaniuk@linaro.org>`

jenswi-linaro:
Squashed and tags applied

jforissier:
@jenswi-linaro could you please rebase on top of master? The CI errors in `xtest 6001` are almost certainly caused by the recent merge of the "no object ID in shared memory" patch, but I'd rather make sure.

jenswi-linaro:
Rebased

jforissier:
All good, thanks.

PR #257, opened 2018-02-12
regression 1010: fix wrong printf format specifier

igoropaniuk:
Fix wrong format specifier, which causes build errors:
error: format '%d' expects argument of type 'int', but argument 4 has type
'size_t {aka long unsigned int}' [-Werror=format=]
     "Invalid memory access %u with %d bytes memref",

`Signed-off-by: Igor Opaniuk <igor.opaniuk@linaro.org>`


jforissier:
Thanks @igoropaniuk!

jenswi-linaro:

igoropaniuk:
tags applied, thanks!

PR #258, opened 2018-02-16
regression: 4003: merge *4003_xts() and *4003_no_xts()

jforissier:
AES XTS tests were kept separate from the other cipher tests for
historical reasons. Merge them together.


jenswi-linaro:
Thanks for doing this

jbech-linaro:
HiKey auto boot fixed. The reference forest had never been updated, i.e., was from November last year, which meant we got a timeout when syncing. I've updated and enabled the reference, so sync time outs should probably not happen in the future.

PR #259, opened 2018-02-16
xtest storage: can't use object_id from shared memory

etienne-lms:
According to the GP spec V1.1, the object_id in create/open/rename
functions is not allowed to reside in the shared memory. This change
insures the test TAs use local buffers to store the target file ID.

Suggested-by: Zeng Tao <prime.zeng@hisilicon.com>
Tested-by: Etienne Carriere <etienne.carriere@linaro.org> (Qemu, GP)

Note: there is no test checking that using object_id from SHM is forbidden. I will look at that.

### DIFF
--- a/ta/storage/storage.c
+++ b/ta/storage/storage.c
@@ -42,6 +42,8 @@ TEE_Result ta_storage_cmd_open(uint32_t param_types, TEE_Param params[4])
 {
 	TEE_Result res;
 	TEE_ObjectHandle o;
+	void *object_id;
+	size_t object_id_size;
 
 	ASSERT_PARAM_TYPE(TEE_PARAM_TYPES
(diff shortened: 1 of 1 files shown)


jforissier:

Thanks for taking care of this!

etienne-lms:
Added a 2nd commit to have a regression test on this.

jforissier:
No please don't add the second test here because I'd like to merge this first, then check that the optee_os PR is OK (CI will use the new `xtest`), then I can merge the optee_os PR. Only after I will  take the new test.
Thanks.

etienne-lms:
ok.

etienne-lms:
Tag applied.

jforissier:
Thanks!

PR #261, opened 2018-02-16
xtest storage: check that we can't use object_id from shared memory

etienne-lms:
According to the GP spec V1.1, the object_id in create/open/rename
functions is not allowed to reside in the shared memory. This change
insures the test TAs use local buffers to store the target file ID.

Suggested-by: Zeng Tao <prime.zeng@hisilicon.com>


etienne-lms:
buggy commit comment:
Should replace `This change insures the test TAs use local buffers to store the target file ID.`
with `This change makes the TAs attempting the use the APIs with faulty object ID buffer location. It is expected that the TA dies.`

host/xtest/regression_6000.c
```diff
@@ -2008,3 +2228,4 @@ ADBG_CASE_DEFINE(regression, 6017, xtest_tee_test_6017,
 		 "Test Persistent objects info");
 ADBG_CASE_DEFINE(regression, 6018, xtest_tee_test_6018, "Large object");
 ADBG_CASE_DEFINE(regression, 6019, xtest_tee_test_6019, "Storage independence");
+ADBG_CASE_DEFINE(regression, 6020, xtest_tee_test_6020, "Forbiden accesses");
```
jforissier: Nit: `s/Forbiden accesses/Object IDs in SHM (negative)/` (or at least fix the typo :))

etienne-lms:
Fixed according the comment + a build issue. Tag applied.

etienne-lms:
rebased

jforissier:
My Ack was not added.

etienne-lms:
looks like i did not even address your comment! It's done now.
Will squash and apply tag.

jforissier:
Thanks, you can finalize the patch now. The CI (HAB) failure looks unrelated, it did not even start `xtest`.

PR #262, opened 2018-02-19
regression 6000: fix buffer part compared in 6010 test.

mel-f:
Second half of output buffer was compared in resgression test 6010,
whereas first part was filled. Now compare correct part.


jforissier:
@mel-f I need your `Signed-off-by:` to merge this patch (see https://github.com/OP-TEE/optee_os/blob/master/Notice.md#contributions). Thanks!

mel-f:
Should be ok now.

jforissier:
It is, thanks again for contributing.
(note: you may/should have added my `Reviewed-by:` too, but no worries, I can add it myself when merging).

jforissier:
Merged manually.

PR #267, opened 2018-03-15
xtest: restore build with GPD test suite

etienne-lms:
To enable the GPD TEE test suite:
	# make -f Makefile-gp-ts CFG_GP_PACKAGE_PATH=... patch

Fix some cast issue reported by compiler.

**FIXME: differentiate 32b/64b modes for linking against the
expected crypto library. Current only supports 32b mode.**

### DIFF
--- a/Makefile-gp-ts
+++ b/Makefile-gp-ts
@@ -1,0 +1,167 @@
+.PHONY: all
+all:
+	$(q)echo "Use target patch to apply patches to the GPD TEE Test Suite" && false
+
+.PHONY: patch
+patch:
+ifdef CFG_GP_PACKAGE_PATH
+CFG_GP_API?=1.0
+CFG_GP_XSL_PACKAGE_PATH?=$(CURDIR)/package/testsuite/global_platform/api_1.0/GP_XSL_TEE_Initial_Configuration-Test_Suite_v1_0_0-2014-12-03-STM
+
+ifeq "$(wildcard $(CFG_GP_XSL_PACKAGE_PATH) )" ""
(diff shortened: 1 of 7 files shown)


jforissier:
What problem are you trying to address exactly? I can build xtest wth GP with the existing Makefile with no problem (I'm not using CMake).
If you're trying to build with GP tests and with CMake, why do you need to introduce yet another Makefile?

etienne-lms:
cmake craps the Makefile. hence another way to apply the gpd ts (patches) even once the optee_test was built. I understood we don't plan to maintain the GNU Makefile.

I have not found my way using Cmake to apply the patches, hence dumped content from the original Makefile.

jforissier:
> cmake craps the Makefile.

Then git checkout can restore it.

> I understood we don't plan to maintain the GNU Makefile.

It's still there so it's still maintained (by me at least, as I said I'm not using CMake yet).

> I have not found my way using Cmake to apply the patches, hence dumped content from the original Makefile.

Fair enough. But, whatever the implementation details (script or Makefile), I'd rather not ask the user to invoke a Makefile with `-f`. That should be `export CFG_GP_PACKAGE_PATH=... && make patch && make`(as it is currently) or even better `export CFG_GP_PACKAGE_PATH=... && make` (letting the build system figure out the files have to be patched since `CFG_GP_PACKAGE_PATH` is set).

That being said, I certainly think it is a good idea to work on making the CMake build support the GP tests so I will happily review and test any update to this PR ;-)

etienne-lms:
gp test patches current depend on legacy gnu make scripts.

PR #270, opened 2018-03-23
regression: 4006: add test case for 3072-bit RSA

jforissier:
Depends on https://github.com/OP-TEE/optee_os/pull/2208.


etienne-lms:

PR #273, opened 2018-04-06
Remove mempool dependency

jenswi-linaro:
Needed to compile with https://github.com/OP-TEE/optee_os/pull/2237


jenswi-linaro:
This can be merged without  OP-TEE/optee_os#2237

jforissier:
First commit: it would be nice to explain why the `pool` parameter is not needed (I just spent 5 minutes trying to figure out before I came across the comment in `mpa_sub_word()`: `/* handle the case when dest is the operand */`).

Other than that:

jenswi-linaro:
Comment addressed, tag applied, rebased on master

PR #276, opened 2018-04-18
Add outbound SHM and SDP memref case

etienne-lms:
Depends on https://github.com/linaro-swg/linux/pull/61 (at least for the changes in regression 1018).

### DIFF
--- a/host/xtest/regression_1000.c
+++ b/host/xtest/regression_1000.c
@@ -1364,23 +1364,55 @@ out:
 ADBG_CASE_DEFINE(regression, 1017, xtest_tee_test_1017,
 		"Test coalescing memrefs");
 
+static void invoke_1byte_out_of_bounds(ADBG_Case_t *c, TEEC_Session *session,
+				       TEEC_SharedMemory *shm)
+{
+	TEEC_Operation op = TEEC_OPERATION_INITIALIZER;
+	TEEC_Result ret = TEEC_ERROR_GENERIC;
+	uint32_t ret_orig = 0;
+
+	op.paramTypes = TEEC_PARAM_TYPES(TEEC_MEMREF_PARTIAL_INPUT,
+					 TEEC_NONE, TEEC_NONE, TEEC_NONE);
+
+	op.params[0].memref.parent = shm;
+	op.params[0].memref.size = shm->size / 2;
+	op.params[0].memref.offset = shm->size - (shm->size / 2) + 1;
+
+	ret = TEEC_InvokeCommand(session, TA_OS_TEST_CMD_PARAMS,
+				 &op, &ret_orig);
+
+	ADBG_EXPECT(c, TEEC_ORIGIN_COMMS, ret_orig);
+	if (ret != TEEC_ERROR_BAD_PARAMETERS && ret != TEEC_ERROR_GENERIC) {
+		ADBG_EXPECT(c, TEEC_ERROR_BAD_PARAMETERS, ret);
+		ADBG_EXPECT(c, TEEC_ERROR_GENERIC, ret);
+	}
+}
+
 static void xtest_tee_test_1018(ADBG_Case_t *c)
 {
 	TEEC_Session session = { };
 	TEEC_Operation op = TEEC_OPERATION_INITIALIZER;
 	uint32_t ret_orig = 0;
 	TEEC_SharedMemory shm = { };
+	TEEC_Result ret = TEEC_ERROR_GENERIC;
 	size_t page_size = 4096;
+	/* Intentionnally not 4kB aligned and odd */
+	uint8_t buffer[6001] = { };
+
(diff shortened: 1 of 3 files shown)


jforissier:
*outbound*? https://en.wiktionary.org/wiki/outbound

host/xtest/regression_1000.c
```diff
@@ -1381,24 +1386,58 @@ static void xtest_tee_test_1017(ADBG_Case_t *c)
...
+	op->paramTypes = TEEC_PARAM_TYPES(TEEC_MEMREF_PARTIAL_INPUT,
+					 TEEC_MEMREF_PARTIAL_INPUT,
+					 TEEC_MEMREF_PARTIAL_OUTPUT,
+					 TEEC_MEMREF_PARTIAL_OUTPUT);
+
+	/*
+	 * The first three memrefs are valid references while last is outbound
+	 * by 1 byte and should lead to failure to invoke the TA.
+	 */
+	op->params[0].memref.parent = shm;
```
jenswi-linaro: Why do we have these three valid memrefs first? To me it looks like it only muddies the water.

etienne-lms: The TA expects 4 params so we have to provide 4.
Yet, we could use 4 invalid references...
I made it to slightly match the designed of the original case 1018.

jenswi-linaro: Please add another TAF or modify it to accept only one parameter too. Especially when it comes to negative tests it's good to not do more than necessary to avoid that something else hides the error.

jenswi-linaro:
It's spelled out of bounds.

host/xtest/regression_1000.c
```diff
@@ -1381,24 +1386,51 @@ static void xtest_tee_test_1017(ADBG_Case_t *c)
...
+
+	op.params[0].memref.parent = shm;
+	op.params[0].memref.size = shm->size / 2;
+	op.params[0].memref.offset = shm->size - (shm->size / 2) + 1;
+
+	/*
+	 * Since reference are not valid the invocation request should NOT
+	 * reach the TA whatever the status. Yet we expect noboby gets hurt.
+	 */
+	ADBG_EXPECT_NOT(c, TEEC_ERROR_TARGET_DEAD,
```
jenswi-linaro: I think we should expect that the driver catches this, so expect whatever error code we should get from that one.

host/xtest/regression_1000.c
```diff
@@ -1381,24 +1386,51 @@ static void xtest_tee_test_1017(ADBG_Case_t *c)
...
+			TEEC_InvokeCommand(session, TA_OS_TEST_CMD_PARAMS, &op,
+					   &ret_orig));
+
+	ADBG_EXPECT_NOT(c, TEEC_ORIGIN_TRUSTED_APP, ret_orig);
```
jenswi-linaro: `TEEC_ORIGIN_COMMS` is expected

host/xtest/regression_1000.c
jenswi-linaro: Expecting `TEEC_ERROR_BAD_PARAMETERS` or `TEEC_ERROR_GENERIC` should be good. libteec is currently reporting all errors by `ENOMEM` as `TEEC_ERROR_GENERIC`. If we ever update `ioctl_errno_to_res()` with better granularity I expect that `EINVAL` would be translated to `TEEC_ERROR_BAD_PARAMETERS`.

host/xtest/regression_1000.c
jenswi-linaro: I thin libteec should try to only do the sanity checks that aren't done in the kernel. For two reasons:
- both avoid cluttering the code with redundant checks
- enable us to check that the kernel does the expected checks or there will be a real security problem.

jenswi-linaro:

etienne-lms:
squashed & tag applied.

jforissier:
Fails to build on HiKey?

jbech-linaro:
@etienne-lms is this something that we could finalize after we have released 3.2.0? It's been pending for a while now.

jenswi-linaro:
Can we try to see this through?

etienne-lms:
ok. I've rebased the changes.

Checkpatch would warn on line split (see below), yet I preserved the implementation for consistency inside `sdp_basic_runner_cmd_parser()`:
```bash
WARNING: quoted string split across lines
#177: FILE: host/xtest/sdp_basic.c:770:
+	verbose("\nSecure Data Path basic access: "
+		"Invoke TA with out of bounds buffer references\n");
```

etienne-lms:
force pushed with @jenswi-linaro R-b tag removed since it is a bit... old. Deserves a new review.

etienne-lms:
Added fixup commit for build warnings reported by IBART.

jenswi-linaro:

etienne-lms:
tag applied

PR #277, opened 2018-04-20
ta: don't include host_include/conf.mk

jenswi-linaro:
Don't include host_include/conf.mk when building the TA as many
unrelated variables are propagated. Some of those variables, for
instance, CFG_RESERVED_VASPACE_SIZE causes compile errors.

Relevant variables are already defined by the dev kit.

Needed by https://github.com/OP-TEE/optee_os/pull/2266

### DIFF
--- a/ta/Makefile
+++ b/ta/Makefile
@@ -8,7 +8,7 @@ ifeq ($(out-dir),)
 $(error invalid output directory (O=$(O)))
 endif
 
-include $(TA_DEV_KIT_DIR)/host_include/conf.mk
+include $(TA_DEV_KIT_DIR)/mk/conf.mk
 
 # Prevent use of LDFLAGS from the environment. For example, yocto exports
 # LDFLAGS that are suitable for the client applications, not for TAs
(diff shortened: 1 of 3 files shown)


jenswi-linaro:
Update

jforissier:
In the commit message: the sentence "Some of those variables, for instance, CFG_RESERVED_VASPACE_SIZE causes compile errors." was true for an intermediate version of https://github.com/OP-TEE/optee_os/pull/2266 only but hopefully not anymore so you should remove it.

jenswi-linaro:
Comment addressed,, squashed and tag applied

jforissier:
The HAB error is caused by the fact that the SDP TA (sdp_basic) is not build, the reason being that `CFG_SECURE_DATA_PATH` is not (yet) defined in the devkit conf.mk (until https://github.com/OP-TEE/optee_os/pull/2266 is merged).

I can confirm the SDP TA is correctly built in my environment with both PRs checked out.

PR #278, opened 2018-04-23
Regression suite case 8101

jenswi-linaro:
Test case for mbedTLS in https://github.com/OP-TEE/optee_os/pull/2260


jenswi-linaro:
Replaced by https://github.com/OP-TEE/optee_test/pull/279

PR #279, opened 2018-05-07
Adds 81xx series for mbedtls tests

jenswi-linaro:
Replaces https://github.com/OP-TEE/optee_test/pull/278


jenswi-linaro:
Needs https://github.com/OP-TEE/optee_os/pull/2260 and https://github.com/OP-TEE/build/pull/259 to enabled all tests

host/xtest/regression_8100.c
```diff
@@ -0,0 +1,29 @@
...
+	if (!ADBG_EXPECT_TEEC_SUCCESS(c, xtest_teec_open_session(
+					      &session, &crypt_user_ta_uuid,
+					      NULL, &ret_orig)))
+		return;
+	ADBG_EXPECT_TEEC_SUCCESS(c,
+		TEEC_InvokeCommand(&session, TA_CRYPT_CMD_MBED_SELF_TESTS, NULL,
+				   &ret_orig));
+	TEEC_CloseSession(&session);
+#else
+	Do_ADBG_Log("mbedTLS self tests not available, skipped");
```
jforissier: `CFG_TA_MBEDTLS_SELF_TEST not set, test skipped` would be (slightly) more helpful

cert/README.txt
```diff
@@ -0,0 +1,20 @@
+This is a Test Ceritificate Authority, only to be used for testing.
```
jforissier: Why not generate the certificates at build time? (basically put these commands in the Makefile).

ta/crypt/include/user_ta_header_defines.h
```diff
@@ -33,7 +33,7 @@
 #define TA_UUID TA_CRYPT_UUID
 
 #define TA_FLAGS		(TA_FLAG_USER_MODE | TA_FLAG_EXEC_DDR)
-#define TA_STACK_SIZE		(2 * 1024)
+#define TA_STACK_SIZE		(32 * 1024)
```
jforissier: Why? Needed for mbedtls? Then, should be in the first commit that uses mbedtls I think.

scripts/file_to_c.py
```diff
@@ -0,0 +1,48 @@
...
+#
+# Copyright (c) 2018, Linaro Limited
+#
+
+def get_args():
+    import argparse
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--name', required=True,
+                        help='Name of variable in c file')
```
jforissier: Uppercase C

scripts/file_to_c.py
```diff
@@ -0,0 +1,48 @@
...
+    parser.add_argument('--out', required=True, help='Name of c file')
```
jforissier: Uppercase C

host/xtest/Makefile
```diff
@@ -205,6 +204,20 @@ $(foreach v, $(GCM_NIST_VECTORS_ENCRYPT), $(eval $(call \
...
+	@echo '  GEN     $$@'
+	$(q)../../scripts/file_to_c.py --inf $$< --out $$@ --name $(1)
+
+$(CURDIR)/regression_8100.c: $(out-dir)/xtest/$(1).h
+endef
+
+$(eval $(call embed-file,regression_8100_ca_crt,../../cert/ca.crt))
+$(eval $(call embed-file,regression_8100_mid_crt,../../cert/mid.crt))
+$(eval $(call embed-file,regression_8100_my_crt,../../cert/my.crt))
+$(eval $(call embed-file,regression_8100_my_csr,../../cert/my.csr))
```
vchong: So the aosp ci detected an actual build regression this time.

```
optee_test/host/xtest/regression_8100.c:23:10: fatal error: 'regression_8100_ca_crt.h' file not found
```

Not sure if we can call a python script in `Android.mk`. @liuyq any ideas?

Or can these header files be pre-generated?

ta/crypt/include/user_ta_header_defines.h
jenswi-linaro: Yes, this is needed for mbedtls. Usually we try to do such things in separate commits.

cert/README.txt
jenswi-linaro: It seems a bit more safe and predictable to have pre-generated keys and certificates. Then we also know that we're testing with the same input each time.

jenswi-linaro:
Update. Tried to generate the headers in `Android.mk`

Android.mk
```diff
@@ -56,6 +57,19 @@ ifeq ($(CFG_SECURE_DATA_PATH),y)
 srcs += sdp_basic.c
 endif
 
+define my-embed-file
+$(TARGET_OUT_HEADERS)/$(1).h: $(2)
```
vchong: Where's `TARGET_OUT_HEADERS` defined? Maybe `$(LOCAL_PATH)/host/xtest`?

@liuyq please comment as necessary

Android.mk
```diff
@@ -56,6 +57,19 @@ ifeq ($(CFG_SECURE_DATA_PATH),y)
...
+	@echo '  GEN     $$@'
+	@scripts/file_to_c.py --inf $$< --out $$@ --name $(1)
+
+$(call LOCAL_PATH)/host/xtest/regression_8100.c: $(TARGET_OUT_HEADERS)/$(1).h
```
vchong: Same as above

Android.mk
```diff
@@ -56,6 +57,19 @@ ifeq ($(CFG_SECURE_DATA_PATH),y)
...
+endef
+
+$(eval $(call my-embed-file,regression_8100_ca_crt,cert/ca.crt))
+$(eval $(call my-embed-file,regression_8100_mid_crt,cert/mid.crt))
+$(eval $(call my-embed-file,regression_8100_my_crt,cert/my.crt))
+$(eval $(call my-embed-file,regression_8100_my_csr,cert/my.csr))
```
vchong: Seems like relative path `cert/*` doesn't work in aosp build env [1]. Can you try `$(LOCAL_PATH)/cert/*` please?

[1] https://ci.linaro.org/job/android-hikey-optee-o/204/consoleText

ta/crypt/mbed_taf.c
```diff
@@ -0,0 +1,343 @@
...
+	if (!ret)
+		return TEE_SUCCESS;
+
+	if (ret != MBEDTLS_ERR_BASE64_BUFFER_TOO_SMALL) {
+		EMSG("mbedtls_x509write_crt_pem: failed: %#x", ret);
+		return TEE_ERROR_GENERIC;
+	}
+
+	/*
+	 * We where called with a too small buffer, let's find out how
```
jbech-linaro: s/where/were/

Android.mk
jenswi-linaro: I think it's defined by the build system

jenswi-linaro:
Update

Android.mk
```diff
@@ -56,6 +57,19 @@ ifeq ($(CFG_SECURE_DATA_PATH),y)
 srcs += sdp_basic.c
 endif
 
+define my-embed-file
+$(TARGET_OUT_HEADERS)/$(1).h: $(LOCAL_PATH)/$(2)
+	@echo '  GEN     $$@'
+	@scripts/file_to_c.py --inf $$< --out $$@ --name $(1)
```
vchong: Sorry, missed this one. `$(LOCAL_PATH)/scripts/file_to_c.py`.

jenswi-linaro:
Update

vchong:
Yay~ ci passed! Thanks @jenswi-linaro!

jforissier:
8103 fails for me on HiKey960:
```
# xtest _81
[snip]
* regression_8103 TA mbedTLS process certificate request
D/TC:0 tee_ta_init_pseudo_ta_session:274 Lookup pseudo TA cb3e5ba0-adf1-11e0-998b-0002a5d5c51b
D/TC:0 load_elf:770 Lookup user TA ELF cb3e5ba0-adf1-11e0-998b-0002a5d5c51b (Secure Storage TA)
D/TC:0 load_elf:770 Lookup user TA ELF cb3e5ba0-adf1-11e0-998b-0002a5d5c51b (REE)
D/TC:0 load_elf_from_store:738 ELF load address 0x4000d000
D/TC:0 tee_ta_init_user_ta_session:946 Processing relocations in cb3e5ba0-adf1-11e0-998b-0002a5d5c51b
regression_8100.c:205: Expression "i > 0" (0 > 0) is false
regression_8100.c:241: check(c, trusted, untrusted, crt) has an unexpected value: 0x0 = false, expected 0x1 = true
D/TC:0 tee_ta_close_session:380 tee_ta_close_session(0x3f06b780)
D/TC:0 tee_ta_close_session:399 Destroy session
D/TC:0 tee_ta_close_session:425 Destroy TA ctx
  regression_8103 FAILED
[snip]
Result of testsuite regression filtered by "_81":
regression_8101 OK
regression_8102 OK
regression_8103 FAILED first error at regression_8100.c:205
```

jenswi-linaro:
Update

jforissier:

ta/crypt/include/user_ta_header_defines.h
jforissier: > Usually we try to do such things in separate commits.

What is the benefit? The drawback I see is it's harder to tell why we had to increase the size. It it's part of a commit that calls mbedtls then it's obvious why.

I'm be happy if you just mention the reason in the commit log however.

ta/crypt/include/mbed_taf.h
```diff
@@ -0,0 +1,12 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/* Copyright (c) 2018, Linaro Limited */
+
+#ifndef __MBED_TAF_H
+#define __MBED_TAF_H
+
+#include <tee_internal_api.h>
+
+TEE_Result ta_entry_mbed_self_tests(uint32_t param_type,
```
jforissier: I suggest `s/mbed/mbedtls/` and `s/MBED/MBEDTLS/` everywhere (function names, file names, macros) because mbed means nothing. What do you think?

cert/README.txt
jforissier: I'm still not convinced. I think we can safely assume that we have a working OpenSSL implementation and it would reduce the amount of opaque data committed into the repository.
But it's not a strong opinion, I'm OK with the current proposal.

host/xtest/regression_8100.c
```diff
@@ -0,0 +1,316 @@
...
+		TEEC_InvokeCommand(&session, TA_CRYPT_CMD_MBED_SELF_TESTS, NULL,
+				   &ret_orig));
+	TEEC_CloseSession(&session);
+#else
+	Do_ADBG_Log("CFG_TA_MBEDTLS_SELF_TEST not set, test skipped");
+#endif
+}
+ADBG_CASE_DEFINE(regression, 8101, test_8101, "TA mbedTLS self tests");
+
+static int __printf(2, 3) myasprintf(char **strp, const char *fmt, ...)
```
jforissier: Why not use `asprintf()` from `stdio.h>`? Android I suppose?

jenswi-linaro: It was actually missing in the libc provided by buildroot too.

ta/crypt/include/mbed_taf.h
jenswi-linaro: Updated

jenswi-linaro:
Updated and tags applied

PR #282, opened 2018-05-24
xtest: Makefile: link against OpenSSL if MBed TLS is enabled in TA

jforissier:
Commit 950ea1cda6d3 ("regression: add case 8103") introduces a test that
needs OpenSSL to verify a certificate generated by a TA with Mbed TLS. It
updates xtest's CMakeLists.txt to link against OpenSSL when found, but the
regular Makefile is not updated.

This commit adds support for the new test to the non-CMake build by
linking xtest against the OpenSSL static library when CFG_TA_MBEDTLS=y.


etienne-lms:

PR #284, opened 2018-05-31
Add dynamically linked TA test

jforissier:


ta/os_test_lib/include/os_test_lib.h
```diff
@@ -0,0 +1,31 @@
...
+void os_test_shlib_panic(void);
+
+#else
+
+#include <compiler.h>
+#include <tee_internal_api.h>
+
+static int os_test_shlib_add(int a __unused, int b __unused)
+{
+	TEE_Panic(0);
```
jbech-linaro: So if have haven't enabled `CFG_DYNLINK_TA` then this (and `os_test_shlib_panic`) test should bail out? I.e., the `ifdef` around `xtest_tee_test_1019` should ensure that this never will be called, but in case someone does so anyways from somewhere, then we just bail out? Is that the intention here?

jforissier: Exactly.

jbech-linaro: Ok, with that you have my RB.

jbech-linaro:

jforissier:
I still need to figure out the CMake and Android aspects. I suppose it won't work magically since there is a dependency between the `os_test` TA and the shared library.

jforissier:
With CMake, how to make sure that `ta/os_test_lib` is built before `ta/os_test`? Should I express this dependency in `ta/CMakeLists.txt`, and if so, how?

jforissier:
@vchong this will certainly not build on Android because the "TA dev kit" provides no recipe to build a shared library to start with. What would be the easiest way to disable this test? Force `CFG_TA_DYNLINK=n` somehow in the main `Android.mk`?

jforissier:
OK, I think I've managed to address both CMake/buildroot and Android.

`Tested-by: Jerome Forissier <jerome.forissier@linaro.org> (QEMU Makefile/CMakelists.txt)`

@vchong could you please test on Android? (`xtest` should pass and `xtest 1019` should do nothing)

jforissier:
@vchong confirmed that this PR does not break the AOSP build (thanks!) so it can be merged.

Issue #285, opened 2018-06-04
"for_gp/include: No such file or directory" issue


igoropaniuk:
Commit that causes this issue https://github.com/OP-TEE/optee_test/commit/43d58a5fce8fd968eb00fd8a7a94039a78b1318a

```
make -C /home/igoropaniuk/linaro/reps/repo/default/build/../optee_test \
	CROSS_COMPILE_HOST="/home/igoropaniuk/linaro/reps/repo/default/build/../toolchains/aarch64/bin/aarch64-linux-gnu-" \
	CROSS_COMPILE_TA="/home/igoropaniuk/linaro/reps/repo/default/build/../toolchains/aarch32/bin/arm-linux-gnueabihf-" \
	TA_DEV_KIT_DIR="/home/igoropaniuk/linaro/reps/repo/default/build/../optee_os/out/arm/export-ta_arm32" \
	OPTEE_CLIENT_EXPORT="/home/igoropaniuk/linaro/reps/repo/default/build/../optee_client/out/export"
make[1]: Entering directory '/home/igoropaniuk/linaro/reps/repo/default/optee_test'
make[2]: Entering directory '/home/igoropaniuk/linaro/reps/repo/default/optee_test/host/xtest'
  CC      /home/igoropaniuk/linaro/reps/repo/default/optee_test/host/xtest/regression_4000.c
cc1: error: for_gp/include: No such file or directory [-Werror=missing-include-dirs]
cc1: all warnings being treated as errors

```

PR #286, opened 2018-06-04
xtest: OpenSSL fix

jforissier:
xtest needs to use OpenSSL in two cases:
- When the GlobalPlatform tests are enabled (CFG_GP_PACKAGE_PATH is set),
- When the MbedTLS tests are enabled (CFG_TA_MBEDTLS=y).
For the first case, it is the "make patch" target that takes care of
copying the header files under host/xtest/for_gp/include/openssl.
For the second case however, the copy does not occur and only the -I flag
is added by [1], resulting in the following error:

  cc1: error: for_gp/include: No such file or directory [-Werror=missing-include-dirs]

Make sure the OpenSSL header files are made available in both cases by
copying the files using a $(shell ...) command. It *is* ugly, but arguably
not much more than the previous patch-openssl target.

In addition, the clean target will now remove the copied files.

Fixes: [1] 43d58a5fce8f ("xtest: Makefile: link against OpenSSL if MBed TLS is enabled in TA")
Fixes: https://github.com/OP-TEE/optee_test/issues/285
Reported-by: Igor Opaniuk <igor.opaniuk@linaro.org>


igoropaniuk:
`Tested-by: Igor Opaniuk <igor.opaniuk@linaro.org>`

jenswi-linaro:
Pardon my ignorance, but why are we copying anything at all? Why doesn't `-I<some-path>` work?

jforissier:
@jenswi-linaro

It's because that would be `-I/usr/include`. We have for instance `#include <openssl/bn.h>` and the file is `/usr/include/openssl/bn.h`. Using this `-I` would then pull files that are specific to the build system (x86 or x86_64 typically), resulting in compilation errors. That's the first thing I tried ;) I should have mentioned the reason in the commit log.

Perhaps we could avoid the problem by using `#include <bn.h>` instead, and adding `-I/usr/include/openssl`. We would need to change the buildroot `-I` flag accordingly, but that might be a better solution. I can try that tomorrow, what do you think?

jenswi-linaro:
This copying stuff is broken and only works as a coincidence, the headers used when compiling the OpenSSL lib (or at least for the same version) should be used or we'll face problems hard to track down from time to time.

jforissier:
I agree. It has been broken since day one.
How about importing the headers into the Git repo then? We do have the libs already.

jenswi-linaro:
I was considering that too, please go ahead and import the headers. I guess this is only needed when compiling with the old legacy make files. Or is it needed for Android too?

jforissier:
OK, will do.
AFAICT the Android build does not support the GP test suite I think, and it will not set OPENSSL_FOUND either so it should be OK.

jforissier:
Update

jenswi-linaro:
"xtest: use imported OpenSSL" seem to update libcrypto.a

jforissier:
Yes, it removes the old libraries that are not used anymore. I have moved that part to a separate commit.

jenswi-linaro:

igoropaniuk:
`Tested-by: Igor Opaniuk <igor.opaniuk@linaro.org> (poplar)`

PR #288, opened 2018-06-07
ta: os_test_lib: add aosp support

vchong:
depends on https://github.com/OP-TEE/optee_os/pull/2380

@liuyq please ack!

### DIFF
--- a/ta/os_test/Android.mk
+++ b/ta/os_test/Android.mk
@@ -1,4 +1,7 @@
 LOCAL_PATH := $(call my-dir)
 
 local_module := 5b9e0e40-2636-11e1-ad9e-0002a5d5c51b.ta
+ifeq ($(CFG_TA_DYNLINK),y)
+local_module_deps := ffd2bded-ab7d-4988-95ee-e4962fff7154.ta
+endif
 include $(BUILD_OPTEE_MK)
--- a/host/xtest/regression_1000.c
+++ b/host/xtest/regression_1000.c
@@ -61,7 +61,7 @@ static void xtest_tee_test_1015(ADBG_Case_t *Case_p);
 static void xtest_tee_test_1016(ADBG_Case_t *Case_p);
 static void xtest_tee_test_1017(ADBG_Case_t *Case_p);
 static void xtest_tee_test_1018(ADBG_Case_t *Case_p);
-#if defined(CFG_TA_DYNLINK) && !defined(__ANDROID__)
+#if defined(CFG_TA_DYNLINK)
 static void xtest_tee_test_1019(ADBG_Case_t *Case_p);
 #endif
 
@@ -97,7 +97,7 @@ ADBG_CASE_DEFINE(regression, 1017, xtest_tee_test_1017,
 		"Test coalescing memrefs");
 ADBG_CASE_DEFINE(regression, 1018, xtest_tee_test_1018,
 		"Test memref out of bounds");
-#if defined(CFG_TA_DYNLINK) && !defined(__ANDROID__)
+#if defined(CFG_TA_DYNLINK)
 ADBG_CASE_DEFINE(regression, 1019, xtest_tee_test_1019,
 		"Test dynamically linked TA");
 #endif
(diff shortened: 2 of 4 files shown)


ta/os_test/Android.mk
```diff
@@ -1,4 +1,7 @@
 LOCAL_PATH := $(call my-dir)
 
 local_module := 5b9e0e40-2636-11e1-ad9e-0002a5d5c51b.ta
+ifeq ($(CFG_TA_DYNLINK),y)
+local_dependency_ta := ffd2bded-ab7d-4988-95ee-e4962fff7154.ta
```
jforissier: That is not correct. The true dependency is: `<out-dir-for-os_test>/$(local_module).elf: <out-dir-for-os_test_lib>/libos_test.so`. Not describing the exact dependencies might result in build errors.

liuyq: is it better to make the compiling of libos_test.so a separate module, and set the os_test ta and os_test_lib ta and other tas to depends on libos_test.so?

Now it seems that the os_test ta depends on the os_test_lib ta, and that os_test_lib ta depends on the libos_test.so, so making  os_test ta depends on the os_test_lib ta finally meets the requirement of os_test ta .

BTW, two small questions here:
1. is libos_test.so need to be integrated into the device as the ta files?
2. not sure if my understanding above is right, if without libos_test.so the compiling of os_test ta will fail, then it might be better to not add the check of ifeq ($(CFG_TA_DYNLINK),y). If libos_test.so is only required when  CFG_TA_DYNLINK is enabled, then the check is necessary here.

jforissier: > [...] so making os_test ta depends on the os_test_lib ta finally meets the requirement of os_test ta 

No, it is not equivalent. Try this: `make; rm <out-path>/libos_test.so; touch <src-path>/ta/os_test_lib/os_test_lib.c; make`. It will fail (unless you somehow have added `libos_test.so` to the 'all' target, which should not be the case, see below).

> 1. is libos_test.so need to be integrated into the device as the ta files?

No, it is needed only at build (link) time. The `.ta` file is a signed version of the `.so`, OP-TEE OS can only load signed files.

> 2. [...] it might be better to not add the check of ifeq ($(CFG_TA_DYNLINK),y)

Well, the library (`ffd2bded-ab7d-4988-95ee-e4962fff7154.ta`) is not used if `CFG_TA_DYNLINK != y`, so it seems more logical not to build it at all.

vchong: I'll try to make the compiling of libos_test.so a separate module.

vchong:
Comments addressed

ta/os_test/Android.mk
vchong: @jforissier Thinking more about this, maybe the fixup patch isn't needed. Similar to how `ta/Makefile` specifies `os_test_lib` before `os_test` in `TA_DIRS` to get the build order, and hence dependency, right, specifying `local_dependency_ta := ffd2bded-ab7d-4988-95ee-e4962fff7154.ta` for `local_module := 5b9e0e40-2636-11e1-ad9e-0002a5d5c51b.ta` is really just to get the AOSP build system to call `os_test_lib/Makefile` before `os_test/Makefile`. It isn't really specifying a Makefile target dependency in the strict sense of the example above (`<out-dir-for-os_test>/$(local_module).elf: <out-dir-for-os_test_lib>/libos_test.so`), i.e. we're not really specifying `5b9e0e40-2636-11e1-ad9e-0002a5d5c51b.ta: ffd2bded-ab7d-4988-95ee-e4962fff7154.ta`.

jforissier: Hmmm, you may very well be right :astonished: 

I completely missed the fact that AOSP runs `$(MAKE)` for each TA or library directory. I was incorrectly thinking that the AOSP build was much more different, that all the rules were imported in a single Makefile (like what happens in optee_os).

Since the `$(MAKE)` line https://github.com/OP-TEE/optee_os/blob/master/mk/aosp_optee.mk#L94 does not specify any target, it will run `all:` and thus we should get all our files (`.so`, `.ta`), indeed.

Sorry for the bad advice :(

vchong: Yep, np. Your advice/comments are always appreciated! ;) Will remove fixup and add tag. Thanks!

vchong:
`s/local_dependency_ta/local_module_deps`, removed fixup, added tag and rebased. Thanks!

PR #290, opened 2018-06-19
crypto: add ifdef guards to seed_rng_taf header

Emantor:
Without these guards compiling optee_test fails because it can't find the
system_pta.h file.

Signed-off-by: Rouven Czerwinski <rouven@czerwinskis.de>

### DIFF
--- a/ta/crypt/include/seed_rng_taf.h
+++ b/ta/crypt/include/seed_rng_taf.h
@@ -6,8 +6,12 @@
 #ifndef SEED_RNG_TAF_H
 #define SEED_RNG_TAF_H
 
+#ifdef CFG_SYSTEM_PTA
+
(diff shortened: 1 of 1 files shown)


jforissier:
Hi @Emantor,

Thanks for the patch. I suspect you're building optee_test with a "TA dev kit" produced by an older optee_os, aren't you?
If so, this is simply not supported. optee_test and optee_os have to be kept in sync. That is why we did not bother adding the #ifdef guard in the first place.
If however the build fails for another reason, then a fix is needed -- please give more details.
Thanks!

Emantor:
Yes, you are right. My local master branch was stuck in the beginning of may, works with latest optee master. Thanks :+1:

PR #291, opened 2018-06-19
xtest: Makefile: fix COMPILE_NS_USER not being set by default

denix0:
The build for TI ARM32 platforms attempts linking with aarch64 libcrypto.a
and breaks, even when CFG_ARM32_core is set in the optee-os TA devkit.

Fix it by moving setting of COMPILE_NS_USER based on CFG_ARM32_core outside
of ifdef CFG_GP_PACKAGE_PATH, as it's also needed for CFG_TA_MBEDTLS.

Signed-off-by: Denys Dmytriyenko <denys@ti.com>
Cc: Andrew F. Davis <afd@ti.com>

### DIFF
--- a/host/xtest/Makefile
+++ b/host/xtest/Makefile
@@ -24,6 +24,13 @@ OBJCOPY		?= $(CROSS_COMPILE)objcopy
 OBJDUMP		?= $(CROSS_COMPILE)objdump
 READELF		?= $(CROSS_COMPILE)readelf
 
+# by default, the client application is compiled as the kernel of optee-os
+ifeq ($(CFG_ARM32_core),y)
+COMPILE_NS_USER ?= 32
+else
+COMPILE_NS_USER ?= 64
+endif
+
 # OpenSSL is used by GP tests series 8500 and Mbed TLS test 8103
 ifneq (,$(filter y,$(CFG_GP_PACKAGE_PATH) $(CFG_TA_MBEDTLS)))
 CFLAGS += -I../openssl/include -DOPENSSL_FOUND=1
@@ -121,13 +128,6 @@ CFLAGS += -I../../ta/GP_TTA_answerSuccessTo_OpenSession_Invoke
 CFLAGS += -I../../ta/GP_TTA_check_OpenSession_with_4_parameters
 CFLAGS += -I../../ta/GP_TTA_testingClientAPI
 
-
-# by default, the client application is compiled as the kernel of optee-os
-ifeq ($(CFG_ARM32_core),y)
-COMPILE_NS_USER ?= 32
-else
-COMPILE_NS_USER ?= 64
-endif
 
 endif
 


jforissier:

glneo:
While this is not wrong, I don't think it gets to the core of the issue.

The first issue we still have is that most do not set CFG_ARM32_core to "y" for when building 32 bit OP-TEE, it is the default. You set CFG_ARM64_core to "y" when you want 64bit. So the logic should be:

```
ifeq ($(CFG_ARM64_core),y)
COMPILE_NS_USER ?= 64
else
￼COMPILE_NS_USER ?= 32
endif
```

And even when building 64bit OP-TEE, OP-TEE builds both 64 and 32 bit TA devkits. So we still do not know which one is being pointed to by "TA_DEV_KIT_DIR" here when building xtest.

COMPILE_NS_USER should be set based on what TA_DEV_KIT_DIR we are building against.

I think all that can be added later though, so for this patch:

```
Acked-by: Andrew F. Davis <afd@ti.com>
```

jforissier:
@glneo

In fact, there is absolutely no relationship between `COMPILE_NS_USER` and the arch of the TEE kernel or the TAs (TA devkit).  Note that we're talking about the client, non-secure application here. It can be 64 bits while everything else is 32.

> COMPILE_NS_USER should be set based on what TA_DEV_KIT_DIR we are building against.

I suppose you mean the optee_client library, not TA dev kit. In which case I agree.

denix0:
@glneo that would be optee-os change to clean up the defaults there first...

For now I modified optee-os recipe to build with CFG_ARM64_core="y" on Aarch64 and CFG_ARM32_core="y" otherwise. Also, only 1 type of TA devkit is built for us.

I'm fine if anyone wants to take it further and do more thorough cleanup. I just needed to fix this breakage with minimal changes.

glneo:
@denix0 That's fine, as long as it builds again. But for K3, we always build the 32bit support, and also build 64bit TA kit when building for 64bit TEE. https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/plat-k3/conf.mk#L23

@jforissier I mean this project, xtest, should build 32-bit TAs if you set TA_DEV_KIT_DIR to a 32-bit dev kit. (same for 64-bit). optee_client bit-ness only matters for what normal world kernel you are ruining, the TAs can be either bit-ness. I suppose this is why we have CROSS_COMPILE_HOST and CROSS_COMPILE_TA for this project.

jforissier:
> I mean this project, xtest, should build 32-bit TAs if you set TA_DEV_KIT_DIR to a 32-bit dev kit. (same for 64-bit).

Agreed.

> optee_client bit-ness only matters for what normal world kernel you are ruining, the TAs can be either bit-ness. I suppose this is why we have CROSS_COMPILE_HOST and CROSS_COMPILE_TA for this project.

Yes, so I guess we should get rid of `COMPILE_NS_USER` and detect the bit-ness of `CROSS_COMPILE_HOST` instead. Simpler and no risk of inconsistency.

glneo:
@jforissier yes, having both COMPILE_NS_USER and CROSS_COMPILE_HOST allows for them to be different, which is never the right thing to have.

etienne-lms:
ping!
looks like this is needed!
This fixes `COMPILE_NS_USER` being used before being set.

jforissier:
@etienne-lms I don't see how it is needed, since it is actually incorrect? As I said above, what is the relationship between `CFG_ARM32_core` and `COMPILE_NS_USER`? They are completely orthogonal.

etienne-lms:
At least this change make the inline statement true:
> `# by default, the client application is compiled as the kernel of optee-os`

Without this patch, this statement is false.
Change build env to add `COMPILE_NS_USER=32` (or `=64`), which is  is a bit hacky, at least it should be something like `CFG_COMPILE_NS_USER=...`.

jforissier:
>At least this change make the inline statement true:
>> # by default, the client application is compiled as the kernel of optee-os

OK, you're right. At least by moving the code up we make things consistent. That's why I Acked the patch initially. But, this statement about the kernel of OP-TEE really annoys me, it makes no sense to begin with. So rather than just moving it up in the Makefile, I'd rather pick a sensible default... Here is what I had in mind:

```
define CC_ARCH
$(shell $(1) -v 2>&1 | grep Target | sed 's/Target: \([^-]*\).*/\1/')
endef

define CC_BITS
$(if $(filter arm, $(1)),32,$(if $(filter aarch64, $(1)),64,unknown-arch))
endef

COMPILE_NS_USER ?= $(call CC_BITS, $(call CC_ARCH, $(CC)))
```

Basically if "$(CC) -v" outputs arm-something, we use 32 and if it's aarch64-something, we use 64. Makes more sense, and should fix the initial problem, too.

What do you think? Do you still prefer the "quick and dirty" fix?

etienne-lms:
Agree with you. The current hack aims at finding a way if env does not provide 32b/64b directive for the host tools. Looking at toolchain capabilities looks nice.

I wonder if all arm toolchains obey to this `arm`/`aarch64` labeling and if there is no single compiler tool that can generate 32b and 64b compilers according to a CLI option, like `--aarch32`/`--aarch64`.
That said, I'm in favor for your way since up to my knowledge, there is no single cross compiler that builds both 32b and 64b Arm code.

jbech-linaro:
> That said, I'm in favor for your way since up to my knowledge, there is no single cross compiler that builds both 32b and 64b Arm code.

Clang? :)

jforissier:
> Clang?

Yep, tested with that too ;-)

etienne-lms:
I think this change should hit 3.5.0 as it fixes `COMPILE_NS_USER` support.
A later change could default `COMPILE_NS_USER` to toolchain info, as proposed by @jforissier.

jbech-linaro:
@denix0 , can you please add Etienne's Reviewed-By tag (force push) and rebase?

glneo:
@jforissier could you post your toolchain detection patch, this is still broken and I don't think this patch needs to go in at all as long as your fix defines COMPILE_NS_USER before it is used.

jforissier:
@glneo there you go: https://github.com/OP-TEE/optee_test/pull/372

jforissier:
I'm closing this since #372 is now merged.

PR #293, opened 2018-06-21
Android.mk: make xtest installed in the vendor partition

liuyq:
so that we could make optee work with Treble enabled builds too

Acked-by: Victor Chong <victor.chong@linaro.org>

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -19,7 +19,10 @@ TA_DEV_KIT_DIR ?= ../invalid_include_path
 
 include $(CLEAR_VARS)
 LOCAL_MODULE := xtest
+LOCAL_VENDOR_MODULE := true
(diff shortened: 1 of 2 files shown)


vchong:
ack

@petegriffin fyi

liuyq:
The changes to support Treble enabled builds are as following:
https://github.com/OP-TEE/optee_client/pull/119
https://github.com/OP-TEE/optee_test/pull/293
https://github.com/OP-TEE/optee_os/pull/2413
https://android-review.linaro.org/#/c/18404/ (this change is only for treble support, need to apply other old changes on hikey device repository as well)

tested with xtest command on  hikey + aosp master + 4.14 kernel build on local side.

Please help to check if it's OK to have them merged, thanks in advance!

PR #294, opened 2018-06-22
xest: OpenSSL may be provided by the system

etienne-lms:
Introduce build directive `OPTEE_OPENSSL_EXPORT` in GNU make script to get OpenSSL from the target rather than using the provided builds of the library.

### DIFF
--- a/host/xtest/Makefile
+++ b/host/xtest/Makefile
@@ -26,13 +26,19 @@ READELF		?= $(CROSS_COMPILE)readelf
 
 # OpenSSL is used by GP tests series 8500 and Mbed TLS test 8103
 ifneq ($(CFG_GP_PACKAGE_PATH)$(filter y,$(CFG_TA_MBEDTLS)),)
-CFLAGS += -I../openssl/include -DOPENSSL_FOUND=1
+CFLAGS += -DOPENSSL_FOUND=1
+ifneq ($(OPTEE_OPENSSL_EXPORT),)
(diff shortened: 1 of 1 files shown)


etienne-lms:
This is intended for rc2.
I have a setup where the prebuilt **libcrypto.a** fails to be loaded at runtime (illegal instruction). The system can embed OpenSSL **libcrypto.so** to fix the issue.

host/xtest/Makefile
```diff
@@ -26,13 +26,19 @@ READELF		?= $(CROSS_COMPILE)readelf
...
+else #OPTEE_OPENSSL_EXPORT
+CFLAGS += -I../openssl/include
 ifeq ($(COMPILE_NS_USER),32)
 LDFLAGS += ../openssl/lib/arm/libcrypto.a -ldl
 else
 LDFLAGS += ../openssl/lib/aarch64/libcrypto.a -ldl
 endif
-endif
+endif #OPTEE_OPENSSL_EXPORT
+endif #require OenSSL
```
jenswi-linaro: OpenSSL

jenswi-linaro:
In the commit message, please change make script to makefile.
With that:

host/xtest/Makefile
etienne-lms: thanks

etienne-lms:
Comment and commit comment fixed.
Tag applied.

PR #295, opened 2018-06-22
regression: 400x: test all subcases regardless of return value from TEE Cleint API.

mozoko:
regression: 400x: test all subcases regardless of return value from TEE Cleint API.

In subcases of regression 40xx, if TEE Client API returns except TEEC_SUCCESS, xtest jumps to TEEC_CloseSession immediately.
So, the rest subcases are not tested.

This problem occurs, for example, when the parameters in the subcase are not supported by the cryptographic library in OPTEE OS.
In this case, TEE Client API returns TEEC_ERROR_NOT_SUPPORTED or TEEC_ERROR_BAD_PARAMETERS and the xtest closes the session leaving the rest subcases untested.

To avoid this, the xtest should jump to Do_ADBG_EndSubCase in the loop of subcases if TEE Client API returns except TEEC_SUCCESS.
This makes it possible to test all subcases regardless of the return code from TEE Client API.

To test this change, I changed the return value of ta_crypt_cmd_allocate_operation and ta_crypt_cmd_allocate_transient_object
to TEEC_ERROR_NOT_SUPPORTED and run "xtest -l 15 regression_4".
I confirmed all subcases are tested.

Signed-off-by: Akira Shibakawa <akira.shibakawa@miraclelinux.com>

### DIFF
--- a/host/xtest/regression_4000.c
+++ b/host/xtest/regression_4000.c
@@ -991,23 +991,23 @@ static void xtest_tee_test_4001(ADBG_Case_t *c)
 			ta_crypt_cmd_allocate_operation(c, &session, &op1,
 							hash_cases[n].algo,
 							TEE_MODE_DIGEST, 0)))
-			goto out;
+			goto endsub;
 
 		if (!ADBG_EXPECT_TEEC_SUCCESS(c,
 			ta_crypt_cmd_allocate_operation(c, &session, &op2,
 							hash_cases[n].algo,
 							TEE_MODE_DIGEST, 0)))
-			goto out;
+			goto endsub;
 
 		if (!ADBG_EXPECT_TEEC_SUCCESS(c,
 			ta_crypt_cmd_digest_update(c, &session, op1,
 						   hash_cases[n].in,
 						   hash_cases[n].in_incr)))
-			goto out;
+			goto endsub;
 
 		if (!ADBG_EXPECT_TEEC_SUCCESS(c,
 			ta_crypt_cmd_copy_operation(c, &session, op2, op1)))
-			goto out;
+			goto endsub;
 
 		out_size = sizeof(out);
 		memset(out, 0, sizeof(out));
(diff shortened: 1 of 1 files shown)


etienne-lms:
Looks good. It seems regression tests 4003, 4005, 4006 and 4009 could get the same changes.
Minor: maybe each sub test should open/close a session so that session is in good state when the test starts.

mozoko:
> Looks good. It seems regression tests 4003, 4005, 4006 and 4009 could get the same changes.
Yes. So, I've changed regression tests 400[1-3], 400[5-7] and 4009.
I'm sorry, I should have commented which tests I changed.

host/xtest/regression_4000.c
```diff
@@ -4517,11 +4516,9 @@ static void xtest_test_keygen_dh(ADBG_Case_t *c, TEEC_Session *session)
...
 		}
 
-		if (!ADBG_EXPECT_TRUE(c,
-			generate_and_test_key(c, session, TEE_TYPE_DH_KEYPAIR,
-				*key_types[n].private_bits,
-				key_types[n]. key_size, params, param_count)))
-			break;
+		generate_and_test_key(c, session, TEE_TYPE_DH_KEYPAIR,
+			*key_types[n].private_bits,
+			key_types[n]. key_size, params, param_count);
```
jforissier: Here and below the result of `generate_and_test_key()` is not tested anymore. Why?

host/xtest/regression_4000.c
```diff
@@ -991,23 +991,23 @@ static void xtest_tee_test_4001(ADBG_Case_t *c)
...
 							TEE_MODE_DIGEST, 0)))
-			goto out;
+			goto endsub;
 
 		if (!ADBG_EXPECT_TEEC_SUCCESS(c,
 			ta_crypt_cmd_allocate_operation(c, &session, &op2,
 							hash_cases[n].algo,
 							TEE_MODE_DIGEST, 0)))
-			goto out;
+			goto endsub;
```
jenswi-linaro: The `op1` allocated above is leaked

host/xtest/regression_4000.c
```diff
@@ -991,23 +991,23 @@ static void xtest_tee_test_4001(ADBG_Case_t *c)
...
 
 		if (!ADBG_EXPECT_TEEC_SUCCESS(c,
 			ta_crypt_cmd_digest_update(c, &session, op1,
 						   hash_cases[n].in,
 						   hash_cases[n].in_incr)))
-			goto out;
+			goto endsub;
```
jenswi-linaro: `op1` and `op2` above are leaked

PR #310, opened 2018-10-02
regression 6018: reduce even more blocks

jbech-linaro:
We are still getting out-of-memory errors in 6018 when we have enabled
the pager and are using RPMB on HiKey. By reducing the number of blocks
written to RPMB a bit more we get rid of the problem. The problem seems
to be that the heap gets too fragmented when we are using the pager, so
even though this solves the problem here and now, we would need to look
into whether we can do something about the fragmented heap in the long
run.

Related, we have reduced this in the past:
  20949ea945f2 ("regression 6018: reduce the number of blocks written")

Fixes: https://github.com/OP-TEE/optee_os/issues/2580

Tested-by: Joakim Bech <joakim.bech@linaro.org> (HiKey 6220)

### DIFF
--- a/host/xtest/regression_6000.c
+++ b/host/xtest/regression_6000.c
@@ -1827,7 +1827,7 @@ static void xtest_tee_test_6018_single(ADBG_Case_t *c, uint32_t storage_id)
 
 	if (storage_is(storage_id, TEE_STORAGE_PRIVATE_RPMB)) {
 		/* RPMB FS is a bit resource constrained */
-		num_blocks = 10;
+		num_blocks = 2;
 		block_size = 1024;
 	} else {
 		num_blocks = 20;


jforissier:
I'd be curious to know how many blocks we can use in 6018 if the `bget` allocation policy is set to `BestFit` (`lib/libutils/isoc/bget.c`). And also if the impact on allocation speed is reasonable.

jbech-linaro:
Tested using QEMU and only running 6018:
```
QEMU v8 BestFit enabled  --> numblocks = 22 : OK (23 and higher gives OOM).
QEMU v8 BestFit disabled --> numblocks = 14 : OK (15 and higher gives OOM).
```

jforissier:
As for speed, I have run comparisons with and without BestFit on my HiKey960. The results are inconclusive -- there is too much variation from one test run to the other. The difference (if any) is in the noise.

HiKey960, 64-bit TEE core, build options: CFG_TEE_CORE_LOG_LEVEL=2 CFG_WITH_PAGER=y.
"real" value reported by `time xtest` in seconds.
Baseline: 156, 144, 139, 193
With BestFit (TEE core only): 179, 142, 188

I'm starting to think that BestFit could be a better choice for our purpose.

jbech-linaro:
Right, I did a test run also, since the bget doc says that BestFit is much slower. But on QEMU v8 I couldn't see any difference. However, I usually don't put too much trust in figures coming from QEMU, since everything else you do on your computer affects the results.
```
BestFit enabled                                                                                                               
===                                                                                                                           
16409 subtests of which 0 failed                                                                                              
75 test cases of which 0 failed                                                                                               
0 test case was skipped                                                                                                       
TEE test application done!                                                                                                    
real    3m 48.66s                                                                                                             
user    0m 0.99s                                                                                                              
sys     3m 14.88s                                                                                                             
                                                                                                                              
BestFit disabled                                                                                                              
===                                                                                                                           
16409 subtests of which 0 failed                                                                                              
75 test cases of which 0 failed                                                                                               
0 test case was skipped                                                                                                       
TEE test application done!                                                                                                    
real    3m 53.71s                                                                                                             
user    0m 0.97s                                                                                                              
sys     3m 17.86s
```

> I'm starting to think that BestFit could be a better choice for our purpose.

I think so too.

jbech-linaro:
Looking much better with BestFit on HiKey also.
```
BestFit=y --> num_blocks = 17 is OK.
BestFit=n --> num_blocks = 3 is OK.
```
One thing that I noted is that if I for example run combinations (`BestFit=y` and `num_blocks=17` in all cases here).
```
1. xtest --> OK
2. xtest 6018, xtest 6018, xtest 6018, xtest 6018, xtest 6018,  .... ---> OK
3. xtest, xtest 6018 --> Fail OOM.
```
I.e, running entire xtest (passes), but running `6018` after that will make `num_blocks=17` fail. So probably the stuff after corrupts the heap enough to make it fail running it as standalone afterwards.

Having that said, by keeping `num_blocks = 10` (i.e., close this PR) and add a new patch that enables BestFit resolves the issue and I think this is a good compromise for now.

PR #311, opened 2018-10-16
add checkpatch scripts

wamserma:
copied the checkpatch-helper scripts from optee_os repo
fixed README.md to reflect checkpatch targets still missing

Signed-off-by: Markus S. Wamser <markus.wamser@mixed-mode.de>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -210,6 +210,15 @@ patch:
 	$(q) echo "Please define CFG_GP_PACKAGE_PATH" && false
 endif
 
+.PHONY: checkpatch checkpatch-staging checkpatch-working
+checkpatch: checkpatch-staging checkpatch-working
+
+checkpatch-working:
+	@./scripts/checkpatch.sh
+
+checkpatch-staging:
+	@./scripts/checkpatch.sh --cached
+
 install:
 	$(echo) '  INSTALL ${DESTDIR}/lib/optee_armtz'
 	$(q)mkdir -p ${DESTDIR}/lib/optee_armtz
--- a/scripts/checkpatch.sh
+++ b/scripts/checkpatch.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+if [ -z "$OPTEE_OS_PATH" ]; then
+   OPTEE_OS_PATH="../optee_os";
+fi
+
+source $OPTEE_OS_PATH/scripts/checkpatch.sh "$@"
--- a/typedefs.checkpatch
+++ b/typedefs.checkpatch
@@ -1,0 +1,2 @@
+TEE_Result
+TEE_UUID


scripts/checkpatch_inc.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+
+CHECKPATCH="${CHECKPATCH:-checkpatch.pl}"
+# checkpatch.pl will ignore the following paths
+CHECKPATCH_IGNORE=$(echo \
```
jbech-linaro: This part bring in things not relevant to optee_test.

wamserma: Yes, but doesn't seem to hurt either. I simply copied the files from the optee_os repo. Unless there is a pressing reason to branch off the code at this point, I'd suggest keeping it that way. For the time being this helps with maintenance issues. A long term goal could be to move the checkpatch (and probably other developer support scripts) to its own repo, so it can be re-used as a submodule at multiple places in the codebase and for TA development.

jforissier:
Duplicating a script doesn't help with maintainance, even if the file is not modified. What's the problem with using the files from the optee_os directory? Anyone working with optee_test is likely to have optee_os checked out, no?

wamserma:
@jforissier That would lead to using (relative) symlinks, which have the tendency to break. I strongly favour a separate repo for these scripts. This PR is more of an immediate cure to get things working smoothly.

jforissier:
What  is the problem with `PATH=/path/to/linux/scripts; /path/to/optee_os/scripts/checkpatch.sh [args...]`?

wamserma:
I don't have the right machine here to confirm, but iirc this made `checkpatch_inc.sh` complain because of the missing `typedefs.checkpatch`, when I called it this way (from `/path/to/optee_test/`).

wamserma:
@jforissier I modified the PR: now `scripts/checpatch.sh` dispatches to `$(OPTEE_OS_SRC)/scripts/checkpatch.sh` using a sensible default relative path.
I also added make targets and updated the Readme

README.md
```diff
@@ -1,162 +1,10 @@
...
-
-There are also targets for common use cases in the [Makefile](Makefile):
-
-```
-make checkpatch			#check staging and working area
-make checkpatch-staging #check staging area (added, but not committed files)
-make checkpatch-working #check working area (modified, but not added files)
-```
\ No newline at end of file
+[optee_test]: https://optee.readthedocs.io/building/gits/optee_test.html
```
jforissier: Unfortunately, this URL doesn't work, you have to keep the `en/latest/` part: `https://optee.readthedocs.io/en/latest/building/gits/optee_test.html`

wamserma: Took the URL from master, fixed in https://github.com/OP-TEE/optee_test/pull/393, will cherry-pick onto this PR.

jforissier:
This needs to be squashed as 3 commits:

- Redirect checkpatch.sh to ../optee_os
- Makefile: add checkpatch targets
- Remove documentation and point to readthedocs.io

jbech-linaro:
Please add Jerome's tag and squash patches where necessary, when done we can merge this.

wamserma:
Please merge #393 first, this will resolve the conflict before merging this branch.

wamserma:
Rebased onto current master and removed (now empty) commits.

jforissier:
One thing I should have stated from the beginning is: the `optee_test` source is far from being entirely formatted as per the kernel coding rules, and for consistency we often merge new code that does not follow the rules either. This PR can still be useful however.

PR #315, opened 2018-11-05
Android.mk: add required header files to LOCAL_ADDITIONAL_DEPENDENCIES

liuyq:
Android.mk: add required header files to LOCAL_ADDITIONAL_DEPENDENCIES

to avoid making them depenencies for regression_8100.c file
to workaround the PHONY target limitation.
https://android.googlesource.com/platform/build/+/HEAD/Changes.md#phony_targets

Otherwise, it will report "writing to readonly directory:" for regression_8100.c file
during the android aosp master build

As if we use $(LOCAL_PATH)/host/xtest/regression_8100.c as target,
it will break the 2nd rule of following:
    The target is a real file, but it's outside the output directories.
    All outputs from the build system should be within the output directory,
    otherwise m clean is unable to clean the build, and future builds may not work properly.
like described in the ".PHONY rule enforcement" section here:
https://android.googlesource.com/platform/build/+/master/Changes.md#phony_targets

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -65,7 +65,8 @@ $(TARGET_OUT_HEADERS)/$(1).h: $(LOCAL_PATH)/$(2)
 	@echo '  GEN     $$@'
 	@$(LOCAL_PATH)/scripts/file_to_c.py --inf $$< --out $$@ --name $(1)
 
-$(LOCAL_PATH)/host/xtest/regression_8100.c: $(TARGET_OUT_HEADERS)/$(1).h
+LOCAL_ADDITIONAL_DEPENDENCIES += $(TARGET_OUT_HEADERS)/$(1).h
+
 endef
 
 $(eval $(call my-embed-file,regression_8100_ca_crt,cert/ca.crt))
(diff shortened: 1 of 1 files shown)


liuyq:
@vchong @jforissier please help to check if we could have this change merged
Thanks,
Yongqin Liu

jforissier:
I don't see how this change relates to the usage of `.PHONY`, so the commit description could be a bit more detailed I suppose. The change looks good anyway.

liuyq:
> I don't see how this change relates to the usage of `.PHONY`, so the commit description could be a bit more detailed I suppose. The change looks good anyway.

we need to change to avoid following compiling error:

[100% 860/860] writing build rules ...
FAILED: 
external/optee_test/Android.mk:71: error: writing to readonly directory: "external/optee_test/host/xtest/regression_8100.c"
17:17:36 ckati failed with: exit status 1

liuyq:
@jforissier I updated the commit message, please help to check there again, Thanks!

jforissier:
@liuyq thanks, now I understand the problem. I will reword the commit message when merging the patch, if you don't mind.

jforissier:
Merged manually.

PR #316, opened 2018-11-09
ta: cmake: header files rely on C support

etienne-lms:
Without specify the TA header files are in C source file scope
cmake may attempt to look for resources as g++. When building
with environments that do not provide such tools as when building
from native buildroot ofr a qemu target, optee_client fails to
build. This change ensure a minimal C support allows to build
optee_client with cmake.


jforissier:

PR #318, opened 2018-12-14
regression: split 4007 into key types

jenswi-linaro:
Splits regression case 4007 into
- 4007_symmetric for symmetric keys
- 4007_rsa for RSA keys, which also is complemented with 3072 and 4096 keys
- 4007_dh for Diffie-Hellman keys
- 4007_ecc for ECC keys


jenswi-linaro:
This helps when testing for SWG-347

etienne-lms:

jenswi-linaro:
Tag applied

PR #319, opened 2018-12-19
regression: 6010: add rename conflict case (with close)

jforissier:
Adds a test case for issue https://github.com/OP-TEE/optee_os/issues/2707.


jenswi-linaro:

Issue #296, opened 2018-06-22
GCC 8.1.0 Errors in adbg_run.c and regression_4000.c (see patches attached)


simonqhughes:
The following errors are generated by optee_test being built with openembedded-core (tip of masters)

```
 /mnt/data/default_20180621_083657/build-xxx/tmp-xxx-glibc/work/imx7s_warp_xxx-oe-linux-gnueabi/optee-test/2.6.0+gitAUTOINC+45218eb59b-r0/git/host/xtest/adbg/src/adbg_run.c: In function 'Do_ADBG_AppendToSuite':
| /mnt/data/default_20180621_083657/build-xxx/tmp-xxx-glibc/work/imx7s_warp_xxx-oe-linux-gnueabi/optee-test/2.6.0+gitAUTOINC+45218eb59b-r0/git/host/xtest/adbg/src/adbg_run.c:103:3: error: 'strncpy' specified bound depends on the length of the source argument [-Werror=stringop-overflow=]
|    strncpy(p, Source_p->SuiteID_p, size);
|    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
| /mnt/data/default_20180621_083657/build-xxx/tmp-xxx-glibc/work/imx7s_warp_xxx-oe-linux-gnueabi/optee-test/2.6.0+gitAUTOINC+45218eb59b-r0/git/host/xtest/adbg/src/adbg_run.c:88:9: note: length computed here
|   size = strlen(Source_p->SuiteID_p);
|          ^~~~~~~~~~~~~~~~~~~~~~~~~~~
| cc1: all warnings being treated as errors
| Makefile:165: recipe for target '/mnt/data/default_20180621_083657/build-xxx/tmp-xxx-glibc/work/imx7s_warp_xxx-oe-linux-gnueabi/optee-test/2.6.0+gitAUTOINC+45218eb59b-r0/git/out/xtest/adbg/src/adbg_run.o' failed
| make[1]: *** [/mnt/data/default_20180621_083657/build-xxx/tmp-xxx-glibc/work/imx7s_warp_xxx-oe-linux-gnueabi/optee-test/2.6.0+gitAUTOINC+45218eb59b-r0/git/out/xtest/adbg/src/adbg_run.o] Error 1
| make[1]: *** Waiting for unfinished jobs....
[...]

PR #320, opened 2018-12-19
xtest: prevent potential overflow

etienne-lms:
**(edited)**

Fix issues reported by GCC 8.1.0 with  `-Werror=stringop-overflow=.` **as reported per https://github.com/OP-TEE/optee_test/issues/296**.
**Issues also reported with a GCC 8.2.0 from a standard build command, i.e:
`make TA_DEV_KIT_DIR=path/to/devkit OPTEE_CLIENT_EXPORT=path/to/client-export`**

  host/xtest/adbg/src/adbg_run.c:103:3: error: 'strncpy' specified bound depends on the length of the source argument [-Werror=stringop-overflow=]

  host/xtest/regression_4000.c:4986:3: error: 'memmove' pointer overflow between offset [0, 8] and size [4294967295, 2147483647] accessing array 'tmp' with type 'uint8_t[1024]' {aka 'unsigned char[1024]'} [-Werror=array-bounds]

Fixes issue https://github.com/OP-TEE/optee_test/issues/296

~~This change also adds test on buffer size output value to prevent buffer overflows in few other specific cases.~~

Signed-off-by: Simon Hughes <simon.hughes@arm.com>

### DIFF
--- a/host/xtest/adbg/src/adbg_run.c
+++ b/host/xtest/adbg/src/adbg_run.c
@@ -100,7 +100,7 @@ int Do_ADBG_AppendToSuite(
 		snprintf(p, size, "%s+%s", Dest_p->SuiteID_p,
 			 Source_p->SuiteID_p);
 	else
-		strncpy(p, Source_p->SuiteID_p, size);
+		snprintf(p, size, "%s", Source_p->SuiteID_p);
 	free((void *)Dest_p->SuiteID_p);
 	Dest_p->SuiteID_p = p;
 
--- a/host/xtest/regression_4000.c
+++ b/host/xtest/regression_4000.c
@@ -5018,18 +5018,28 @@ static void xtest_tee_test_4011(ADBG_Case_t *c)
 				out, out_size, tmp, &tmp_size)))
 			goto out;
 
+		if (!ADBG_EXPECT_COMPARE_UNSIGNED(c, tmp_size, <=, sizeof(tmp)))
+			goto out;
+
 		/* 4.1 */
-		for (n = 0; n < tmp_size; n++)
+		for (n = 0; n < tmp_size - i; n++)
 			if (tmp[n] == 0xff)
 				break;
+
+		/* Shall find at least a padding start before buffer end */
+	        if (!ADBG_EXPECT_COMPARE_UNSIGNED(c, n, <, tmp_size - i - 1))
+			goto out;
+
 		for (m = n + 1; m < tmp_size; m++)
 			if (tmp[m] != 0xff)
 				break;
+
 		/* 4.2 */
 		memmove(tmp + n + i, tmp + m, tmp_size - m);
+
 		/* 4.3 */
-		for (n = n + tmp_size - m + i; n < tmp_size; n++)
-			tmp[n] = 0;
+		n = n + i + tmp_size - m;
+		memset(tmp + n, 0, tmp_size - n);
 
 		/* 5 */
 		out_size = sizeof(out);


etienne-lms:
I have refined this P-R. Fixes the 2 and only 2 issues reported when built with a GCC 8.2.0.

host/xtest/adbg/src/adbg_run.c
```diff
@@ -100,7 +100,7 @@ int Do_ADBG_AppendToSuite(
 		snprintf(p, size, "%s+%s", Dest_p->SuiteID_p,
 			 Source_p->SuiteID_p);
 	else
-		strncpy(p, Source_p->SuiteID_p, size);
+		snprintf(p, size, "%s", Source_p->SuiteID_p);
```
jenswi-linaro: This seems to be fixing a false positive. The compiler doesn't seem to understand that `p` is also allocated based on `size.

etienne-lms: I agree. If it was suspicious, `snprintf()` should not match neither.
Interesting article on that from [readhat](https://developers.redhat.com/blog/2018/05/24/detecting-string-truncation-with-gcc-8/):

> _Using strncpy Safely_
In general, it is not possible to avoid string truncation by strncpy except by sizing the destination to be at least a byte larger than the length of the source string. With that approach, however, **using strncpy becomes unnecessary and the function can be avoided in favor of other APIs such as strcpy or (less preferably) memcpy.** Much has been written about the problems with strncpy and we recommend to avoid it whenever possible. It is, however, worth keeping in mind that unlike other standard string-handling functions, strncpy always writes exactly as many characters as specified by the third argument; if the source string is shorter, the function fills the remaining bytes with NULs.

From the bold part, we should use `strcpy` !!! :) the one prohibited!
I think `snprintf` is fine and consistent with alternate case. Let's use it.

jenswi-linaro: As far as I can tell was the code already correct so we're not really fixing anything. I think the commit message should be updated to reflect that.

etienne-lms: ok, will do.

host/xtest/regression_4000.c
```diff
@@ -5026,6 +5026,14 @@ static void xtest_tee_test_4011(ADBG_Case_t *c)
 			if (tmp[m] != 0xff)
 				break;
 		/* 4.2 */
+		if (!ADBG_EXPECT_COMPARE_UNSIGNED(c, n + i, < , sizeof(tmp)) ||
```
jenswi-linaro: I'd prefer if we fixed the code in a way a human (instead of a compiler) would understand.
That means checking right line 5019 that `tmp_size <= sizeof(tmp)`
The loop at line 5022 should only run to `tmp_size - i`
And so on.

etienne-lms: ok. I'll send a fix-up.

host/xtest/regression_4000.c
```diff
@@ -5018,26 +5018,28 @@ static void xtest_tee_test_4011(ADBG_Case_t *c)
 				out, out_size, tmp, &tmp_size)))
 			goto out;
 
+		if (tmp_size > sizeof(tmp))
```
jenswi-linaro: ADBG_EXPECT_COMPARE_UNSIGNED

etienne-lms: yes, thanks.

jenswi-linaro:
LGTM

etienne-lms:
squashed, tag applied.

etienne-lms:
Sorry: buggy header comment in 1st commit.

etienne-lms:
updated, ready to be merged.

PR #322, opened 2019-01-21
Build with GP test suite with CMake

etienne-lms:
Fix build issues when compiling GPD test sutie (v1.1.0.4)  with recent GCC.
Update CMake scripts to build GPD test suite resources.


jforissier:
Thanks for taking care of this :)

- Could we avoid adding yet another config flag (`CFG_GP_TEST_SUITE=y`) and rely on `CFG_GP_PACKAGE_PATH` being defined instead?
- When you say "previously patched using `CFG_GP_PACKAGE_PATH`", I suppose you refer to `make patch CFG_GP_PACKAGE_PATH=...`? (that is, using the regular Makefile instead of CMake). I'd rather have the patch rules included in the CMake files. Otherwise we can't deprecate the older Makefile. In any case I'd like to be able to run "export CFG_GP_PACKAGE_PATH=...; make` from the build directory (I mean build.git from our manifest forest), and have everything build properly. Like when GP is disabled.

etienne-lms:
> Could we avoid adding yet another config flag (`CFG_GP_TEST_SUITE=y`) and rely on `CFG_GP_PACKAGE_PATH` being defined instead?

Yes  but note cmake only builds, it does not apply patches hence does not need the path value.
I'll see if defining a path fits cmake scripts expectations.
Note we can introduce `CFG_GP_TEST_SUITE=y/n` without defining it in **CMakeLists.txt**. @lorc showed running `cmake -DCFG_GP_TEST_SUITE=y (...)` allows to use `CFG_GP_TEST_SUITE` in the cmake files.

> (...) I'd rather have the patch rules included in the CMake files (...) 

Looks like cmake is not really patch aware, unless using [ExternalProject](https://cmake.org/cmake/help/v3.12/module/ExternalProject.html?highlight=apply%20patch) stuff.

> (...) I'd like to be able to run `export CFG_GP_PACKAGE_PATH=...; make` (...) 

I think cmake does not permit that. 
The directive must be defined from the cmake env, not the user env. Yet I don't know much about cmake...


> (...) Otherwise we can't deprecate the older Makefile (...)

Do we want to deprecate the regular gnu makefiles?

jbech-linaro:
> Do we want to deprecate the regular gnu makefiles?

I'm cannot make up my mind regarding that.
**Pros**: We would have just a single set of makefile for plain Linux builds.
**Cons**: Without our old GNU makefiles you **cannot** rebuild just a single git (i.e., `make optee-examples` and friends doesn't work).

For example, quite often I'm working like this:
1. Build QEMU with `QEMU_VIRTFS_ENABLE=y QEMU_USERNET_ENABLE=y`
2. Boot up as usual (and keep it running)
3. Mount a folder to my PC and symlink to the TA I'm working with
4. Edit, compile the TA.
5. Run the TA directly without rebooting.

Step 4 and 5 can be repeated many time and is done in matter of seconds, which means that there is basically no turn-around time to test changes you are doing with a TA. Without our old makefiles I cannot do this. That is the major drawback for me if we get rid of this. We could probably come up with something similar for CMake, but we don't have that right now.

(FYI, for `optee_examples` I using [this](https://github.com/jbech-linaro/optee_examples/blob/build_using_shell/build.sh), which basically wraps 3,4 and 5 in my examples).

etienne-lms:
We also need gnu makefiles for the secure code (core, libs, TAs) as cmake can't handle them.
I think it is nice to support several build ways (make cmake android) as long as the maintenance is not painful.

etienne-lms:
Rebased (conflicts resolved).

Note that with the current proposal  GP tests patches must be applied before running `cmake` after which  one can prepare test build with GP test embedded using:
```
  sh> CFG_GP_PACKAGE_PATH=path/to/package cmake -DOPTEE_TEST_SDK=path/to/sdk
or
  sh> export CFG_GP_PACKAGE_PATH=path/to/package
  sh> cmake -DOPTEE_TEST_SDK=path/to/sdk
```
...followed by...
```
  sh> make
```

lorc:
> We also need gnu makefiles for the secure code (core, libs, TAs) as cmake can't handle them.

I think, cmake can handle it. You just need custom rules or/and custom toolchain configuration.  Cmake can build baremetal firmware (https://github.com/lorc/stm32f1, yes it is obsolete a bit:) ), so I can't see why it can't built TA or even OP-TEE core.

jforissier:
@etienne-lms I'd like to have the commands in the commit message, showing how to use this with our build environment (manifest.git, build.git), I mean QEMU for instance.

etienne-lms:
Ok, i'll squash the commits.

jenswi-linaro:
This seems useful. Can we try to see this through?

PR #323, opened 2019-01-22
Fix possibly uninitialized variable in regression tests

etienne-lms:
Building optee_test with `cmake -DCMAKE_BUILD_TYPE=Release (...)` show errors reported by my GCC (8.2.0).

This P-R fixes the reported errors. The strange thing is that there are equivalent occurrences not reported by GCC, as the `obj{0|1|2}` in **[regression_6000.c](https://github.com/OP-TEE/optee_test/blob/master/host/xtest/regression_6000.c#L1145)**. This P-R fixes only the one reported by GCC.

### DIFF
--- a/host/xtest/regression_4000.c
+++ b/host/xtest/regression_4000.c
@@ -3655,7 +3655,7 @@ static void xtest_tee_test_4006(ADBG_Case_t *c)
 	uint8_t out_enc[512];
 	size_t out_enc_size;
 	uint8_t ptx_hash[TEE_MAX_HASH_SIZE];
-	size_t ptx_hash_size;
+	size_t ptx_hash_size = 0;
 	size_t max_key_size;
 	size_t num_key_attrs;
 	uint32_t ret_orig;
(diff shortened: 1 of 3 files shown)


jenswi-linaro:

host/xtest/regression_4000.c
```diff
@@ -3655,7 +3655,7 @@ static void xtest_tee_test_4006(ADBG_Case_t *c)
 	uint8_t out_enc[512];
 	size_t out_enc_size;
 	uint8_t ptx_hash[TEE_MAX_HASH_SIZE];
-	size_t ptx_hash_size;
+	size_t ptx_hash_size = 0;
 	size_t max_key_size;
```
jbech-linaro: How come it complains at the one above, but not this one?

jbech-linaro:

host/xtest/regression_4000.c
etienne-lms: I don't know how far compiler can check instruction blocks conditions matches or not when variable init and use are split as here (`if () ...; switch() { ... }  ...; switch() { ... } `).

lorc:
`Reviewed-by: Volodymyr Babchuk <vlad.babchuk@gmail.com>`

Had the same issue, actually. Worked around it with `-Wno-maybe-uninitialized`. 
But it is nice to have proper fix. Thanks.

PR #325, opened 2019-01-24
ta/cryp_taf: correct return value

sticfeynman:
ta_entry_copy_object_attributes() should faithfully returns
TEE_CopyObjectAttributes1()'s value instead of always overwriting it with
TEE_SUCCESS.

Signed-off-by: Oliver Chiang <rockerfeynman@gmail.com>


sticfeynman:
Tag fixed. Thanks!

jforissier:
@sticfeynman thanks!

PR #326, opened 2019-01-29
regression 4100: update string conversion loop

etienne-lms:
Change the loop used to convert string into numerical value.
The original loop was fine but its implementation hits toolchain
unsafe-loop-optimizations feature. The new implementation
proposed here simplifies a bit the loop and prevents toolchain
from complaining when directive -Werror=unsafe-loop-optimizations
is enabled.

Issue reported by the Buildroot cross toolchain [1] with the
following error traces:

build/armv7/build/optee-test-3.4.0/host/xtest/regression_4100.c:447:8: error: missed loop optimization, the loop counter may overflow [-Werror=unsafe-loop-optimizations]
  while (spos) {
        ^
build/optee-test-3.4.0/host/xtest/regression_4100.c:454:6: error: missed loop optimization, the loop counter may overflow [-Werror=unsafe-loop-optimizations]
   if (!spos)
      ^

[1] arm-buildroot-linux-uclibcgnueabihf-gcc.br_real (Buildroot 2019.02-git-00933-gb75e93c) 7.4.0

### DIFF
--- a/host/xtest/regression_4100.c
+++ b/host/xtest/regression_4100.c
@@ -446,15 +446,18 @@ static TEEC_Result convert_from_string(ADBG_Case_t *c, TEEC_Session *s,
 
 	while (spos) {
 		nibble = digit_value(str[spos - 1]);
-		if (nibble == -1)
+		if (nibble == -1) {
+			spos--;
 			break;
+		}
 		os[ospos] = nibble;
 
 		if (spos > 1) {
 			nibble = digit_value(str[spos - 2]);
-			if (nibble == -1)
+			if (nibble == -1) {
+				spos -= 2;
 				break;
-
+			}
 			os[ospos] |= nibble << 4;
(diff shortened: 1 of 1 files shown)


jbech-linaro:
IBART seems unhappy with the changes. Is there any need to update our toolchain to something even more recent (I'm already working with that, so that's why I'm curious).

etienne-lms:
We should not need to update the toolchain. There must be something wring in my change.

etienne-lms:
@jbech-linaro, an alternative to this hacky workaround would be to remove toolchain directive `-Wunsafe-loop-optimizations` that is currently set from top [CMakeLists.txt](https://github.com/OP-TEE/optee_test/blob/3.4.0/CMakeLists.txt#L20) (as well as GNU and Android makefiles).

etienne-lms:
Superseded by #344

PR #327, opened 2019-01-30
regression: 4002: run one CBC MAC test in multiple increments

jforissier:
The .in_incr field of struct xtest_mac_case is currently used for
a single step, that is, update() is called to process in_incr byte,
then the rest of the input buffer is passed to final().

This commit adds a macro which allows to run the CBC MAC update()
several times in chunks of in_incr bytes, then the remaining bytes
in the input buffer (<= in_incr) are passed to final(). One entry is
added to the mac_cases[] array.

### DIFF
--- a/host/xtest/regression_4000.c
+++ b/host/xtest/regression_4000.c
@@ -1497,11 +1497,20 @@ struct xtest_mac_case {
 	size_t in_len;
 	const uint8_t *out;
 	size_t out_len;
+	bool multiple_incr;
 };
 
-#define XTEST_MAC_CASE(algo, key_type, key, in_incr, in, out) \
+#define XTEST_MAC_CASE_1(algo, key_type, key, in_incr, in, out) \
 	{ (algo), (key_type), (key), ARRAY_SIZE(key), \
-	  (in_incr), (in), ARRAY_SIZE(in), (out), ARRAY_SIZE(out) }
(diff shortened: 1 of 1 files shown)


jforissier:
IBART is failing as expected (the new test case depends on fix #2779).

jenswi-linaro:

sticfeynman:
Can you generalize this idea to every test case which has .in_incr in regression_4000?

Wouldn't this multi-update test cover the original single-update version? If so, do we still need the MULT macro?
And I think you need to consider the `.in_incr == 0` in the while condition, I run into this infinite loop trouble in 4005.

jforissier:
> Can you generalize this idea to every test case which has .in_incr in regression_4000?

Yeah that could be done. Let's keep this patch as is, since it tests the very specific bug that you have discovered and fixed. I'll see what I can do and submit a follow-up later.

> Wouldn't this multi-update test cover the original single-update version? If so, do we still need the MULT macro?

I prefer to keep the original tests. The multiple update version will call final() only with small values (<= in_incr). What if we have a bug with final() on a larger buffer?

> And I think you need to consider the `.in_incr == 0` in the while condition, I run into this infinite loop trouble in 4005.

`.in_incr` == 0 is not a valid value

jforissier:
Update. All things considered, I wouldn't mind squashing the two commits. @jenswi-linaro what's your opinion?

jenswi-linaro:
@jforissier, fine with me.

PR #329, opened 2019-02-07
Musl fixes

andred:
These patches fix compilation against the musl C library.

### DIFF
--- a/ta/crypt/arith_taf.c
+++ b/ta/crypt/arith_taf.c
@@ -2,6 +2,7 @@
 /* Copyright (c) 2018, Linaro Limited */
 
 #include <arith_taf.h>
+#include <inttypes.h>
 #include <ta_crypt.h>
 #include <tee_internal_api.h>
 #include <util.h>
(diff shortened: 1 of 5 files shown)


jforissier:
Hi @andred,

Thanks for the patches, look fine to me.

ta/crypt/arith_taf.c
```diff
@@ -2,6 +2,7 @@
 /* Copyright (c) 2018, Linaro Limited */
 
 #include <arith_taf.h>
+#include <inttypes.h>
```
jenswi-linaro: I don't expect that this file is compiled with musl-libc.
I agree with the change, it's just that the commit message isn't accurate for this file.

andred: You're right. I've updated the subjects of all related patches

jenswi-linaro:

jforissier:
@andred would you please add @jenswi-linaro's `Acked-by:` tag to each commit? Then I can merge this PR.

jforissier:
Thanks!

PR #332, opened 2019-02-28
regression 1008: adjust TA corruption test for CFG_ULIBS_SHARED=y

jforissier:
When TAs are linked against shared versions of libutee etc., the
size of the main binary is reduced significantly. This causes test
1008.4 to fail, because the size of the TA drops from ~121 KiB down
to ~9 KiB which is smaller than the offset supplied to
load_corrupt_ta(). Use a smaller offest to fix the issue.


jforissier:
Related to https://github.com/OP-TEE/optee_os/pull/2849

jenswi-linaro:

PR #333, opened 2019-03-04
Allow memrefs with size 0

cneveux:
Add test 1025 verifying temporary memref of 0 bytes or NULL pointer
      buffer conforming with:
      GlobalPlatform Client API v1.0c (section 3.2.5 memory references), which
      can be found at [1].
    
      Link: [1] http://www.globalplatform.org/specificationsdevice.asp

Signed-off-by: Cedric Neveux <cedric.neveux@nxp.com>

### DIFF
--- a/host/xtest/regression_1000.c
+++ b/host/xtest/regression_1000.c
@@ -1909,3 +1909,77 @@ static void xtest_tee_test_1024(ADBG_Case_t *c)
 ADBG_CASE_DEFINE(regression, 1024, xtest_tee_test_1024,
 		 "Test PTA_SYSTEM_GET_TPM_EVENT_LOG Service");
 #endif /* CFG_CORE_TPM_EVENT_LOG */
+
+static void xtest_tee_test_1025(ADBG_Case_t *c)
+{
+	TEEC_Session session = {};
+	TEEC_Operation op = TEEC_OPERATION_INITIALIZER;
+	uint32_t ret_orig = 0;
+	uint8_t *empty_buf = NULL;
+
+	if (!ADBG_EXPECT_TEEC_SUCCESS(c,
+				      xtest_teec_open_session(&session,
+							      &os_test_ta_uuid,
+							      NULL, &ret_orig)))
+		return;
+
+	empty_buf = malloc(1);
+	if (!empty_buf) {
+		(void)ADBG_EXPECT_TEEC_SUCCESS(c, TEEC_ERROR_OUT_OF_MEMORY);
+		goto out;
+	}
+
+	op.paramTypes = TEEC_PARAM_TYPES(TEEC_MEMREF_TEMP_INPUT,
+					 TEEC_MEMREF_TEMP_INPUT,
+					 TEEC_MEMREF_TEMP_OUTPUT,
+					 TEEC_MEMREF_TEMP_OUTPUT);
+
+	Do_ADBG_BeginSubCase(c,
+			     "Input/Output MEMREF Buffer NULL - Size 0 bytes");
+
+	op.params[0].tmpref.buffer = empty_buf;
+	op.params[0].tmpref.size = 0;
+
+	op.params[1].tmpref.buffer = NULL;
+	op.params[1].tmpref.size = 0;
+
+	op.params[2].tmpref.buffer = empty_buf;
+	op.params[2].tmpref.size = 0;
+
+	op.params[3].tmpref.buffer = NULL;
+	op.params[3].tmpref.size = 0;
+
+	ADBG_EXPECT(c, TEE_SUCCESS,
+		    TEEC_InvokeCommand(&session,
+				       TA_OS_TEST_CMD_NULL_MEMREF_PARAMS, &op,
+				       &ret_orig));
+
+	Do_ADBG_EndSubCase(c, "Input/Output MEMREF Buffer NULL - Size 0 bytes");
+
+	Do_ADBG_BeginSubCase(c, "Input MEMREF Buffer NULL - Size non 0 bytes");
+
+	op.params[0].tmpref.buffer = empty_buf;
+	op.params[0].tmpref.size = 1;
+
+	op.params[1].tmpref.buffer = NULL;
+	op.params[1].tmpref.size = 0;
+
+	op.params[2].tmpref.buffer = empty_buf;
+	op.params[2].tmpref.size = 0;
+
+	op.params[3].tmpref.buffer = NULL;
+	op.params[3].tmpref.size = 0;
+
+	ADBG_EXPECT(c, TEE_ERROR_BAD_PARAMETERS,
+		    TEEC_InvokeCommand(&session,
+				       TA_OS_TEST_CMD_NULL_MEMREF_PARAMS, &op,
+				       &ret_orig));
+
+	Do_ADBG_EndSubCase(c, "Input MEMREF Buffer NULL - Size non 0 bytes");
+
+out:
+	TEEC_CloseSession(&session);
+	free(empty_buf);
+}
+ADBG_CASE_DEFINE(regression, 1025, xtest_tee_test_1025,
+		 "Test memref NULL and/or 0 bytes size");
--- a/ta/os_test/include/ta_os_test.h
+++ b/ta/os_test/include/ta_os_test.h
@@ -49,5 +49,6 @@
 #define TA_OS_TEST_CMD_CALL_LIB_DL          16
 #define TA_OS_TEST_CMD_CALL_LIB_DL_PANIC    17
 #define TA_OS_TEST_CMD_GET_GLOBAL_VAR       18
+#define TA_OS_TEST_CMD_NULL_MEMREF_PARAMS   19
 
 #endif /*TA_OS_TEST_H */
--- a/ta/os_test/os_test.c
+++ b/ta/os_test/os_test.c
@@ -1068,6 +1068,30 @@ TEE_Result ta_entry_params(uint32_t param_types, TEE_Param params[4])
 	return TEE_SUCCESS;
 }
 
+TEE_Result ta_entry_null_memref(uint32_t param_types, TEE_Param params[4])
+{
+	if (param_types != TEE_PARAM_TYPES(TEE_PARAM_TYPE_MEMREF_INPUT,
+					   TEE_PARAM_TYPE_MEMREF_INPUT,
+					   TEE_PARAM_TYPE_MEMREF_OUTPUT,
+					   TEE_PARAM_TYPE_MEMREF_OUTPUT))
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	/*
+	 * Tests how client can provide null or non-null memref parameters
+	 * param[0] expected as a 0 byte input mapped memeref.
+	 * param[1] expected as a 0 byte input not-mapped memeref.
+	 * param[2] expected as a 0 byte output mapped memeref.
+	 * param[3] expected as a 0 byte output not-mapped memeref.
+	 */
+	if (!params[0].memref.buffer || params[0].memref.size ||
+	    params[1].memref.buffer || params[1].memref.size ||
+	    !params[2].memref.buffer || params[2].memref.size ||
+	    params[3].memref.buffer || params[3].memref.size)
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	return TEE_SUCCESS;
+}
+
 TEE_Result ta_entry_call_lib(uint32_t param_types,
 			     TEE_Param params[4] __unused)
 {
--- a/ta/os_test/ta_entry.c
+++ b/ta/os_test/ta_entry.c
@@ -109,6 +109,9 @@ TEE_Result TA_InvokeCommandEntryPoint(void *pSessionContext,
 	case TA_OS_TEST_CMD_PARAMS:
 		return ta_entry_params(nParamTypes, pParams);
 
+	case TA_OS_TEST_CMD_NULL_MEMREF_PARAMS:
+		return ta_entry_null_memref(nParamTypes, pParams);
+
 	case TA_OS_TEST_CMD_CALL_LIB:
 		return ta_entry_call_lib(nParamTypes, pParams);
 
--- a/ta/os_test/include/os_test.h
+++ b/ta/os_test/include/os_test.h
@@ -42,6 +42,7 @@ TEE_Result ta_entry_ta2ta_memref(uint32_t param_types, TEE_Param params[4]);
 TEE_Result ta_entry_ta2ta_memref_mix(uint32_t param_types,
 				     TEE_Param params[4]);
 TEE_Result ta_entry_params(uint32_t param_types, TEE_Param params[4]);
+TEE_Result ta_entry_null_memref(uint32_t param_types, TEE_Param params[4]);
 TEE_Result ta_entry_call_lib(uint32_t param_types, TEE_Param params[4]);
 TEE_Result ta_entry_call_lib_panic(uint32_t param_types, TEE_Param params[4]);
 TEE_Result ta_entry_call_lib_dl(uint32_t param_types, TEE_Param params[4]);


jenswi-linaro:
This doesn't compile

cneveux:
Build is fixed. Now there is stage test failure but I don't have access to the details so I can't see what is the issue here.

jenswi-linaro:
The error is expected since it's testing against baseline:
```
  1153:  * regression_1021 Test memref NULL and/or 0 bytes size
  1154:  regression_1000.c:1518: TEEC_InvokeCommand(&session, 14, &op, &ret_orig) has an unexpected value: 0xffff0006, expected 0x0
  1155:    regression_1021 FAILED
  1156:   
```

host/xtest/regression_1000.c
```diff
@@ -1473,3 +1473,63 @@ static void xtest_tee_test_1020(ADBG_Case_t *c)
...
+	TEEC_Session session = { 0 };
+	TEEC_Operation op = TEEC_OPERATION_INITIALIZER;
+	uint32_t ret_orig;
+	uint8_t *empty_buf = NULL;
+
+	/*
+	 * First check if the test is relevant because this feature
+	 * may be not supported
+	 */
+	if (!xtest_teec_ctx.memref_null) {
```
jenswi-linaro: I'm not sure this is needed. All gits are supposed to be on latest when testing.

jforissier: @jenswi-linaro correct

cneveux: Sorry, I'm bit lost. I understood that we are supposed to support old drivers. Hence with a driver that doesn't enable this feature this test should not be run.

jenswi-linaro: No worries. From test point of view we're always assuming that the gits are updated and correctly configured.

The compatibility stuff is for third party code over which we have no control. It may be hard for some projects to do synchronized updates of everything (optee_client, optee_os and the kernel).

cneveux: ok removed

jbech-linaro:
And ... re-open this one too (due to [OP-TEE/optee_client#145](https://github.com/OP-TEE/optee_client/pull/145) just being merged).

host/xtest/regression_1000.c
```diff
@@ -1715,3 +1715,82 @@ static void xtest_tee_test_1021(ADBG_Case_t *c)
 }
 ADBG_CASE_DEFINE(regression, 1021, xtest_tee_test_1021,
 		 "Test panic context release");
+
+static void xtest_tee_test_1022(ADBG_Case_t *c)
+{
+	TEEC_Session session = { 0 };
+	TEEC_Operation op = TEEC_OPERATION_INITIALIZER;
+	uint32_t ret_orig;
```
etienne-lms: `= 0`

host/xtest/regression_1000.c
```diff
@@ -1715,3 +1715,82 @@ static void xtest_tee_test_1021(ADBG_Case_t *c)
...
+	op.params[1].tmpref.size = 0;
+
+	op.params[2].tmpref.buffer = empty_buf;
+	op.params[2].tmpref.size = 0;
+
+	op.params[3].tmpref.buffer = NULL;
+	op.params[3].tmpref.size = 0;
+
+	ADBG_EXPECT(c, TEE_SUCCESS,
+			TEEC_InvokeCommand(&session,
```
etienne-lms: indent to `(`

host/xtest/regression_1000.c
```diff
@@ -1715,3 +1715,82 @@ static void xtest_tee_test_1021(ADBG_Case_t *c)
...
+				TA_OS_TEST_CMD_PARAMS_TMPMEM, &op, &ret_orig));
+
+	Do_ADBG_EndSubCase(c, NULL);
```
etienne-lms: prefer "Input/Output MEMREF Buffer NULL - Size 0 bytes" over NULL for closing.
Here and below.

host/xtest/regression_1000.c
```diff
@@ -1715,3 +1715,82 @@ static void xtest_tee_test_1021(ADBG_Case_t *c)
...
+
+	TEEC_CloseSession(&session);
+out:
+	free(empty_buf);
+}
+ADBG_CASE_DEFINE(regression, 1022, xtest_tee_test_1022,
+		"Test memref NULL and/or 0 bytes size");
```
etienne-lms: indent to `(`.

ta/os_test/os_test.c
```diff
@@ -1058,6 +1058,29 @@ TEE_Result ta_entry_params(uint32_t param_types, TEE_Param params[4])
...
 
+TEE_Result ta_entry_params_tmpmem(uint32_t param_types, TEE_Param params[4])
+{
+	if (param_types != TEE_PARAM_TYPES(TEE_PARAM_TYPE_MEMREF_INPUT,
+					   TEE_PARAM_TYPE_MEMREF_INPUT,
+					   TEE_PARAM_TYPE_MEMREF_OUTPUT,
+					   TEE_PARAM_TYPE_MEMREF_OUTPUT))
+		return TEE_ERROR_BAD_PARAMETERS;
+
+	if ((!params[0].memref.buffer) || (params[0].memref.size))
```
etienne-lms: remove inner parentheses.
could:
```
	if (!params[0].memref.buffer || params[0].memref.size ||
	    !params[1].memref.buffer || params[1].memref.size ||
	    !params[2].memref.buffer || params[2].memref.size ||
	    !params[3].memref.buffer || params[3].memref.size)
		return TEE_ERROR_BAD_PARAMETERS;

File: host/xtest/regression_1000.c [hunk 7]
```diff
@@ -1715,3 +1715,82 @@ static void xtest_tee_test_1021(ADBG_Case_t *c)
...
+	op.params[3].tmpref.buffer = NULL;
+	op.params[3].tmpref.size = 0;
+
+	ADBG_EXPECT(c, TEE_SUCCESS,
+			TEEC_InvokeCommand(&session,
+				TA_OS_TEST_CMD_PARAMS_TMPMEM, &op, &ret_orig));
+
+	Do_ADBG_EndSubCase(c, NULL);
+
+	Do_ADBG_BeginSubCase(c, "Input MEMREF Buffer NULL - Size non 0 bytes");
```
etienne-lms: suggest move empty line above to below this line, before comment/

host/xtest/regression_1000.c
```diff
@@ -1715,3 +1715,82 @@ static void xtest_tee_test_1021(ADBG_Case_t *c)
...
+{
+	TEEC_Session session = { 0 };
+	TEEC_Operation op = TEEC_OPERATION_INITIALIZER;
+	uint32_t ret_orig;
+	uint8_t *empty_buf = NULL;
+
+	if (!ADBG_EXPECT_TEEC_SUCCESS(c,
+		xtest_teec_open_session(&session, &os_test_ta_uuid, NULL,
+			&ret_orig)))
+		goto out;
```
etienne-lms: can return

host/xtest/regression_1000.c
```diff
@@ -1715,3 +1715,82 @@ static void xtest_tee_test_1021(ADBG_Case_t *c)
...
+	op.params[3].tmpref.buffer = NULL;
+	op.params[3].tmpref.size = 0;
+
+	ADBG_EXPECT(c, TEE_ERROR_BAD_PARAMETERS,
+			TEEC_InvokeCommand(&session,
+				TA_OS_TEST_CMD_PARAMS_TMPMEM, &op, &ret_orig));
+
+	Do_ADBG_EndSubCase(c, NULL);
+
+	TEEC_CloseSession(&session);
```
etienne-lms: clone session after out.

host/xtest/regression_1000.c
cneveux: Test and description are correct. The "Input/Ouput MEMREF Buffer NULL - Size 0 bytes" is done before. Here is to test the correct behaviour when Input MEMREF buffer is NULL pointer but size non 0 bytes:

jbech-linaro: @cneveux , I think @etienne-lms  refers to just have the "group" confined, i.e., same message 
as in "`Do_ADBG_BeginSubCase`", correct @etienne-lms ?

etienne-lms: true.

```C
	/*
	 * Check temp buffer parameters with a valid address
	 * but 0 bytes length
	 */
	Do_ADBG_BeginSubCase(c,
			     "Input/Output MEMREF Buffer NULL - Size 0 bytes");

	(...)

	Do_ADBG_EndSubCase(c, "Input/Output MEMREF Buffer NULL - Size 0 bytes");


	/*
	 * Check temp buffer parameters with a valid address
	 * but non 0 bytes length
	 */
	Do_ADBG_BeginSubCase(c, "Input MEMREF Buffer NULL - Size non 0 bytes");

	(...)

	Do_ADBG_EndSubCase(c, "Input MEMREF Buffer NULL - Size non 0 bytes");
```

Note that inline comment are redundant with log message, could remove the inline comment.

host/xtest/regression_1000.c
```diff
@@ -1879,3 +1879,82 @@ static void xtest_tee_test_1023(ADBG_Case_t *c)
...
+
+	ADBG_EXPECT(c, TEE_SUCCESS,
+		    TEEC_InvokeCommand(&session, TA_OS_TEST_CMD_PARAMS_TMPMEM,
+				       &op, &ret_orig));
+
+	Do_ADBG_EndSubCase(c, NULL);
+
+	/*
+	 * Check temp buffer parameters with a valid address
+	 * but non 0 bytes length
```
etienne-lms: From this description, we would expect buffer = NULL at lines 1937 and 1943
and size != 0 at line 1941 and 1947.

Maybe passing a unique parameter would be easier to understand the test.
test param#0 = memref IN / ptr=NULL / sz=0 => expect TEEC_SUCCESS.
test param#0 = memref OUT / ptr=NULL / sz=0 => expect TEEC_SUCCESS.
test param#0 = memref IN / ptr=NULL / sz=1 => expect TEEC_ERROR_BAD_PARAMETERS.
test param#0 = memref OUT / ptr=NULL / sz=1 => expect TEEC_ERROR_BAD_PARAMETERS.

host/xtest/regression_1000.c
```diff
@@ -1879,3 +1879,82 @@ static void xtest_tee_test_1023(ADBG_Case_t *c)
...
+	}
+
+	op.paramTypes = TEEC_PARAM_TYPES(TEEC_MEMREF_TEMP_INPUT,
+					 TEEC_MEMREF_TEMP_INPUT,
+					 TEEC_MEMREF_TEMP_OUTPUT,
+					 TEEC_MEMREF_TEMP_OUTPUT);
+
+	/*
+	 * Check temp buffer parameters with a valid address
+	 * but 0 bytes length
```
etienne-lms: From this description, we would expect buffer = NULL at lines 1916 and 1919.

host/xtest/regression_1000.c
cneveux: ok

host/xtest/regression_1000.c
cneveux: The current test implementation is verifying expected behavior. I would not change it here.

jbech-linaro:
`Tested-by: Joakim Bech <joakim.bech@linaro.org>` (QEMU)

jbech-linaro:
IBART issues etc is due to the dependencies to https://github.com/linaro-swg/linux/pull/67 https://github.com/OP-TEE/optee_os/pull/2851 etc ... and vice versa.

host/xtest/regression_1000.c
```diff
@@ -1879,3 +1879,74 @@ static void xtest_tee_test_1023(ADBG_Case_t *c)
...
+{
+	TEEC_Session session = { 0 };
+	TEEC_Operation op = TEEC_OPERATION_INITIALIZER;
+	uint32_t ret_orig = 0;
+	uint8_t *empty_buf = NULL;
+
+	if (!ADBG_EXPECT_TEEC_SUCCESS(c,
+		xtest_teec_open_session(&session, &os_test_ta_uuid, NULL,
+			&ret_orig)))
+		return;
```
etienne-lms: nitpicking, please fix indentation so that it is not confusing. Proposing:
```C
				      xtest_teec_open_session(&session,
							      &os_test_ta_uuid,
							      NULL, &ret_orig)))
		return;

File: host/xtest/regression_1000.c [hunk 13]
```diff
@@ -1879,3 +1879,74 @@ static void xtest_tee_test_1023(ADBG_Case_t *c)
...
+
+	empty_buf = malloc(0);
+	if (!empty_buf) {
```
etienne-lms: From `malloc()` man page:
> If size is  0,  then malloc()  returns either NULL, or a unique pointer value that can later be successfully passed to free().

So i think here should better allocate a non-zero sized buffer to be sure to get a valid pointer.

host/xtest/regression_1000.c
```diff
@@ -1879,3 +1879,74 @@ static void xtest_tee_test_1023(ADBG_Case_t *c)
...
+	op.params[3].tmpref.buffer = NULL;
+	op.params[3].tmpref.size = 0;
+
+	ADBG_EXPECT(c, TEE_SUCCESS,
+		    TEEC_InvokeCommand(&session, TA_OS_TEST_CMD_PARAMS_TMPMEM,
+				       &op, &ret_orig));
+
+	Do_ADBG_EndSubCase(c, "Input/Output MEMREF Buffer NULL - Size 0 bytes");
+
+	Do_ADBG_BeginSubCase(c, "Input MEMREF Buffer NULL - Size non 0 bytes");
```
etienne-lms: Sorry, I don"t catch this last test.
The test sets param#0 as a valid buffer (address and size) and invokes test TA returns an error since it expects a valid address and a null size. What does this really test?  I think this test could be removed, only the one above is useful.

ta/os_test/include/ta_os_test.h
```diff
@@ -49,5 +49,6 @@
 #define TA_OS_TEST_CMD_CALL_LIB_DL          16
 #define TA_OS_TEST_CMD_CALL_LIB_DL_PANIC    17
 #define TA_OS_TEST_CMD_GET_GLOBAL_VAR       18
+#define TA_OS_TEST_CMD_PARAMS_TMPMEM        19
```
etienne-lms: Suggest to rename `TA_OS_TEST_CMD_NULL_MEMREF_PARAMS`.

ta/os_test/os_test.c
```diff
@@ -1068,6 +1068,23 @@ TEE_Result ta_entry_params(uint32_t param_types, TEE_Param params[4])
 	return TEE_SUCCESS;
 }
 
+TEE_Result ta_entry_params_tmpmem(uint32_t param_types, TEE_Param params[4])
```
etienne-lms: I think a inline comment should explicit what this TA command expects, i.e.:
```c
/*
 * Tests how client can provide null or non-null memref parameters
 * param[0] expected as a 0 byte input mapped memeref.
 * param[1] expected as a 0 byte input not-mapped memeref.
 * param[2] expected as a 0 byte output mapped memeref.
 * param[3] expected as a 0 byte output not-mapped memeref.
```

host/xtest/regression_1000.c
cneveux: ok

host/xtest/regression_1000.c
cneveux: ok

host/xtest/regression_1000.c
cneveux: no sorry, you confuse me. This is the goal to have a buffer of 0 bytes allocated

host/xtest/regression_1000.c
cneveux: The test is to verify the buffer is 0 bytes but param size say that buffer is not 0 bytes. Hence a BAD_PARAMETERS error must be returned

ta/os_test/include/ta_os_test.h
cneveux: ok

ta/os_test/os_test.c
cneveux: ok

host/xtest/regression_1000.c
etienne-lms: I don't agree.

This test does not verify that the buffer provided `empty_buf` here relates to 0 bytes or not.
It just check that value set in `tmpref.size` is 0 or not, whatever effective allocated size is used when `empty_buffer` is allocated at line 1895 above.

Note that the effective size of `empty_buffer` is not known by TEE driver or TEE Core (or the TA).
For example, if you allocate 10 byte, and set `.size` to 11 bytes, it is likely that these are already mapped in CA, hence TEE driver/Core will successfully handle the 11 bytes buffer, but that the one that will write to the last byte will corrupt something in CA memory.

host/xtest/regression_1000.c
etienne-lms: I don't agree. The goal here is to have a valid pointer (that is meaningful in the CA and that TEE driver can convert into a valid physical memory address) and to invoke a TA providing this pointer together with a size info set to 0.

So here, whatever memory is allocated, we don't care. We just care having a non-NULL pointer.
But `malloc()` man page says that `malloc()` may return NULL, depending of the libc, OS, ... Thus to ensure this test is portable, you must make sure `empty_buf` is not NULL, hence either malloc some byte(s), or use some valid reference, i.e.

```c
	char dummy_buf;

	empty_buf = &dummy_buf;
```

host/xtest/regression_1000.c
cneveux: ok

host/xtest/regression_1000.c
```diff
@@ -1879,3 +1879,77 @@ static void xtest_tee_test_1023(ADBG_Case_t *c)
...
+		    TEEC_InvokeCommand(&session,
+				       TA_OS_TEST_CMD_NULL_MEMREF_PARAMS, &op,
+				       &ret_orig));
+
+	Do_ADBG_EndSubCase(c, "Input/Output MEMREF Buffer NULL - Size 0 bytes");
+
+	Do_ADBG_BeginSubCase(c, "Input MEMREF Buffer NULL - Size non 0 bytes");
+
+	op.params[0].tmpref.buffer = empty_buf;
+	op.params[0].tmpref.size = 10;
```
etienne-lms: Sorry complaining again :(

I think this is a bit dangerous. If you state here size is 10 bytes whereas `empty_buf` effective size is 1 byte, then TEE client library may try to read 10 bytes from `empty_buf` (as this param is a `MEMREF_TEMP_INPUT`). This may (or may not) lead to a segment fault which is not what we expect from xtest.

I still think this 2nd part is useless. However if you really want to get it in this test, at least make sure `.size` here does not overflow `empty_buf` effective size. You could use `.size = 1` for example.

cneveux: done

ta/os_test/ta_entry.c
```diff
@@ -109,6 +109,9 @@ TEE_Result TA_InvokeCommandEntryPoint(void *pSessionContext,
 	case TA_OS_TEST_CMD_PARAMS:
 		return ta_entry_params(nParamTypes, pParams);
 
+	case TA_OS_TEST_CMD_NULL_MEMREF_PARAMS:
+		return ta_entry_params_tmpmem(nParamTypes, pParams);
```
etienne-lms: for consistency, prefer rename `ta_entry_params_tmpmem()` into `ta_entry_null_memref()`.

etienne-lms:
LGTM. Please squash, rebase and apply tags (if not too tricky rebased conflict, and if @jbech-linaro confirms his).

jbech-linaro:
> and if @jbech-linaro confirms his

I'm all OK with this!

jbech-linaro:
> LGTM. Please squash, rebase

Rebase definitely needed, the test case (1024) clashes with the newly added TPM test case. So, a "1025" is probably what we need.

jbech-linaro:
With this
- https://github.com/OP-TEE/optee_test/pull/333 rebased (and xtest 1024/1025 clash fixed)
- Latest from today https://github.com/linaro-swg/linux/pull/67
- Latest from today https://github.com/OP-TEE/optee_os/pull/2851

xtest passes (including 1025 of course) and my Widevine test using this passes.
`Tested-by: Joakim Bech <joakim.bech@linaro.org>` (QEMU)

jbech-linaro:
Hold your horses, when wiping my out-folder and recompiling I got a build error.

```bash
os_test.c:1071:12: warning: no previous prototype for ‘ta_entry_null_memref’ [-Wmissing-prototypes]                                                                                                                                                             
 TEE_Result ta_entry_null_memref(uint32_t param_types, TEE_Param params[4])                                                                                                                                                                                                     
            ^~~~~~~~~~~~~~~~~~~~                                                                                                                                                                                                                                                
  CC      out/ta_entry.o                                                                                                                                                                                                                                                        
ta_entry.c: In function ‘TA_InvokeCommandEntryPoint’:                                                                                                                                                                                                                           
ta_entry.c:113:10: error: implicit declaration of function ‘ta_entry_null_memref’; did you mean ‘ta_entry_ta2ta_memref’? [-Werror=implicit-function-declaration]                                                                                                    
   return ta_entry_null_memref(nParamTypes, pParams);                                                                                                                                                                                                                           
          ^~~~~~~~~~~~~~~~~~~~                                                                                                                                                                                                                                                  
          ta_entry_ta2ta_memref                                                                                                                                                                                                                                                 
ta_entry.c:113:10: warning: nested extern declaration of ‘ta_entry_null_memref’ [-Wnested-externs]                                                                                                                                                                              
cc1: some warnings being treated as errors
```
I.e., this is also needed.

```patch
diff --git a/ta/os_test/include/os_test.h b/ta/os_test/include/os_test.h
index 627a376..62734d5 100644
--- a/ta/os_test/include/os_test.h
+++ b/ta/os_test/include/os_test.h
@@ -42,6 +42,7 @@ TEE_Result ta_entry_ta2ta_memref(uint32_t param_types, TEE_Param params[4]);
 TEE_Result ta_entry_ta2ta_memref_mix(uint32_t param_types,
 				     TEE_Param params[4]);
 TEE_Result ta_entry_params(uint32_t param_types, TEE_Param params[4]);
+TEE_Result ta_entry_null_memref(uint32_t param_types, TEE_Param params[4]);
 TEE_Result ta_entry_params_tmpmem(uint32_t param_types, TEE_Param params[4]);
 TEE_Result ta_entry_call_lib(uint32_t param_types, TEE_Param params[4]);
 TEE_Result ta_entry_call_lib_panic(uint32_t param_types, TEE_Param params[4]);

```

jbech-linaro:
Testing re-done after rm -rf out-br and rebase + squashed this and fixed build error above, still no regressions, so my TB tag still holds.

etienne-lms:
> I.e., this is also needed.

To fix as:
```patch
+TEE_Result ta_entry_null_memref(uint32_t param_types, TEE_Param params[4]);
-TEE_Result ta_entry_params_tmpmem(uint32_t param_types, TEE_Param params[4]);

**Comment by jbech-linaro** [comment]
> ```diff
> ta_entry_params_tmpmem
> ```

Yes, the "minus" is also needed, I remember your review comment about changing the name. Thanks!

**Comment by cneveux** [comment]
Squashed and tagged.

**Comment by jforissier** [comment]
@jbech-linaro I suppose the manifests need to be updated for IBART to be able to test this? (linux points to an SHA-1, not to the tip of the optee branch).

**Comment by jbech-linaro** [comment]
> @jbech-linaro I suppose the manifests need to be updated for IBART to be able to test this? (linux points to an SHA-1, not to the tip of the optee branch).

Indeed, not only for IBART, for all our builds, since we have 1025 that actually will test it.

**Comment by jforissier** [comment]
OK, how about changing all manifests that use `linaro-swg/linux.git` to point to branch `optee` instead of a fixed commit? There is no reason not to track latest for all the Linaro stuff, `linux.git` looks like it is currently an unwanted exception.

**Comment by jbech-linaro** [comment]
> OK, how about changing all manifests that use `linaro-swg/linux.git` to point to branch `optee` instead of a fixed commit?

In fact that is what I'd prefer, so +1. We used to track master on everything, in hindsight I think that was better than what we are doing now. Sure, we don't have to deal with issues in other gits that often. But now when the delta are getting bigger, I'm afraid it can be a minor nightmare to update them all. But for linux.git tracking the "optee" branch +1.

**Comment by jforissier** [comment]
IBART is failing due to insufficient memory in the lockdep code. It's annoying but not directly related to the feature tested here obviously. I need to take a closer look in case there would be a leak. But perhaps `CFG_TEE_CORE_HEAP_SIZE` just needs to be increased a bit.

**Comment by jforissier** [comment]
> IBART is failing [...] not directly related to the feature tested here [...]

Uh oh, I failed to notice:
```
    67:  * regression_1006 Test Basic OS features
    68:  ta_entry_basic: enter
    69:  Getting properties for current TA
    70:  Getting properties for current client
    71:  Getting properties for implementation
    72:  E/TC:? 0 
    73:  E/TC:? 0 TA panicked with code 0xffff0000
    74:  E/LD:  Status of TA 5b9e0e40-2636-11e1-ad9e-0002a5d5c51b
    75:  E/LD:   arch: arm
    76:  E/LD:  region  0: va 0x40004000 pa 0x00000000 size 0x002000 flags rw-s (ldelf)
    77:  E/LD:  region  1: va 0x40006000 pa 0x00000000 size 0x007000 flags r-xs (ldelf)
    78:  E/LD:  region  2: va 0x4000d000 pa 0x00000000 size 0x001000 flags rw-s (ldelf)
    79:  E/LD:  region  3: va 0x4000e000 pa 0x00000000 size 0x003000 flags rw-s (ldelf)
    80:  E/LD:  region  4: va 0x40011000 pa 0x00000000 size 0x001000 flags r--s
    81:  E/LD:  region  5: va 0x40012000 pa 0x00000000 size 0x003000 flags rw-s (stack)
    82:  E/LD:  region  6: va 0x40015000 pa 0x384db930 size 0x001000 flags rw-- (param)
    83:  E/LD:  region  7: va 0x40053000 pa 0x00000000 size 0x002000 flags r-xs [1]
    84:  E/LD:  region  8: va 0x40055000 pa 0x00001000 size 0x002000 flags rw-s [1]
    85:  E/LD:  region  9: va 0x40063000 pa 0x00001000 size 0x00e000 flags r-xs [0]
    86:  E/LD:  region 10: va 0x40071000 pa 0x0000f000 size 0x0e5000 flags rw-s [0]
    87:  E/LD:   [0] 5b9e0e40-2636-11e1-ad9e-0002a5d5c51b @ 0x40063000
    88:  E/LD:   [1] ffd2bded-ab7d-4988-95ee-e4962fff7154 @ 0x40053000
    89:  E/LD:  Call stack:
    90:  E/LD:   0x40053e8c
    91:  E/LD:   0x40067f55
    92:  E/LD:   0x40063a93
    93:  regression_1000.c:563: TEEC_InvokeCommand(&session, 5, &op, &ret_orig) has an unexpected value: 0xffff3024 = TEE_E[    7.400380] random: fast init done
    94:  RROR_TARGET_DEAD, expected 0x0 = TEEC_SUCCESS
    95:    regression_1006 FAILED
```

I can reproduce the error with QEMU and QEMUv8.

**Comment by jforissier** [comment]
@cneveux the regression is introduced by [9d2e798360b5](https://github.com/OP-TEE/optee_os/commit/9d2e798360b5b8dd2f5032cc020bb2bb6f7c3f06) ("core: TEE capability for null sized memrefs support"), if reverted `xtest 1006` passes.

After a quick look it seems this is the failing call in xtest: [here](https://github.com/OP-TEE/optee_test/blob/master/ta/os_test/os_test.c#L475-L483), causes the TA to panic. Something seems wrong in optee_os `tee_mmu_map_param()`, because it returns [here](https://github.com/OP-TEE/optee_os/blob/9d2e798360b5b8dd2f5032cc020bb2bb6f7c3f06/core/arch/arm/mm/tee_mmu.c#L951). But at this point I have reached the limits of my understanding of the code :-/

Can you please check this?

**Comment by cneveux** [comment]
OK, I'll have a look

**Comment by jbech-linaro** [comment]
Not sure if I'm onto something here, but it seems like we've been able to mix up the capability even though we had it up for discussions in review comments.

```bash
$ rg _CAP_MEMREF_NULL | grep define
optee_client/libteec/include/linux/tee.h:#define TEE_GEN_CAP_MEMREF_NULL        (1 << 3) /* Support NULL MemRef */
linux/include/uapi/linux/tee.h:#define TEE_GEN_CAP_MEMREF_NULL  (1 << 3)/* NULL MemRef support */
linux/drivers/tee/optee/optee_smc.h:#define OPTEE_SMC_SEC_CAP_MEMREF_NULL               BIT(4)
optee_os/core/arch/arm/include/sm/optee_smc.h:#define OPTEE_SMC_SEC_CAP_MEMREF_NULL             (1 << 4)
```
`OPTEE_SMC_SEC_CAP_VIRTUALIZATION`  has been assigned to '1 << 3', so  everything above should be `(1 << 4)`, shouldn't it?

jforissier:
@jbech-linaro that would perhaps be more consistent but I tried to use (1 << 4) everywhere and it's the same. AFAICT the current code is not really wrong, the user space/kernel interface uses (1 << 3) while the kernel/TEE interface uses (1 << 4).

jbech-linaro:
> the user space/kernel interface uses (1 << 3) while the kernel/TEE interface uses (1 << 4).

Oh, I didn't even reflect that they could be different. This seems to be a source for confusion, we have things defined in 4 files and they are same ... but different. Maybe we should try to align them to avoid confusion?

jenswi-linaro:
The user space ABI is completely unhooked from the secure world ABI. So the confusion is more that they are similar enough to be confusing.

cneveux:
Please refer to #3729

PR #337, opened 2019-03-19
[RFC] build TAs from CMake

etienne-lms:
```
This change allows one to build the optee_test host and TA parts from
CMake. TA SDK does not yet support CMake builds. This change uses the
legacy GNU makefile ways to build TAs but launched from a custom
target/commands defined in the CMake scripts.

Original ta/Makefile is renamed ta/Makefile.tas so that it can be
called from a CMake build even once all Makefiles have been generated.
For backward compatibility, ta/Makefile explicitly includes
ta/Makefile.tas.

ta/Makefile.tas is modified with a extra make rule: pre-install.
This rule stores each built TA a intermediate directory. At install time,
content of this intermediate directory is installed at target TA
binaries path.
```

This RFC aims at simplifying build setups that rely on CMake. This is inspired from @jbech-linaro work on https://github.com/jbech-linaro/optee_standalone/.

An equivalent RFC is proposed on optee_examples: https://github.com/linaro-swg/optee_examples/pull/56.


etienne-lms:
Requires update in **build**: https://github.com/OP-TEE/build/pull/353.

jforissier:
> TA SDK does not yet support CMake builds.

How about extending the TA SDK to support CMake instead? If it could make it easier for users of CMake to build OP-TEE apps, it could be a better option.

> Original ta/Makefile is renamed ta/Makefile.tas so that it can be
> called from a CMake build even once all Makefiles have been generated.

Do you mean that `ta/Makefile` gets overwritten by CMake? If so, it's bad.

etienne-lms:
> > TA SDK does not yet support CMake builds.
> 
> How about extending the TA SDK to support CMake instead? If it could make it easier for users of CMake to build OP-TEE apps, it could be a better option.

This is a big task. SDK is produced from optee_os source tree and based on core build scheme.
Until it's available we need to go though private makefiles. 

This RFC proposes a intermediate step: caller get all build from a single cmake build sequence, instead of using a cmake + a custo TA build.


> > Original ta/Makefile is renamed ta/Makefile.tas so that it can be called from a CMake build even once all Makefiles have been generated.
> 
> Do you mean that ta/Makefile gets overwritten by CMake? If so, it's bad.

It does, if building in source tree. Remember this [thread](https://github.com/OP-TEE/optee_test/pull/267#issuecomment-373434263).
When you use CMake, it generates `Makefile` files for each `CMakeLists.txt` found. `ta/CMakeLists.txt` is used to export TA header files to host CA builds, and it can scratch `ta/Makefile`.

jforissier:
>> > TA SDK does not yet support CMake builds.
>> 
>> How about extending the TA SDK to support CMake instead? If it could make it easier for users of CMake to build OP-TEE apps, it could be a better option.
>
> This is a big task. SDK is produced from optee_os source tree and based on core build scheme.

I'm not saying optee_os should be built with CMake, but simply that it should export some files in CMake syntax containing the rules to build TAs. Essentially, this:
```
$ ls out/arm/export-ta_arm32/mk/
cleandirs.mk  compile.mk  conf.mk  gcc.mk  link.mk  link_shlib.mk  subdir.mk  ta_dev_kit.mk

```

> > Do you mean that ta/Makefile gets overwritten by CMake? If so, it's bad.
>
> It does, if building in source tree. Remember this thread.

So if we don't build in source tree there is no problem. OK, that's acceptable I think. We can count on `build.git` to do the right thing.

jbech-linaro:
> > > Do you mean that ta/Makefile gets overwritten by CMake? If so, it's bad.
> > 
> > 
> > It does, if building in source tree. Remember this thread.
> 
> So if we don't build in source tree there is no problem. OK, that's acceptable I think. We can count on `build.git` to do the right thing.

You shouldn't build with CMake from the "root" directory. When using CMake you shall always create another temporary build folder and refer to the root by `..`, like what I'm doing [here](https://github.com/jbech-linaro/optee_standalone/blob/cmake_optee/Makefile.cmake#L33-L35).

etienne-lms:
Assuming the initial intention of this RFC: allow one to build optee_test from a single cmake build, that is for example:
```
mkdir -p $(CMAKE_OUT) && cd $(CMAKE_OUT) && \
	cmake $(CMAKE_FLAGS) .. && \
	make &&
	make DESTDIR=some/path install
```
The first `make` command will not see optee_test original `ta/Makefile` but the generated `ta/Makefile`  in the output build dir `$(CMAKE_OUT)`. Therefore the renaming proposed in this P-R.

I would like to `make`/`make install`/`make clean` after cmake pre-processing to process both host and TA parts.

jbech-linaro:
> The first `make` command will not see optee_test original `ta/Makefile` but the generated `ta/Makefile` in the output build dir `$(CMAKE_OUT)`.

Alright, I didn't realize that it was that was going on. Anyhow, this is tricky and hard to do in a non-hacky way. The ultimate goal would be to:
```bash
$ git clone optee_test
$ mkdir -p optee_test/build
$ cd optee_test/build
$ cmake ..
$ make
```
But that is not really possible and even trying to trigger the build of optee_os and optee_client as part of that is not easy at all.

etienne-lms:
If you consider all OP-TEE packages but optee_os (that still needs a non cmake `make`), we could reach such a `cmake .. && make` way.

For build environments (yocto, buildroot, etc...) that means:
- optee_os build recipe is a regular `make` build process
- other optee_* build recipes are each regular `cmake` build process

No package would need a mix of private `make` and `cmake`.

etienne-lms:
> But that is not really possible and even trying to trigger the build of optee_os and optee_client as part of that is not easy at all.

Sorry, i did not answer your comment.
I think that when using cmake, you never build packages alone, but always all in a global cmake env. (Am I right?)

If optee_os was part of this cmake build env, then all deps, install, would be handle.

Current optee_os is not. But deps on optee_os from the other optee_* all limited to the TA SDK path. optee_test already expects this path as `OPTEE_TEST_SDK`. A unique cmake config shared by all optee_* would be nice, i.e `OPTEE_TA_SDK`.

jforissier:
>  A unique cmake config shared by all optee_* would be nice, i.e `OPTEE_TA_SDK`.

That's what I was suggesting above -- https://github.com/OP-TEE/optee_test/pull/337#issuecomment-474842514.

etienne-lms:
> > A unique cmake config shared by all optee_* would be nice, i.e OPTEE_TA_SDK.
> 
> That's what I was suggesting above -- #337 (comment).

I don't think SDK to provide cmake scripts is a solution here. SDK already contains a cmake script:  `conf.cmake`. But as optee_os itself does not build from cmake, we still need to set a specific directive for SDK path when building other packages.
Something like:
```shell
cmake -DOPTEE_TA_SDK=path/to/prebuilt/sdk ..
make
make DESTDIR=some/path install
```

Aside from that, whether TA SDK builds TAs from cmake makefile or from our legacy Makefile, that's TA SDK internal stuff. Even if TAs where built from cmake makefiles, the SDK path still needs to be manually provided unless optee_os is also built from cmake.

jbech-linaro:
As I think I mentioned in an email or SWG weekly, the major headache is related to such a config, namely this include:
https://github.com/OP-TEE/optee_test/blob/master/host/xtest/CMakeLists.txt#L3

That file is only to be found after successfully building optee_os. I.e., there is a strong dependency there. Due to that your are forced to:
- 1: cd optee_os && make, then cd optee_test && cmake && make

Since CMake parse includes etc on initial invocation, you **cannot** do.

- 2: cd optee_test && cmake (we will get an error message already here) && make, which is modified to
   - build optee_os first
   - then build the optee_test components.

Without the include I referred to above or if we would have had a static config, then "2" would work.

If we could have something like defconfigs that contained most flags, then the rest could be sent as command line parameters when invoking CMake. We can put this on hold of a week or so, since I've proposed a hacking session that is exactly about this.

jforissier:
> cmake -DOPTEE_TA_SDK=path/to/prebuilt/sdk ..
> make
> make DESTDIR=some/path install

Looks good, is there any way to set a default value for OPTEE_SDK_PATH, assuming the "usual" layout for Git repositories (optee_os, optee_client etc. in the same directory)?

jforissier:
> That file is only to be found after successfully building optee_os. I.e., there is a strong dependency there. 

Ah, got it now.

>  cd optee_os && make, then cd optee_test && cmake && make

Well we could have a build script to do just that?

> We can put this on hold of a week or so, since I've proposed a hacking session that is exactly about this.

Yep.

jbech-linaro:
> > cd optee_os && make, then cd optee_test && cmake && make
> 
> Well we could have a build script to do just that?

Of course, that is one way to do it and that is basically what I've done [here](https://github.com/jbech-linaro/optee_standalone/blob/cmake_optee/Makefile.cmake) ("standalone") and that is also what is done in `build.git` (not standalone). There is a [dependency rule](https://github.com/OP-TEE/build/blob/master/common.mk#L200) to build `optee_os` before building Buildroot. But what I would hope is that it would be easy to build without doing to many external hacks like that. What makes this extra complicated is that we have many flags that needs to be set correctly depending on how optee_os was built and where optee_os can be found. That makes it pretty hard to build things standalone if you're not a core OP-TEE hacker. The "cleanest" I managed to end up with is [this](https://github.com/jbech-linaro/optee_standalone/blob/cmake_optee/Makefile.cmake#L15-L19) and [this](https://github.com/jbech-linaro/optee_standalone/blob/cmake_optee/Makefile.cmake#L27). That is still quite fragile and it's not possible to use/build with other flags. In fact I had to create [another patch in optee_test](https://github.com/jbech-linaro/optee_test/commit/58ccb8bf53226ef2e8573f83f79291b3d7dbab3a) to be able to make that hack work.

Ideally you shall be able to build a CMake git/project like [my comment above](https://github.com/OP-TEE/optee_test/pull/337#issuecomment-474926754). That's all you need to do for most other CMake based projects that I've tried. Sometimes there is a lib or so that they rely on, but that is usually just a single parameter that needs to be added and not a plethora of parameters that you need to provide.

I think we need to take a step back and look at this from a higher view, I.e., think more in terms of What do we want to support? What and how should CMake builds work? What and how should regular Make work and so on (AOSP, OE ...). I don't think it's possible to end up with a universal system that solves all our problems. But I believe that if we do some brainstorming we can probably come up with something that is easier (for people who are not using our "OP-TEE developer setup") to work with and a bit more well defined.

ta/CMakeLists.txt
```diff
@@ -18,3 +18,25 @@ target_include_directories(${PROJECT_NAME}
...
+	TARGET build-tas
+	COMMAND make -f Makefile.tas O=${CMAKE_CURRENT_SOURCE_DIR}/out TA_DEV_KIT_DIR=${OPTEE_TEST_SDK} all
+	COMMENT "Compiling xtest Trusted Applications"
+	VERBATIM
+	)
+
+install (
+	DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/out/ta-binaries/ DESTINATION /lib/optee_armtz
+	PATTERN "ta-binaries/*"
+	)
```
jforissier: I'd rather invoke `make -f Makefiles.tas ... install DESTDIR=$(TARGET_DIR)`. Avoids needless copies, and can be useful on its own in the GNU makefile.

ta/Makefile.tas
```diff
@@ -0,0 +1,82 @@
...
+define rm-build-subdir
+echo `[ -d $1 ] && { cd $1; $(RMDIR) $(out-dir)/ta $(out-dir) 2> /dev/null; true; }` > /dev/null
+endef
+
+.PHONY: clean
+clean:
+	$(q)$(foreach dir,$(TA_DIRS), [ ! -d $(dir) ] || $(MAKE) -C $(dir) O=$(out-dir)/ta/$(dir) $@ &&) true
+	$(q)$(foreach dir,$(TA_DIRS), $(call rm-build-subdir,$(dir));)
+	$(q)$(foreach ta,$(wildcard $(out-dir)/ta/$(dir)/*.ta), rm -f $(out-dir)/ta-binaries/$(basename $(ta)) &&) true
+	$(q)rm-build-subdir,$(out-dir)/ta-binaries
```
jforissier: ...and finally these two last lines are not needed.

ta/Makefile.tas
```diff
@@ -0,0 +1,82 @@
...
+all: ta pre-install
+
+.PHONY: ta
+ta:
+	$(q)$(foreach dir,$(TA_DIRS), $(MAKE) -C $(dir) O=$(out-dir)/ta/$(dir) &&) true
+
+.PHONY: pre-install
+pre-install:
+	mkdir -p $(out-dir)/ta-binaries
+	$(foreach ta,$(wildcard $(out-dir)/ta/*/*.ta), cp -a $(ta) $(out-dir)/ta-binaries &&) true
```
jforissier: ...and here, call this `install` and copy to `(DESTDIR)/lib/optee_armtz`.

ta/Makefile.tas
```diff
@@ -0,0 +1,82 @@
...
+	   GP_TTA_Time \
+	   GP_TTA_answerErrorTo_Invoke \
+	   GP_TTA_answerErrorTo_OpenSession \
+	   GP_TTA_answerSuccessTo_OpenSession_Invoke \
+	   GP_TTA_check_OpenSession_with_4_parameters \
+	   GP_TTA_testingClientAPI
+endif
+
+.PHONY: all
+all: ta pre-install
```
jforissier: Assuming the above change to the `CMakeList.txt` `install` recipe: remove `pre-install`...

PR #338, opened 2019-03-20
[RFC] Add test case for dlopen()/dlsym()/dlclose() (1021)

jforissier:
Adds a new test case for the dynamic link API in libdl. This is enabled
when CFG_TA_DL=y.
- A new shared library is added in ta/os_test_lib_dl. It contains two
  funtions: one performs a simple operation (adds two integers) and
  the other one calls TEE_Panic(), which triggers a nice crash dump on
  the secure console, allowing to check how things are actually mapped
  in the TA memory.
- In xtest 1021, the dl API is used to load the test library and
  retrieve pointers to the functions which are then executed. dlsym()
  is also used to get a pointer to a function in the main executable.

### DIFF
--- a/host/xtest/regression_1000.c
+++ b/host/xtest/regression_1000.c
@@ -1473,3 +1473,31 @@ out:
 }
 ADBG_CASE_DEFINE(regression, 1020, xtest_tee_test_1020,
 		"Test lockdep algorithm");
+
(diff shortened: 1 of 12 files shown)


jforissier:
Superseeded by #358.

PR #339, opened 2019-03-28
xtest: Explicitly initialize local variables

etienne-lms:
This change initializes all local variables to prevent build issues
(warnings and/or errors) in OP-TEE test package. It uses memset()
to initialize structured and typed variables.

This change replaces all TEEC_Session instance initialization from
= { 0 } into use of memset().

### DIFF
--- a/host/xtest/regression_1000.c
+++ b/host/xtest/regression_1000.c
@@ -51,8 +51,8 @@ struct xtest_crypto_session {
 
 static void xtest_crypto_test(struct xtest_crypto_session *cs)
 {
-	uint32_t ret_orig;
-	uint8_t crypt_out[16];
+	uint32_t ret_orig = 0;
+	uint8_t crypt_out[16] = { };
 	uint8_t crypt_in[16] = { 22, 17 };
 
 	crypt_in[15] = 60;
@@ -81,7 +81,7 @@ static void xtest_crypto_test(struct xtest_crypto_session *cs)
 	Do_ADBG_BeginSubCase(cs->c, "AES decrypt");
 	{
 		TEEC_Operation op = TEEC_OPERATION_INITIALIZER;
-		uint8_t out[16];
+		uint8_t out[16] = { };
 
 		op.params[0].tmpref.buffer = crypt_out;
 		op.params[0].tmpref.size = sizeof(crypt_out);
@@ -118,7 +118,7 @@ static void xtest_crypto_test(struct xtest_crypto_session *cs)
 			0xb0, 0x03, 0x61, 0xa3, 0x96, 0x17, 0x7a, 0x9c,
 			0xb4, 0x10, 0xff, 0x61, 0xf2, 0x00, 0x15, 0xad
 		};
-		uint8_t out[32] = { 0 };
+		uint8_t out[32] = { };
 
 		op.params[0].tmpref.buffer = (void *)sha256_in;
 		op.params[0].tmpref.size = sizeof(sha256_in);
@@ -160,7 +160,7 @@ static void xtest_crypto_test(struct xtest_crypto_session *cs)
 			0xE9, 0xC3, 0xEF, 0x8A, 0xB2, 0x34, 0x53, 0xE6,
 			0xF0, 0x74, 0x9C, 0xD6, 0x36, 0xE7, 0xA8, 0x8E
 		};
-		uint8_t out[sizeof(exp_out)];
+		uint8_t out[sizeof(exp_out)] = { };
 
 		op.params[0].tmpref.buffer = (void *)in;
 		op.params[0].tmpref.size = sizeof(in);
@@ -202,7 +202,7 @@ static void xtest_crypto_test(struct xtest_crypto_session *cs)
 			0x10, 0x11, 0x12, 0x13, 0x14, 0x15, 0x16, 0x17,
 			0x18, 0x19, 0x1a, 0x1b, 0x1c, 0x1d, 0x1e, 0x1f
 		};
-		uint8_t out[sizeof(exp_out)];
+		uint8_t out[sizeof(exp_out)] = { };
 
 		op.params[0].tmpref.buffer = (void *)in;
 		op.params[0].tmpref.size = sizeof(in);
@@ -232,9 +232,9 @@ static void xtest_crypto_test(struct xtest_crypto_session *cs)
 
 static void xtest_tee_test_1001(ADBG_Case_t *c)
 {
-	TEEC_Result res;
-	TEEC_Session session = { 0 };
-	uint32_t ret_orig;
+	TEEC_Result res = TEEC_ERROR_GENERIC;
+	TEEC_Session session = { };
+	uint32_t ret_orig = 0;
 
 	/* Pseudo TA is optional: warn and nicely exit if not found */
 	res = xtest_teec_open_session(&session, &pta_invoke_tests_ta_uuid, NULL,
@@ -253,13 +253,13 @@ ADBG_CASE_DEFINE(regression, 1001, xtest_tee_test_1001, "Core self tests");
 
 static void xtest_tee_test_1002(ADBG_Case_t *c)
 {
-	TEEC_Result res;
-	TEEC_Session session = { 0 };
+	TEEC_Result res = TEEC_ERROR_GENERIC;
+	TEEC_Session session = { };
 	TEEC_Operation op = TEEC_OPERATION_INITIALIZER;
-	uint32_t ret_orig;
-	uint8_t buf[16 * 1024];
+	uint32_t ret_orig = 0;
+	uint8_t buf[16 * 1024] = { };
 	uint8_t exp_sum = 0;
-	size_t n;
+	size_t n = 0;
 
 	/* Pseudo TA is optional: warn and nicely exit if not found */
 	res = xtest_teec_open_session(&session, &pta_invoke_tests_ta_uuid, NULL,
@@ -304,9 +304,9 @@ struct test_1003_arg {
 static void *test_1003_thread(void *arg)
 {
 	struct test_1003_arg *a = arg;
-	TEEC_Session session = { 0 };
+	TEEC_Session session = { };
 	size_t rounds = 64 * 1024;
-	size_t n;
+	size_t n = 0;
 
 	a->res = xtest_teec_open_session(&session, &pta_invoke_tests_ta_uuid,
 					 NULL, &a->error_orig);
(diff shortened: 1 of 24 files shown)


jenswi-linaro:
There's a compile error

etienne-lms:
Yes, I saw. Sorry. I reviewed, tested and pushed to get the CI status. but I left some bad things as tested pointed out. Will fix.

host/xtest/regression_1000.c
```diff
@@ -355,28 +349,27 @@ static void *test_1003_thread(void *arg)
 	return NULL;
 }
 
+#define TEST_1003_THREAD_COUNT		(3 * 2)
+
 static void xtest_tee_test_1003(ADBG_Case_t *c)
 {
-	size_t num_threads = 3 * 2;
```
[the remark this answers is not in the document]
jenswi-linaro: Fair enough.

host/xtest/regression_8100.c
```diff
@@ -189,9 +187,7 @@ static bool check(ADBG_Case_t *c, STACK_OF(X509) *trusted,
 	X509_STORE_CTX *csc = NULL;
 	X509_VERIFY_PARAM *pm = NULL;
 	int i = 0;
-	time_t vfy_time;
-
-	memset(&vfy_time, 0, sizeof(vfy_time));
+	time_t vfy_time = { 0 };
```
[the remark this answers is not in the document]
jenswi-linaro: So it should be initialized with `0` instead.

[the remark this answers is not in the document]
etienne-lms: Why not using universal initializer? It makes this code agnostic of `time_t` effective type.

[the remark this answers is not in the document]
jenswi-linaro: `time_t` is a scalar as required by the C standard. Universal initializer is not the first choice.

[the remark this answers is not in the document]
etienne-lms: I understand that:
* `= { 0 }` is a C standard zeroies initializer for non-scalar variables.
* scalar variable mandates a scaler init value, i.e. `= 0`.
* some (maybe) opaque defined types as `time_t` need scalar treatment as scalar per standard.
* I guess there are standard types that are non-scalar per standard, as mandates `{ 0 }`.
* some types as `phtread_t` may depend on toolchain and need common scalar/non-scalar initializer. Here, look like `memset()` is a way.

Do you agree with this statement? And I to update on going P-Rs (#339, #340, https://github.com/OP-TEE/optee_client/pull/154) accordingly?

Note: An earlier comment worried about toolchain complain when using `{ 0 }` over a `struct { void *ptr; }` type. Should not worry as `{ 0 }` is a universal form of `{ NULL }`.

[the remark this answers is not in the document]
jenswi-linaro: Here's my take on this:

Types like `time_t` which are standardized as a scalar (and all other scalars for that matter) are initialized with `0`, unless another value makes more sense.

For optee_client we need maximum portability, so only initialize struct types (and `pthread_t`) with `memset()` unless there's a good reason not to do so.

For the rest of the gits we assume that a recent version of GCC or Clang is used so we initialize structs with `{ }` in order to avoid the more clumsy `memset()` procedure. Types like `pthread_t` which can be a scalar or a composite type are initialized with `memset()` in order to minimize the amount of future headache.

I don't think we need to use `{ 0 }` in any situation, if there is one we're probably better off with `memset()` instead (minimize future headache clause).

Let's take one PR at a time to avoid too many changes back and forth.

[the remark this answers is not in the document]
jforissier: @jenswi-linaro sounds good to me. Assuming we agree on this, I think this text should be added to https://github.com/OP-TEE/optee_docs/blob/master/general/coding_standards.rst.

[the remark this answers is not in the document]
etienne-lms: > Let's take one PR at a time to avoid too many changes back and forth.

ok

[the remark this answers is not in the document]
jenswi-linaro: PR https://github.com/OP-TEE/optee_docs/pull/19 created

host/xtest/regression_1000.c
```diff
@@ -419,7 +420,7 @@ static void xtest_tee_test_1003(ADBG_Case_t *c)
 			    (double)(repeat * num_readers);
 
 	Do_ADBG_Log("    Number of parallel threads: %zu (%zu writers and %zu readers)",
-		    num_threads, num_writers, num_readers);
+		    TEST_1003_THREAD_COUNT, num_writers, num_readers);
```
jenswi-linaro: `TEST_1003_THREAD_COUNT` should be formatted as an `int` (`%d`) instead of `size_t`

host/xtest/regression_1000.c
```diff
@@ -1087,21 +1087,22 @@ static void xtest_tee_test_1013_single(ADBG_Case_t *c, double *mean_concurrency,
 		0xb4, 0x10, 0xff, 0x61, 0xf2, 0x00, 0x15, 0xad
 	};
 	uint8_t out[32] = { 0 };
```
jenswi-linaro: `{ }`

host/xtest/regression_2000.c
```diff
@@ -42,11 +42,9 @@ static TEE_Result socket_tcp_open(TEEC_Session *session, uint32_t ip_vers,
 				  struct socket_handle *handle,
 				  uint32_t *error, uint32_t *ret_orig)
 {
-	TEE_Result res;
+	TEE_Result res = TEE_ERROR_GENERIC;
 	TEEC_Operation op = TEEC_OPERATION_INITIALIZER;
 
-	memset(handle, 0, sizeof(*handle));
```
jenswi-linaro: Why isn't this needed any longer?

host/xtest/regression_2000.c
```diff
@@ -72,11 +70,9 @@ static TEE_Result socket_udp_open(TEEC_Session *session, uint32_t ip_vers,
...
```
jenswi-linaro: Why isn't this needed any longer?

host/xtest/regression_2000.c
```diff
@@ -511,11 +507,13 @@ static void *xtest_tee_test_2002_thread(void *arg)
...
 {
-	pthread_t thr[NUM_THREADS];
 	struct test_2002_barrier bar = { .mu = PTHREAD_MUTEX_INITIALIZER };
-	struct test_2002_arg arg[NUM_THREADS];
-	size_t n;
-	size_t nt;
+	struct test_2002_arg arg[NUM_THREADS] = { };
+	size_t n = 0;
+	size_t nt = 0;
+	pthread_t thr[NUM_THREADS];
```
jenswi-linaro: I wonder if this couldn't be initialized with `{ }`, given that it's an array.

host/xtest/regression_6000.c
```diff
@@ -1477,46 +1478,50 @@ ADBG_CASE_DEFINE(regression, 6010, xtest_tee_test_6010, "Test Storage");
...
+	if (!ADBG_EXPECT_TEEC_SUCCESS(
+			c, xtest_teec_open_session(&sess, &gp_tta_ds_uuid,
+						   NULL, &orig)))
+		return;
 
-    if (!ADBG_EXPECT_TEEC_SUCCESS(c, ds_init_object_and_attributes(&sess,
-            0xa0000010, 0x100, attr_meta, sizeof(attr_meta), attr_data,
-            sizeof(attr_data), 0)))
-        goto exit;
+	if (!ADBG_EXPECT_TEEC_SUCCESS(
```
jenswi-linaro: I'm not so sure this reformatting is an improvement.

host/xtest/regression_6000.c
```diff
@@ -1911,6 +1915,7 @@ static void xtest_tee_test_6018_single(ADBG_Case_t *c, uint32_t storage_id)
 		uint8_t br[block_size];
 		uint32_t count = 0;
 
+		memset(br, 0, sizeof(br));
```
jenswi-linaro: How about using an initializer above instead?

host/xtest/sdp_basic.c
```diff
@@ -338,10 +335,10 @@ static int dump_sdp_data(struct tee_ctx *ctx,
...
 {
 	TEEC_SharedMemory *shm = (TEEC_SharedMemory *)shm_ref;
-	TEEC_Result teerc;
-	TEEC_Operation op;
-	uint32_t err_origin;
-	unsigned cmd;
+	TEEC_Result teerc = TEEC_ERROR_GENERIC;
+	TEEC_Operation op = TEEC_OPERATION_INITIALIZER;
+	uint32_t err_origin = 0;
+	unsigned cmd = 0;
```
jenswi-linaro: Please make this `unsigned int` while you're at it.

host/xtest/regression_1000.c
```diff
@@ -81,7 +81,7 @@ static void xtest_crypto_test(struct xtest_crypto_session *cs)
 	Do_ADBG_BeginSubCase(cs->c, "AES decrypt");
 	{
 		TEEC_Operation op = TEEC_OPERATION_INITIALIZER;
-		uint8_t out[16];
+		uint8_t out[16] = { 0 };
```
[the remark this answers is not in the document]
etienne-lms: yes, here at at several other places.

host/xtest/regression_1000.c
etienne-lms: ok, thanks.

host/xtest/regression_1000.c
etienne-lms: ditto

host/xtest/regression_2000.c
etienne-lms: true, sorry. thanks to the catch.

host/xtest/regression_2000.c
etienne-lms: thanks again.

host/xtest/regression_2000.c
etienne-lms: good suggestion. I'll do.

host/xtest/regression_6000.c
etienne-lms: ok, will propose another.

host/xtest/sdp_basic.c
etienne-lms: ok, here and line 301 above.

host/xtest/regression_6000.c
etienne-lms: because the table size is considered as variable, our current toolchain maybe initialize properly:
`... error: variable-sized object may not be initialized`

jenswi-linaro: Fair enough

jenswi-linaro:

etienne-lms:
Fixup commits squashed.
Commit message updated:
```
This change initializes all local variables to prevent build issues
(warnings and/or errors) in OP-TEE test package. This change uses
= { } for non-scalar variables and = 0 for scalar variables when there
is no obvious value. This change falls back to memset() only when
neither = { } nor = 0 is applicable, here pthread_t typed variable
and variable size arrays.
```
Tag applied.

PR #340, opened 2019-03-28
TAs: explicitly initialize local variables

etienne-lms:
This change initializes all local variables to prevent build issues
(warnings and/or errors) in OP-TEE test package. It uses TEE_MemFill()
to initialize structured and typed variables.

### DIFF
--- a/ta/crypt/handle.c
+++ b/ta/crypt/handle.c
@@ -25,9 +25,9 @@ void handle_db_destroy(struct handle_db *db)
 
 int handle_get(struct handle_db *db, void *ptr)
 {
-	size_t n;
-	void *p;
-	size_t new_max_ptrs;
+	size_t n = 0;
+	void *p = NULL;
+	size_t new_max_ptrs = 0;
 
 	if (!db || !ptr)
 		return -1;
@@ -60,7 +60,7 @@ int handle_get(struct handle_db *db, void *ptr)
 
 void *handle_put(struct handle_db *db, int handle)
 {
-	void *p;
+	void *p = NULL;
 
 	if (!db || handle < 0 || (size_t)handle >= db->max_ptrs)
 		return NULL;
--- a/ta/os_test/include/testframework.h
+++ b/ta/os_test/include/testframework.h
@@ -34,7 +34,6 @@
 
 #include "tb_macros.h"
 #include "tb_asserts.h"
-#include "mpa.h"
 
 /* define the max size of generated numbers */
 /* this is number of hex chars in the number */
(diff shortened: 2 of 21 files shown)


ta/os_test/include/testframework.h
```diff
@@ -34,7 +34,9 @@
 
 #include "tb_macros.h"
 #include "tb_asserts.h"
-#include "mpa.h"
+
+//#include <mpa.h>
```
jenswi-linaro: Please remove

jenswi-linaro:

ta/os_test/include/testframework.h
etienne-lms: Oh... here is another bad thing i left in. Will remove.

etienne-lms: Looks like <string.h> below is not required.
Will submit a fixup commit and let CI check that.

ta/crypt/handle.c
```diff
@@ -60,7 +60,7 @@ int handle_get(struct handle_db *db, void *ptr)
 
 void *handle_put(struct handle_db *db, int handle)
 {
-	void *p;
+	void *p = 0;
```
jenswi-linaro: Please use `NULL` for pointers.

ta/os_test/os_test.c
```diff
@@ -329,7 +323,7 @@ static TEE_Result test_malloc(void)
 static TEE_Result test_properties(void)
 {
 	TEE_Result res = TEE_ERROR_GENERIC;
-	TEE_PropSetHandle h;
+	TEE_PropSetHandle h = { 0 };
```
jenswi-linaro: `{ }`

ta/os_test/os_test.c
```diff
@@ -944,25 +927,21 @@ static void incr_values(size_t bufsize, uint8_t *a, uint8_t *b, uint8_t *c)
...
 {
 	static const TEE_UUID test_uuid = TA_OS_TEST_UUID;
 	TEE_TASessionHandle sess = TEE_HANDLE_NULL;
-	TEE_Param l_params[4];
-	const size_t bufsize = 2 * 1024;
-	uint8_t in[bufsize];
-	uint8_t inout[bufsize];
-	uint8_t out[bufsize];
+	TEE_Param l_params[4] = { };
+	const size_t bufsize = TA2TA_BUF_SIZE;
```
jenswi-linaro: It seems the const variable is unused now.
Any particular reason why you're using a macro instead?

ta/os_test/os_test.c
jenswi-linaro: Right, so `TEE_HANDLE_NULL` should be used instead.

ta/os_test/os_test.c
jenswi-linaro: Fair enough.

ta/os_test/os_test.c
etienne-lms: ok

jenswi-linaro:

etienne-lms:
fixup commits squashed.
commit message updated: 
```
This change initializes all local variables to prevent build issues
(warnings and/or errors) in OP-TEE test package. This change uses
= { } for non-scalar variables and = 0 for scalar variables.

This change also removes some useless header files inclusion.
This change also fixes few bad indentations.
```
tag applied.

PR #341, opened 2019-04-14
Fix paths relative to OPTEE_CLIENT_EXPORT

mimok:
optee_client repository has changed, now `include`, `lib` and `sbin` directories
are placed in `$(OPTEE_CLIENT_EXPORT)/usr` directory. This PR updates the paths
relative to `OPTEE_CLIENT_EXPORT` in `xtest` makefile.

### DIFF
--- a/host/xtest/Makefile
+++ b/host/xtest/Makefile
@@ -95,7 +95,7 @@ CFLAGS += -I./adbg/include
 CFLAGS += -I./xml/include
 CFLAGS += -I$(out-dir)/xtest
 
-CFLAGS += -I$(OPTEE_CLIENT_EXPORT)/include
+CFLAGS += -I$(OPTEE_CLIENT_EXPORT)/usr/include
 CFLAGS += -I$(TA_DEV_KIT_DIR)/host_include
 
 CFLAGS += -I../../ta/include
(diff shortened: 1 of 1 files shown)


jenswi-linaro:

jforissier:

etienne-lms:
Shouldn't it be `OPTEE_CLIENT_EXPORT` that defines where headers and libs are found, including the `usr/` subdir path? Only the build env knows where these files are installed.

jbech-linaro:
> Shouldn't it be `OPTEE_CLIENT_EXPORT` that defines where headers and libs are found, including the `usr/` subdir path? Only the build env knows where these files are installed.

I was thinking the same when I looked at the patch.

jforissier:
@etienne-lms @jbech-linaro good point. I was a bit carried away by the name of the variable I suppose 
 (`OPTEE_CLIENT_EXPORT` which currently corresponds to the `export` dir in the optee_client build output, but that's not the important thing).

The caller should be fixed instead.

mimok:
@etienne-lms, @jbech-linaro, @jforissier: Not sure to understand. Without that PR, I would have to manage two different values for `OPTEE_CLIENT_EXPORT`, one for v3.4.0 and another for v3.5.0. Indeed I have no mean to automatically know where optee_client puts these directories. Another solution would be to override the location of these directories to be sure that they are placed in `usr` directory whatever the actual version of `optee_client` is. 

Another thing, I don't understand how `common.mk` from `build` repo can work as it points to `export` directory instead of `export/usr`

mimok:
@etienne-lms : ok, i'm going to do that. Thanks

PR #344, opened 2019-04-25
regression 41xx: prevent unsafe-loop-optimizations build error

etienne-lms:
Add -Wno-unsafe-loop-optimizations directive since regression_4100.c
fails to build on some recent toolchains as GCC 8.2 with a error
trace like:

/path/to/optee_test/host/xtest/regression_4100.c: In function ‘convert_from_string’:
/path/to/optee_test/host/xtest/regression_4100.c:448:8: error: missed loop optimization, the loop counter may overflow [-Werror=unsafe-loop-optimizations]
  while (spos) {
        ^
/path/to/optee_test/host/xtest/regression_4100.c:455:6: error: missed loop optimization, the loop counter may overflow [-Werror=unsafe-loop-optimizations]
   if (!spos)
      ^

The GNU Makefile build sequence defines -Wno-unsafe-loop-optimizations
for the whole xtest sources while CMake build sequence defines it
specifically for regression_4100.c among xtest source files.

### DIFF
--- a/host/xtest/CMakeLists.txt
+++ b/host/xtest/CMakeLists.txt
@@ -61,6 +61,10 @@ set (SRC
 	xtest_test.c
 )
 
+set_source_files_properties(
+	regression_4100.c PROPERTIES COMPILE_FLAGS -Wno-unsafe-loop-optimizations
+)
+
 if (CFG_GP_SOCKETS)
 	list (APPEND SRC
 		regression_2000.c
--- a/host/xtest/Makefile
+++ b/host/xtest/Makefile
@@ -155,6 +155,7 @@ CFLAGS += -Wall -Wcast-align -Werror \
 	  -Wshadow -Wstrict-prototypes -Wswitch-default \
 	  -Wwrite-strings \
 	  -Wno-declaration-after-statement \
+	  -Wno-unsafe-loop-optimizations \
 	  -Wno-missing-field-initializers -Wno-format-zero-length
 endif
 


jforissier:
I don't see this error with:
```
gcc version 8.2.1 20180802 (GNU Toolchain for the A-profile Architecture 8.2-2018-08 (arm-rel-8.23))
```

(using either the GNU Makefile or CMake, 32 or 4 bits)

What toolchain are you using? Any special optimization flag?

etienne-lms:
Actually, I initially faced this issue using the Builtroot toolchain, see https://github.com/OP-TEE/optee_test/pull/326.) I faced it again this morning when trying to testing some toolchains, hence this P-R. I though I was using gcc 8.2... but since I used a buggy setup using a x86 toolchain: "gcc (Ubuntu 7.3.0-27ubuntu1~18.04) 7.3.0".
Maybe some gcc 7.x are touchy regarding "unsafe-loop-optimizations".

No specific optimization directive.

etienne-lms:
I confirm that using buildroot gcc 7.3.0 and 7.4.0 (depending on Buildroot version), fails to build optee_test with `error: missed loop optimization, the loop counter may overflow`.

jforissier:
With the GCC versions fixed in the commit message (please mention explicitly 7.3.0 and 7.4.0 and also that 8.2.1 is fine, because otherwise it's hard to understand why our "official" configuration which uses 8.2.1 builds fine):

etienne-lms:
Commit message updated and tag applied. (@jforissier, check if the message looks fine to you)

jforissier:
All good, thanks!

PR #335, opened 2019-03-12
Adding tests for panic instance use cases

omihalac:
These tests are validating the implementation proposed in [PR#2838](https://github.com/OP-TEE/optee_os/pull/2838) of releasing the memory in case of a panicked TA instance.
They are trying to cover all possible scenarios in case of:
    - Multiple Instance TAs
    - Single Instance TAs
    - Single Instance Keep Alive TAs

PR #347, opened 2019-05-17
xtest 1021: test panic during session opening + describe TA API.

etienne-lms:


etienne-lms:
Dear all, could we review and merge this if acked?
https://github.com/OP-TEE/optee_os/pull/3067 refers to a panicking TA open session sequence. This P-R brings such test.

ta/sims/include/ta_sims_test.h
```diff
@@ -32,10 +32,58 @@
...
 
-#define TA_SIMS_CMD_READ                 1
-#define TA_SIMS_CMD_WRITE                2
-#define TA_SIMS_CMD_GET_COUNTER          3
-#define TA_SIMS_CMD_PANIC                4
-#define TA_SIMS_OPEN_TA_SESSION          5
+/*
+ * Open session invocation interface
+ *
+ * When openinig a session, client may provide additional request through
```
jbech-linaro: s/openinig/opening/

ta/sims/include/ta_sims_test.h
```diff
@@ -32,10 +32,58 @@
...
+ * bellow lists the parameters usage.
+ *
+ * param#0: [in]	value.a = Request based on below supported values:
+ *			TA_SIMS_CMD_PANIC: TA to panic during session opening
+ */
+
+/* Command invocation interface */
+
+/*
+ * TA_SIMS_CMD_READ	Read a binary blob previosuly writen
```
jbech-linaro: s/previosuly writen/previously written/

ta/sims/include/ta_sims_test.h
```diff
@@ -32,10 +32,58 @@
...
+ *
+ * param#0: [in]	value.a = index of blob in the storage array
+ * param#1: [out]	memref = output buffer for blob
+ */
+#define TA_SIMS_CMD_READ		1
+
+/*
+ * TA_SIMS_CMD_WRITE	Read a binary blob previosuly writen
```
jbech-linaro: s/previosuly writen/previously written/

ta/sims/include/ta_sims_test.h
```diff
@@ -32,10 +32,58 @@
...
-#define TA_SIMS_CMD_WRITE                2
-#define TA_SIMS_CMD_GET_COUNTER          3
-#define TA_SIMS_CMD_PANIC                4
-#define TA_SIMS_OPEN_TA_SESSION          5
+/*
+ * Open session invocation interface
+ *
+ * When openinig a session, client may provide additional request through
+ * the TEE invocation parameters. When parameter types are not all of type None
+ * bellow lists the parameters usage.
```
jbech-linaro: s/bellow/below/

ta/sims/ta_entry.c
```diff
@@ -45,18 +45,16 @@ void TA_DestroyEntryPoint(void)
 }
 
 /* Called each time a session is opened */
-TEE_Result TA_OpenSessionEntryPoint(uint32_t nParamTypes, TEE_Param pParams[4],
-				    void **ppSessionContext)
+TEE_Result TA_OpenSessionEntryPoint(uint32_t param_types, TEE_Param params[4],
+				    void **psession)
```
jbech-linaro: Since you are renaming this, why not keep the names as in the GP spec?
```c
TEE_Result TA_EXPORT TA_OpenSessionEntryPoint(
   uint32_t paramTypes,
   [inout] TEE_Param params[4],
   [out][ctx] void**sessionContext );
```

ta/sims/ta_sims.c
```diff
@@ -212,3 +212,32 @@ TEE_Result sims_entry_panic(void *session_context, uint32_t param_types,
...
+	struct sims_session *loc_ctx = NULL;
+	TEE_Result res = TEE_ERROR_GENERIC;
+	uint32_t loc_ptypes = 0;
+	TEE_Param loc_params[4] = { };
+
+	res = open_client_session(&loc_ctx);
+	if (res != TEE_SUCCESS)
+		return res;
+
+	if (param_types ==
```
jbech-linaro: If it is not GitHub fooling me here, then the indentation line 228-231 makes it pretty hard to see what is what.

ta/sims/ta_entry.c
etienne-lms: camel case I guess... You prefer the GPCamelCaseWay?

ta/sims/ta_sims.c
etienne-lms: :)

ta/sims/ta_entry.c
jbech-linaro: No, since this is a file implemented by us, I'd go for our coding style, but use the same names as in the GP spec. i.e. like this, `param_types` , `params` and `session_context`.

ta/sims/include/ta_sims_test.h
etienne-lms: done

jbech-linaro:

etienne-lms:
tag applied on the 3 commits.

jenswi-linaro:
Since the fix isn't merged yet, why doesn't the added test fail?

jenswi-linaro:
The test case 1021 is broken, see the line https://github.com/OP-TEE/optee_test/pull/347/files#diff-064dfb0b0eea45b602d79c034f07893cR1575 where the greater part of the test case is skipped.

jforissier:
> The test case 1021 is broken [...]

Right, and the problem is, fixing the broken line results in failures at https://github.com/OP-TEE/optee_test/pull/347/files#diff-064dfb0b0eea45b602d79c034f07893cR1583.

@omihalac, could you please give a look?

omihalac:
@jforissier 

Sure. It will take some time to track the entire use case.

jforissier:
@omihalac no worries. Thanks.

omihalac:
@jforissier 

I apologize for the late response. Something very urgent has come up and I could focus on this case only this week.

Unfortunately, I encountered big issues configuring my virtual machine with `pyelftools` requested by lately OPTEE-OS build. I end up with a fresh install of an Ubuntu virtual machine and now I'm able to run `optee_test` suite for QEMU target on the latest baseline.

What I know so far is that this PR cannot reproduce the issue against OPTEE-OS version when [PR#2993](https://github.com/OP-TEE/optee_os/pull/2993) was merged.

jforissier:
@omihalac no need to apologize, we do understand that for many people, pushing code here is not their highest priority task and it's fine. All contributions are appreciated.

Now to clarify the issue: see https://github.com/OP-TEE/optee_test/pull/347#issuecomment-500735819. More precisely if we do this:
```diff
diff --git a/host/xtest/regression_1000.c b/host/xtest/regression_1000.c
index f87a4c9..e22504a 100644
--- a/host/xtest/regression_1000.c
+++ b/host/xtest/regression_1000.c
@@ -1561,7 +1561,7 @@ static void test_panic_ca_to_ta(ADBG_Case_t *c, const TEEC_UUID *uuid,
                goto bail1;
 
        exp_counter = multi_instance ? 0 : 1;
-       if (ADBG_EXPECT(c, exp_counter, counter))
+       if (!ADBG_EXPECT(c, exp_counter, counter))
                goto bail1;
 
        if (!ADBG_EXPECT_TEEC_RESULT(c, TEEC_ERROR_TARGET_DEAD,
```

...which seems to be the intent of the test, then it fails.

omihalac:
@jforissier

Seems that was a problem of understanding from my side. Initially, I thought that the test failure has as result the system hanging as it’s described in [#PR3067]( https://github.com/OP-TEE/optee_os/pull/3067), but it ends by revealing another issue in the test system introduced by [PR#355](https://github.com/OP-TEE/optee_test/pull/335). So, right now the focus has changed to fixing the error in the test system, I suppose.

In my assumption, the patch above https://github.com/OP-TEE/optee_test/pull/347#issuecomment-520841086 basically fixes an error in the test case avoiding in that way skipping the execution of almost half of the test. I don’t really know how this was happened, but I will work closely to have this fix as soon as possible.

Once, I applied the fix I was able to reproduce the regression 1021 failure and investigate the route cause. Mainly, the error is triggered at the line https://github.com/OP-TEE/optee_test/pull/347/files#diff-064dfb0b0eea45b602d79c034f07893cL1579 (this is a bit different than the line where your test is failing as detailed in https://github.com/OP-TEE/optee_test/pull/347#issuecomment-518672182) when an attempt to open a new session on a panicked context should return `TEEC_ERROR_TARGET_DEAD`. This specific call returns `TEEC_SUCCESS` instead. 

This brings us to a more concept related discussion. As we agreed to implement such situation when a panic occurs, once a panic is detected, all references (CA and TA sessions) to the context are removed (context pointer/indicator become `NULL`) and the context is destroyed. No other session linked to the context would be closed, this will have to be done by the user.

In my opinion the test raises a false positive situation, because opening any other session after a panic will succeed afterwards due to the complete removing of any panic context references. By the time a new session is about to be opened the core detects no executable linked to UUID and allocate new resources for that one.

Below I will try to detail what is happening in the scenario above and kindly ask you to correct me if I’m wrong.

    tee_ta_open_session() looking for TA UUID
       |
       |--->tee_ta_init_session()
       |--->tee_ta_context_find()  returns no context found
       |--->tee_ta_init_pseudo_ta_session() looking for pseudo TA UUID,
                                              returns `TEEC_ERROR_ITEM_NOT_FOUND`
       |--->tee_ta_init_user_ta_session() looking for user TA UUID
                    |--->load_elf() new context is created and TA application is loaded
The first new session is created and the core returns `TEEC_SUCCESS`.

The second call of `tee_ta_open_session()` is like:
    
    tee_ta_open_session() looking for TA UUID
        |
        |--->tee_ta_init_session()
        |--->tee_ta_context_find()`  context is found
        |--->tee_ta_init_session_with_context()` TA application is re-opened

The second session is created and the core returns `TEEC_SUCCESS`.

In case the panicked is triggered AFTER the session was already created we may have:

    tee_ta_invoke_command() ---> PANIC
       |
       |--->destroy_ta_ctx_from_session() mark all session linked to the context and
                                        destroy the context

Returns `TEE_ERROR_TARGET_DEAD`


As I mentioned at the beginning, any invoke command (in the tests we have `get_counter()`) to the sessions will return `TEE_ERROR_TARGET_DEAD`, but in case the user is trying to open a new session this would be considered a brand new application and the core will create a new context returning `TEEC_SUCCESS`.

This approach either revealed a limitation or the correct behavior of how the system should react is emphasized by the new test cases introduced by the current PR. Those are validating the use cases when a panic is triggered at the time a new session is about to be opened. 
Below is the call stack which is identical when creating the first session, but would be slightly different when creating the second session under which context the panic will be triggered:

    tee_ta_open_session()`looking for TA UUID
       |
       |--->tee_ta_init_session()
       |--->tee_ta_context_find()  context is found
       |--->tee_ta_init_session_with_context() TA application is re-opened
       |--->enter_open_session() ---> PANIC
       |--->tee_ta_close_session() panic ctx detected
                  |--->destroy_session() session destroyed
                  |--->destroy_context() context destroyed only if `ref_count == 0`

Any further attempt to open a new session on panic context would behave like:

    tee_ta_open_session ()
        |--->destroy_ta_ctx_from_session()` mark all session linked to the context and
                                               destroy the context
and returns `TEE_ERROR_TARGET_DEAD`


In conclusion we have two system behaviors reacting in different ways and I kindly ask you to review the flows before we are going further with the fix. I strongly believe that we have to unified the implementation for handling panic scenarios in TA manager as well.

omihalac:
@jforissier

Did you have by chance some time to review my comments? I thing that we have to agree on the overall system behavior before we would proceed any further. Thank you.

jforissier:
@omihalac 

> Did you have by chance some time to review my comments?

Yes, thanks for the detailed analysis.

> So, right now the focus has changed to fixing the error in the test system, I suppose.

Yes.

> In my opinion the test raises a false positive situation,

I tend to agree. Once the context is destroyed, the TA instance doesn't exist anymore, and I see no reason why new sessions could not be created on a new instance.
Would you mind proposing a change in `xtest` to remove the incorrect checks?

omihalac:
@jforissier

Sorry again for the late response, as always something very urgent come up. 
Below are some code changes that beside enabling execution of the bottom half of the test are also contributing to the test stabilization.

``` diff
diff --git a/host/xtest/regression_1000.c b/host/xtest/regression_1000.c
index 0544f06..294134f 100644
--- a/host/xtest/regression_1000.c
+++ b/host/xtest/regression_1000.c
@@ -1572,7 +1572,7 @@ static void test_panic_ca_to_ta(ADBG_Case_t *c, const TEEC_UUID *uuid,
 		goto bail1;
 
 	exp_counter = multi_instance ? 0 : 1;
-	if (ADBG_EXPECT(c, exp_counter, counter))
+	if (!ADBG_EXPECT(c, exp_counter, counter))
 		goto bail1;
 
 	if (!ADBG_EXPECT_TEEC_RESULT(c, TEEC_ERROR_TARGET_DEAD,
@@ -1589,7 +1589,7 @@ static void test_panic_ca_to_ta(ADBG_Case_t *c, const TEEC_UUID *uuid,
 		goto bail1;
 
 	/* Attempt to open a session on panicked context */
-	if (!ADBG_EXPECT_TEEC_RESULT(c, TEEC_ERROR_TARGET_DEAD,
 			xtest_teec_open_session(&cs[1], uuid, NULL,
 						&ret_orig)))
 		goto bail1;
@@ -1613,7 +1613,8 @@ static void test_panic_ca_to_ta(ADBG_Case_t *c, const TEEC_UUID *uuid,
 	if (!ADBG_EXPECT_TEEC_SUCCESS(c, sims_get_counter(&cs[2], &counter)))
 		goto bail2;
 
-	if (!ADBG_EXPECT(c, 0, counter))
+	exp_counter = multi_instance ? 0 : 1;
+	if (!ADBG_EXPECT(c, exp_counter, counter))
 		goto bail2;
 
 	/*
@@ -1636,9 +1637,9 @@ static void test_panic_ca_to_ta(ADBG_Case_t *c, const TEEC_UUID *uuid,
 		goto bail0;
 
 	exp_res = multi_instance ? TEEC_SUCCESS : TEEC_ERROR_TARGET_DEAD;
-	if (!ADBG_EXPECT_TEEC_RESULT(c, exp_res,
-				     sims_get_counter(&cs[0], &counter)))
-		goto bail1;
+	ADBG_EXPECT_TEEC_RESULT(c, exp_res, sims_get_counter(&cs[0], &counter));
+
+	goto bail0;
 
 bail2:
 	TEEC_CloseSession(&cs[2]);

```

Looking forward to your review.

jforissier:
Hi @omihalac your patch does not apply to master. How about you create a pull request?

omihalac:
@jforissier 

Thank you for pointing that out. The patch was intended to be used on different branch and it makes sense to isolate the issues. Once [PR#366](https://github.com/OP-TEE/optee_test/pull/366) reaches a feasible state and would be eventually merged, we can focus on the problem raised in this PR.

host/xtest/regression_1000.c
```diff
@@ -1603,6 +1616,30 @@ static void test_panic_ca_to_ta(ADBG_Case_t *c, const TEEC_UUID *uuid,
...
+		return;
+
+	if (!ADBG_EXPECT_TEEC_RESULT(c, TEEC_ERROR_TARGET_DEAD,
+			sims_panic_in_open_session(&cs[1], uuid, &ret_orig)))
+		goto bail0;
+
+	exp_res = multi_instance ? TEEC_SUCCESS : TEEC_ERROR_TARGET_DEAD;
+	if (!ADBG_EXPECT_TEEC_RESULT(c, exp_res,
+				     sims_get_counter(&cs[0], &counter)))
+		goto bail1;
```
omihalac: The following piece of code should be place instead in order to avoid `Segmentation Fault` error in test execution. 
``` 
    ADBG_EXPECT_TEEC_RESULT(c, exp_res, sims_get_counter(&cs[0], &counter));
    goto bail0;
```

ta/sims/include/ta_sims_test.h
```diff
@@ -44,10 +44,46 @@
...
+ * Optional parameters:
+ *
+ * param#0: [in]	memref = UUID of a TA
+ *			If provided, open a session towards this TA and
+ *			invoke it with command TA_SIMS_CMD_PANIC.
+ */
+#define TA_SIMS_CMD_PANIC		4
+
+/*
+ * TA_SIMS_OPEN_TA_SESSION	Open a session towards a TA
```
omihalac: Also update the entry name in the comment to `TA_SIMS_CMD_OPEN_TA_SESSION`

PR #348, opened 2019-05-27
os_test: enable __noprof attribute for call_longjmp()

b49020:
longjmp() api called from call_longjmp() intentionally corrupts/discards
stack frame of call_longjmp() api to make a long jump to a location
configured via setjmp(). As most of the profilers (esp. function call
graph) relies on correct stack frames to profile trusted applications
which makes this api non-profilable. So lets enable __noprof attribute
for call_longjmp() api.

Signed-off-by: Sumit Garg <sumit.garg@linaro.org>

### DIFF
--- a/ta/os_test/os_test.c
+++ b/ta/os_test/os_test.c
@@ -660,7 +660,7 @@ static TEE_Result test_float(void)
 }
 #endif /*CFG_TA_FLOAT_SUPPORT*/
 
-static __noinline void call_longjmp(jmp_buf env)
+static __noinline __noprof void call_longjmp(jmp_buf env)
 {
 	DMSG("Calling longjmp");
 	longjmp(env, 1);


b49020:
This patch fixes issue reported in ftrace PR: https://github.com/OP-TEE/optee_os/pull/2912#issuecomment-495712979

jforissier:
I don't think this is the right fix, because it is addressing the symptom not the cause. IMO, and as I said already (https://github.com/OP-TEE/optee_os/pull/2912#issuecomment-491253278), it is not reasonable to ask the users of `longjmp()` to add `__noprof` to their code. Without `__noprof` the app crashes in a weird way and the user may have a hard time figuring out how to fix it :(

jenswi-linaro:
With this you're just brushing the problem under the rug.
There has to be a way where ftrace and `setjmp()` can work together.

b49020:
Continuing from this comment: https://github.com/OP-TEE/optee_os/pull/2912#issuecomment-492237457 as we haven't reached a conclusion there for alternative fix rather than the fix in this PR which seems reasonable to me.

I have tried to create an actual visualisation of `setjmp()/longjmp()` function call chain in `xtest 1006` as follows:

```
test_setjmp () {
  setjmp ();           <<======||
  call_longjmp () {            ||                << Discarded stack for this api.
      longjmp () {     ========||
}
```

I am not sure how you would like to represent that (inconsistent function chain) in ftrace output.

> There has to be a way where ftrace and setjmp() can work together.

We can implement a hack for both of them to work together but output ftrace graph won't represent the above inconsistent function call chain.

> it is not reasonable to ask the users of longjmp() to add __noprof to their code. Without __noprof the app crashes in a weird way and the user may have a hard time figuring out how to fix it :(

We can clearly document `setjmp()/longjmp()` as special api's that corrupts stack frames and changes normal execution flow via changing stack pointer in assembly [1]. So we can't trace them using ftrace as they are designed to trace normal C function calls.

Also, it is something like user implements an assembly stub that corrupts stack frames in one way or another and expects ftrace (dependent on stack frames) to play nicely.

[1] https://github.com/OP-TEE/optee_os/blob/0e3c17074dc0812412a1445c021ed26cece45b22/lib/libutils/isoc/arch/arm/setjmp_a64.S#L70

jenswi-linaro:
I assume you know what `setjmp()` and `longjpmp()` is supposed to do. If you can provide another implementation that works together with ftrace we'll use that instead.

b49020:
> If you can provide another implementation that works together with ftrace we'll use that instead.

IIUC the use-case, I would suggest to simply remove call to `call_longjmp()` as follows and do a long jump inside single function only (`test_setjmp()`) to a particular address configured by `setjmp()`. This would avoid corruption of stack frames and works well with ftrace too.

```
diff --git a/ta/os_test/os_test.c b/ta/os_test/os_test.c
index feca6b4..28be2d2 100644
--- a/ta/os_test/os_test.c
+++ b/ta/os_test/os_test.c
@@ -660,13 +660,6 @@ static TEE_Result test_float(void)
 }
 #endif /*CFG_TA_FLOAT_SUPPORT*/
 
-static __noinline void call_longjmp(jmp_buf env)
-{
-       DMSG("Calling longjmp");
-       longjmp(env, 1);
-       EMSG("error: longjmp returned to calling function");
-}
-
 static TEE_Result test_setjmp(void)
 {
        jmp_buf env = { };
@@ -675,7 +668,9 @@ static TEE_Result test_setjmp(void)
                IMSG("Returned via longjmp");
                return TEE_SUCCESS;
        } else {
-               call_longjmp(env);
+               DMSG("Calling longjmp");
+               longjmp(env, 1);
+               EMSG("error: longjmp returned to calling location");
                return TEE_ERROR_GENERIC;
        }
 }
```

jenswi-linaro:
This is just a sanity test, the functions are supposed to work as described at http://man7.org/linux/man-pages/man3/setjmp.3.html

b49020:
Thanks @jenswi-linaro for this reference. It pretty clear that purpose of these api's is to perform a nonlocal goto. And it looks like Linux doesn't use these as in-kernel apis, so Linux ftrace doesn't need to handle these. But here we are implementing ftrace for OP-TEE user-space which provide these apis, so we need to handle them in a special way.

How about if function graph looks as follows?

```
test_setjmp () {
  setjmp ();
  call_longjmp1 () {
      longjmp () {
  call_longjmp2 () {
      longjmp () {
  call_longjmp3 () {
      longjmp () {
}
```

jenswi-linaro:
The unbalanced `{` and `}` are a bit disturbing. How about:
```
test_setjmp () {
  setjmp ();
  call_longjmp () {
      longjmp () {
      } /* fabricated return from longjmp() */
   } /* fabricated return from call_longjmp() */
}
```

b49020:
Ok fine then will use these fabricated returns although they might confuse the user but these seem to be one option that we all agree upon.

b49020:
Closing this PR.

PR #350, opened 2019-05-31
regression 6020:fix a session not closed

sticfeynman:
Resource leak in core can be observed when this test case is repeated in many
times.

Signed-off-by: Oliver Chiang <<sticfeynman@gmail.com>>

### DIFF
--- a/host/xtest/regression_6000.c
+++ b/host/xtest/regression_6000.c
@@ -2220,7 +2220,7 @@ static void xtest_tee_test_6020_single(ADBG_Case_t *c, uint32_t storage_id)
 				storage_id);
 
 	ADBG_EXPECT_TEEC_RESULT(c, TEE_ERROR_TARGET_DEAD, res);
-
+	TEEC_CloseSession(&sess);
 	return;
 exit1:
 	ADBG_EXPECT_TEEC_SUCCESS(c, fs_unlink(&sess, obj));


jforissier:
Hi @sticfeynman ,

The fix LGTM, but, what do you mean "when this test case is repeated many times"'?
Did you add a loop in the code or run the xtest command several times?
If it is the latter, then I think there is a weakness in the TEE driver. IMO it should close the sessions automatically when the client process exits, in the same way that file descriptors are closed.

sticfeynman:
I do put a loop to run the test and this helps to find tricky bugs in our port sometimes, and yes, all sessions are closed automatically when the client leaves.
I can also observe a OOM error is thrown from the TEE driver when eventually a OPTEE_MSG_RPC_CMD_SHM_ALLOC is requested.
Can you briefly describe the weakness? Then maybe we might figure out a  reasonable test strategy.
Or do you have any suggestion to patch the TEE driver to get around it?

jforissier:
> all sessions are closed automatically when the client leaves.

OK, good, then no weakness in the driver. I was imagining a problem that doesn't exist ;-)

jbech-linaro:

jenswi-linaro:
This fix is useful. Please remove the merge commit and I think we're done.

jforissier:
Updated and merged manually.

PR #352, opened 2019-06-27
regression_8100: use null terminated strings with file_to_c

ricardosalveti:
GCC 9 is more strict with string manipulation, causing the build to
fail as the string data converted via file_to_c is not null terminated,
as described by the following build error:

regression_8100.c:100:29: error: '%*s' directive argument is not a
nul-terminated string [-Werror=format-overflow=]
tlen = myasprintf(&trust, "%*s", (int)sizeof(regression_8100_ca_crt),
                           ^~~
      regression_8100_ca_crt);
      ~~~~~~~~~~~~~~~~~~~~~~

Change file_to_c to terminate the string after conversion and update the
string size to remove the null terminated byte. Also update
regression_8100 to use the size variable defined via file_to_c instead
of manually calling sizeof.

Signed-off-by: Ricardo Salveti <ricardo@foundries.io>

### DIFF
--- a/host/xtest/regression_8100.c
+++ b/host/xtest/regression_8100.c
@@ -91,13 +91,13 @@ static void test_8102(ADBG_Case_t *c)
 		return;
 
 	clen = myasprintf(&chain, "%*s\n%*s",
-			  (int)sizeof(regression_8100_my_crt),
+			  (int)regression_8100_my_crt_size,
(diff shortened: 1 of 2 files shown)


jenswi-linaro:

jforissier:
Merged, thanks.

PR #354, opened 2019-06-28
regression: fix memory leakage in case 1010

jenswi-linaro:
In regression test case 1010 xtest_tee_test_invalid_mem_access2()
allocates shared memory, but doesn't free it. This can affect other test
cases run after this case.

The shared memory is freed when the context is finalized, but that can
be too late since it normally happens when xtest exits.


jforissier:

jenswi-linaro:
Tag applied

PR #357, opened 2019-08-02
Update file_to_c.py to be Python3 compatible

vivinamartins16:
Builds with Python3 package in the common.mk file, would trigger an error in this file becasue older version of python handles arrays differently. This update enables python3 users to build with python3 package without errors. In a few months, Python2 will no longer be supported so I believe this change can be helpful.

### DIFF
--- a/scripts/file_to_c.py
+++ b/scripts/file_to_c.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 # SPDX-License-Identifier: BSD-2-Clause
 #
 # Copyright (c) 2018, Linaro Limited
--- a/scripts/file_to_c.py
+++ b/scripts/file_to_c.py
@@ -3,6 +3,8 @@
 #
 # Copyright (c) 2018, Linaro Limited
 #
+from array import array
+
 
 def get_args():
     import argparse
@@ -16,7 +18,6 @@ def get_args():
     return parser.parse_args()
 
 def main():
-    import array
 
     args = get_args()
 
@@ -26,11 +27,15 @@ def main():
     f.write("/* automatically generated */\n")
     f.write("#include <stdint.h>\n")
     f.write("#include <stddef.h>\n\n")
-
+    
     f.write("const uint8_t " + args.name + "[] = {\n")
+    
+    wordarray = inf.read()
+    b = bytearray(wordarray, 'utf-8')
+    
     i = 0
-    for x in array.array("B", inf.read()):
-        f.write("0x" + '{0:02x}'.format(x) + ",")
+    for x in b:
+        f.write("0x" + "{0:02x}".format(x) + ",")
         i = i + 1
         if i % 8 == 0:
             f.write("\n")


scripts/file_to_c.py
```diff
@@ -3,6 +3,8 @@
 #
 # Copyright (c) 2018, Linaro Limited
 #
+from array import array
+
```
jforissier: Why not `#!/bin/env python3`?

scripts/file_to_c.py
```diff
@@ -26,11 +27,15 @@ def main():
     f.write("/* automatically generated */\n")
     f.write("#include <stdint.h>\n")
     f.write("#include <stddef.h>\n\n")
-
+    
     f.write("const uint8_t " + args.name + "[] = {\n")
+    
+    wordarray = inf.read()
+    b = bytearray(wordarray, 'utf-8')
```
jforissier: Why `utf-8`? Shouldn't the file be open in binary mode above?

scripts/file_to_c.py
jbech-linaro: I guess there are pros and cons with that. If the script work with both Python 2 and 3, then `#!/usr/bin/env python` is better, since it'll work with whatever Python environment that is default on the users machine. However, since we intentionally are aiming for moving all our Python scripts to Python 3 (due to the Python2 deprecation), I think I'm with you @jforissier , better to use what you've proposed.

vivinamartins16: Thanks! Agreed :-) I'll fix it now! Should I create a new commit or just amend this one?

jbech-linaro: > Thanks! Agreed :-) I'll fix it now! Should I create a new commit or just amend this one?

We usually just add a new "fixup" commit to the branch, then later on when where all happy, then we squash those into a single (or more relevant) commits.

scripts/file_to_c.py
vivinamartins16: Yes, I chose utf-8 just to be safe because I wasn't sure about the usage of the script (if it was only in regression tests or anywhere else). But if the only use case is base64, than I agree and we can skip it!

jbech-linaro:
@vivinamartins16 we should pick up this in a week or two after the release. Having said that, there ring a bell that someone made a PR converting most (all?) our Python scripts to Python 3. If that is true, then this PR might not be necessary.

jbech-linaro:
> Are you referring to [OP-TEE/optee_os#3187](https://github.com/OP-TEE/optee_os/pull/3187)? It was optee_os only, I can't find a PR in optee_test other than this one.

Maybe that's the one I had in mind, but in that case this is still a useful PR.

wamserma:
> > Are you referring to [OP-TEE/optee_os#3187](https://github.com/OP-TEE/optee_os/pull/3187)? It was optee_os only, I can't find a PR in optee_test other than this one.
> 
> Maybe that's the one I had in mind, but in that case this is still a useful PR.

Yes, I think so. To clarify: On Python 2.7 the read() returned an ASCII string, which is taken as a bytestring by the script and directly hex-encoded into the output.
With Python 3, the read() returns a unicode string, which cannot be fed to a byte array, hence the error. Simply replacing `inf = open(args.inf, 'r')`with `inf = open(args.inf, 'rb')` fixes the issue without introducing new assumptions on the input. (If there was any unicode in the input, the resulting C-arrays would probably be unusable.)
So this should be a one-line fix without any extra conversions. (Aside from changing "w" to "wt" in the subsequent line, if you like.)

wamserma:
Just to keep track of the corresponding issue: Once merged, this closes https://github.com/OP-TEE/optee_test/pull/357

sbranden:
Could we get something merged to support python3?

jforissier:
@vivinamartins16 I forgot to mention: we need a `Signed-off-by:`.

sbranden:
How about this?
https://github.com/OP-TEE/optee_test/pull/383/commits

jforissier:
> How about this?
> https://github.com/OP-TEE/optee_test/pull/383/commits

Looks more "Python-esque", if you say it works I believe you and I will ack the patches :)

sbranden:
> > How about this?
> > https://github.com/OP-TEE/optee_test/pull/383/commits
> 
> Looks more "Python-esque", if you say it works I believe you and I will ack the patches :)

Please don't believe me.  I'm not a python expert either.  Please run it through your test suite to make sure it works.

wamserma:
Solved by https://github.com/OP-TEE/optee_test/pull/383, can be closed.

PR #358, opened 2019-08-06
Add regression 1022 to test dlopen()/dlsym()/dlclose()

jforissier:
Adds a new test case for the dynamic link API in libdl.
- A new shared library is added in ta/os_test_lib_dl.
- In xtest 1022, the dl API is used to load the test library and
  retrieve pointers to functions.

Signed-off-by: Jerome Forissier <jerome.forissier@linaro.org>

PR #359, opened 2019-08-06
Android.mk: Add missing include dirs

vchong:


vchong:
#358 should be merged first.

jforissier:
Does this fix a build on AOSP? If so, how come we meet the issue only now? (some paths have been existing for quite some time).

vchong:
I've the same question myself. Testing right now.

vchong:
Ok, the TA include dirs are only needed if one of the header files is used by `host/xtest/*`, so I'll close this. Sorry for the noise.

jforissier:
Ok, so now I'm wondering, perhaps some of those paths are not needed and could be removed?

vchong:
> Ok, so now I'm wondering, perhaps some of those paths are not needed and could be removed?

Nope. Already checked. All existing ones are used. ; )

jforissier:
Excellent :)  Thanks!

Issue #363, opened 2019-09-10
Test 4009 uses invalid key pair for ECDSA operation


c-auger:
In test 4009, some implementations might want to check the provided keypair, and fail while doing so.

>		/*
>		 * The public value is not used. This is why we provide
>		 * another buffer
>		 */
>		xtest_add_attr(&param_count, params,
>				TEE_ATTR_ECC_PUBLIC_VALUE_X,
>			       pt->private, size_bytes);
>		xtest_add_attr(&param_count, params,
>				TEE_ATTR_ECC_PUBLIC_VALUE_Y,
>			       pt->private, size_bytes);

Public_x and Public_y are not allowed to be random values for a usable key.
Thus if the cryptographic library in use perform some validation test, then it will fail and the key creation might be rejected.

According to specification of TEE_PopulateTransientObject,

> The Implementation can attempt to detect whether the attribute values are consistent;
> for example, if the numbers supposed to be prime are indeed prime.
> However, it is not required to do these checks fully and reliably.
> If it detects invalid attributes, it SHALL return the error code TEE_ERROR_BAD_PARAMETERS and SHALL NOT panic.
> If it does not detect any inconsistencies, it SHALL be able to later proceed with all operations associated with the object without error.
> In this case, it is not required to make sensible computations, but all computations SHALL terminate and output some result.

Thus, I would tend to think that using legitimate values for public_x and public_y should be better for a non regression test. Such x and y can be obtained simply from `pt->public_x` and `pt->public_y`. If other values than this x and this y should be tested, then it should be possible to generate an other key and extract x and y from it.

jenswi-linaro:
A pull request is welcome! :-)

c-auger:
Fixed in P-R #365

PR #365, opened 2019-09-11
Set use of a valid public point for ECDSA operation in test 4009

c-auger:
As described in issue #363, this patch is to support a valid key rather than providing incorrect public part.

The comment states something true, X and Y are not used (otherwise the test would fail), but some implementations might check the validity of the point in order to accept populating the key.

### DIFF
--- a/host/xtest/regression_4000.c
+++ b/host/xtest/regression_4000.c
@@ -4860,15 +4860,16 @@ static void xtest_tee_test_4009(ADBG_Case_t *c)
 			       TEE_ATTR_ECC_PRIVATE_VALUE,
 			       pt->private, size_bytes);
 		/*
-		 * The public value is not used. This is why we provide
-		 * another buffer
+		 * The public value is not used, but we should provide a valid
+		 * one to avoid rejection in case TEE_PopulateTransientObject()
+		 * checks for key validity.
 		 */
 		xtest_add_attr(&param_count, params,
-				TEE_ATTR_ECC_PUBLIC_VALUE_X,
-			       pt->private, size_bytes);
+			       TEE_ATTR_ECC_PUBLIC_VALUE_X,
+			       pt->public_x, size_bytes);
 		xtest_add_attr(&param_count, params,
-				TEE_ATTR_ECC_PUBLIC_VALUE_Y,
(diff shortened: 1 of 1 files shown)


jforissier:
Please add a commit description (your last sentence above is fine: "The comment states ..."), and add:

jforissier:
Forgot to say: you need to provide a `Signed-off-by:`, too.

c-auger:
Thanks this is my very first pull-request on GitHub. I will do it.

jforissier:
No worries, welcome and thanks for contributing :+1:

c-auger:
Did I amend it correctly?

jforissier:
> Did I amend it correctly?

Yes. "Description:" is useless but it doesn't matter.

jforissier:
Last thing: it is recommended to wrap text at 75 characters.

c-auger:
Should I reamend the patch to fit 75 columns or is it okay?

jforissier:
You add tags once the review is done. You never apply a tag unless the person has posted it. Therefore, which tags are added will depend on who has reviewed the patch.
You may amend the patch as soon as a new patch is posted, or wait a bit to get more tags (for more complex patches usually).

See https://optee.readthedocs.io/general/contribute.html.

> Should I reamend the patch to fit 75 columns or is it okay?

Please update it, if you don't mind. It's definitely no big deal but I like commits to be as "clean" as possible since they will stick forever.

host/xtest/regression_4000.c
```diff
@@ -4860,15 +4860,16 @@ static void xtest_tee_test_4009(ADBG_Case_t *c)
 			       TEE_ATTR_ECC_PRIVATE_VALUE,
 			       pt->private, size_bytes);
 		/*
-		 * The public value is not used. This is why we provide
-		 * another buffer
+		 * The public value is not used, but we should provide a valid
+		 * one to avoid rejection in case TEE_PopulateTransientObject
```
etienne-lms: suggest to add parentheses to the function label in  the message: `TEE_PopulateTransientObject()`.

c-auger:
Should I replace

with

`Reviewed-by: Etienne Carriere <...>`

in the commit?

Issue #362, opened 2019-09-10
Optee test 1011 uses a 0-second timeout while INFINITE_TIMEOUT should probably be better suited.


c-auger:
Hi, in `ta/rpc_test/ta_rpc.c`, timeouts provided to TEE_OpenTASession and other TA entrypoints are 0 milliseconds timeout.

According to the GP specification, this could end up with a TEE_ERROR_CANCEL returned, and the test expects TEE_SUCCESS.

I think, that it should be safer to replace these invocations with TEE_TIMEOUT_INFINITE as timeouts.

See section 4.10 of TEE Internal Core API Specification:

If the operation has not reached the TA yet but has been queued in the TEE, then it SHALL be retired from the queue and fail with the return code:
o For an REE Client, TEEC_ERROR_CANCEL and the origin TEEC_ORIGIN_TEE;
o For a TEE Client, TEE_ERROR_CANCEL and the origin TEE_ORIGIN_TEE.

0 is no special value. TEE_TIMEOUT_INFINITE is specified as being 0xFFFFFFFF.

Thus a conformant application should probably fail in this test.

An easy fix would be to replace all 0 timeouts with TEE_TIMEOUT_INFINITE.

jenswi-linaro:
I agree. Would you mind creating a pull request?

c-auger:
Fixed in pull request #367.

PR #367, opened 2019-09-11
Use infinite timeouts in ta_rpc instead of 0ms timeouts

c-auger:
As indicated in #362, a 0 seconds timeout should likely lead to a `TEE_ERROR_CANCEL` error. This is the case in particular if the command has been queued in the TEE but did not yet reached the trusted application.

Setting the timeout to `TEE_TIMEOUT_INFINITE` should solve the issue whether or not the command is queued.


etienne-lms:
All fine, yet there are other occurrences of calls to `TEE_OpenTASession()` and `TEE_InvokeTACommand()` in **optee_test** that gets a null timeout argument. 
Could you fix them also?

jforissier:
+1 with @etienne-lms, I think it makes sense to fix all instances in the same commit.

c-auger:
Ok, I am on it. I spotted this place as I had the error on my implementation, but I may have skipped some tests due to not implemented features. I will have to check if other do not play with timeouts in a meaningful way (ie. for which using TEE_INFINITE_TIMEOUT will break its purpose).

c-auger:
There was only one place that I spotted where an explicit timeout was provided.
This place is in the `ta_entry_client_with_timeout()` function of file `ta/os_test/os_test.c`, it was for a `TEE_InvokeTACommand()`.
I did not see any equivalent test for `TEE_OpenTASession()`.

jforissier:
LGTM. I suggest "Remove **0ms** timeouts...".

etienne-lms:
I checked the other optee_* and found occurrences in optee_os only. @c-auger, you want to create a P-R for it?
```Make
optee_os/lib/libutee/tee_socket_pta.c:	return TEE_InvokeTACommand(sess, 0, cmd_id, param_types, params, NULL);
optee_os/lib/libutee/arch/arm/gprof/gprof_pta.c:	res = TEE_InvokeTACommand(sess, 0, cmd_id, param_types, params, NULL);
optee_os/lib/libutee/tee_system_pta.c:	return TEE_InvokeTACommand(sess, 0, cmd_id, param_types, params, NULL);

PR #368, opened 2019-09-11
TEE_CreatePersistentObject is now called with a valid pointer.

c-auger:
According to GP specification, passing a NULL pointer to
TEE_{Open,Create}PersistentObject should panic the trusted application.
Previous code was using a NULL pointer and expected success.
After creating (and opening as stated by the specification) the new object,
it is also closed, as in anyway we could not expect to do anything on it
previously.

Signed-off-by: Cedric Auger <cedric.auger@provenrun.com>


c-auger:
> `Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` although local variable `o` could be rename `obj` or `object_hdl`.

For `o`, I was just following surrounding code convention. `ta_storage_cmd_read()` and `ta_storage_cmd_create()` used this name as local variable too. In fact, only `ta_storage_cmd_loop()` uses a more meaningful name in this file.

jforissier:
Missing parentheses in subject, you did not copy verbatim what I proposed :-/

`Call TEE_CreatePersistentObject() with a valid pointer`

With that fixed:

c-auger:
> Missing parentheses in subject, you did not copy verbatim what I proposed :-/
> 
> `Call TEE_CreatePersistentObject() with a valid pointer`
> 
> With that fixed:

My bad, I changed to the imperative style but forgot the parenthesis. I will fix it.

PR #371, opened 2019-09-19
Clang support

jforissier:
Fixing two issues that make regression 1010 fail with Clang (tested on HiKey960).


jenswi-linaro:

PR #372, opened 2019-10-09
xtest/Makefile: drop COMPILE_NS_USER and detect arch automatically

jforissier:
COMPILE_NS_USER is used by the xtest Makefile to link against either
the 32-bit or 64-bit version of OpenSSL (libcrypto.a). Historically,
the top level Makefile of the OP-TEE build project would take
COMPILE_NS_USER as an input, with value 32 or 64, and derive the path
to the appropriate cross-compiler. But given that the cross-compiler
may also be customized via CROSS_COMPILE/CROSS_COMPILE_HOST,
inconsistencies can happen.

There is no reason to keep COMPILE_NS_USER, since its value can be
derived from the compiler itself. This commit adds macros to do just
that and uses them to select the proper libcrypto library.

Reported-by: Denys Dmytriyenko <denys@ti.com>

### DIFF
--- a/host/xtest/Makefile
+++ b/host/xtest/Makefile
@@ -24,6 +24,15 @@ OBJCOPY		?= $(CROSS_COMPILE)objcopy
 OBJDUMP		?= $(CROSS_COMPILE)objdump
 READELF		?= $(CROSS_COMPILE)readelf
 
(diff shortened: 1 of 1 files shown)


glneo:
Worked for me,
```
Acked-by: Andrew F. Davis <afd@ti.com>
```

PR #373, opened 2019-10-15
build/Makefile: fix os_test with dynamic linking

wamserma:
When CFG_TA_DYNLINK is enabled, os_test depends on os_test_lib
to be build beforehand. Presence of the lib is ensured by executing
a proper submake.

Signed-off-by: Markus S. Wamser <markus.wamser@mixed-mode.de>

### DIFF
--- a/ta/os_test/Makefile
+++ b/ta/os_test/Makefile
@@ -7,9 +7,19 @@ ifeq ($(CFG_TA_DYNLINK),y)
 # follows the same scheme as the output path of the os_test application.
 # Therefore, simply replacing os_test by os_test_lib in the absolute path
 # should get us a valid path
-LDADD = -L$(subst os_test,os_test_lib,$(abspath $(link-out-dir))) -los_test
+OS_TEST_LIB_LDADD = -L$(subst os_test,os_test_lib,$(abspath $O)) -los_test
+OS_TEST_LIB = os_test_lib
 endif
 
 LDADD += -ldl
 
-include ../ta_common.mk
+# ensure os_test_lib is present
+.PHONY: os_test_lib
+os_test_lib:
+	@$(MAKE) -C $(subst os_test,os_test_lib,$(abspath $(@D))) all
+
+all: $(OS_TEST_LIB)
+	@LDADD="$(OS_TEST_LIB_LDADD) $(LDADD)" $(MAKE) -f ../ta_common.mk BINARY=$(BINARY) $@
+
+%:
+	@LDADD="$(OS_TEST_LIB_LDADD) $(LDADD)" $(MAKE) -f ../ta_common.mk BINARY=$(BINARY) $@


jforissier:
Do we really need this? In principle, yes, I agree. But the libraries and applications are supposed to be built in the correct order by the parent Makefile (see `TA_DIRS`), so I wonder if it's worth making things more complicated.

Or is there a scenario I am missing?

wamserma:
> Update: I see the original issue only now :-/
> 
> Is this only to be able to use `ninja`? Is the problem reproducible with GNU make (I don't think so)?

It's the other way round: It probably won't fix Ninja, but I hit this in when building `optee_test` through buildroot (as per the the Manifest from https://github.com/OP-TEE/build
But as this is a specialty of the `os_test` TA which already had a hack in its Makefile, I think this is the right place to fix this.

jforissier:
@wamserma OK, it's weird because I have never had any issue on my side (and I use `make -j10` always). OK let me review the patch now ;-)

ta/os_test/Makefile
```diff
@@ -2,14 +2,29 @@ include $(TA_DEV_KIT_DIR)/mk/conf.mk
 
 BINARY = 5b9e0e40-2636-11e1-ad9e-0002a5d5c51b
 
+OS_TEST_LIB = ""
+OS_TEST_LIB_LDADD = ""
```
jforissier: Not needed.

ta/os_test/Makefile
```diff
@@ -2,14 +2,29 @@ include $(TA_DEV_KIT_DIR)/mk/conf.mk
...
+
 ifeq ($(CFG_TA_DYNLINK),y)
 # Hack: to locate the os_test library, we assume that its output path
 # follows the same scheme as the output path of the os_test application.
 # Therefore, simply replacing os_test by os_test_lib in the absolute path
 # should get us a valid path
-LDADD = -L$(subst os_test,os_test_lib,$(abspath $(link-out-dir))) -los_test
+OS_TEST_LIB_LDADD = -L$(subst os_test,os_test_lib,$(abspath $O)) -los_test
```
jforissier: Why replace `$(link-out-dir)` by `$O`? I think it's wrong, because  `$(link-out-dir)`  should take a default value when `$O` is empty.
See https://github.com/OP-TEE/optee_test/blob/master/Makefile#L1-L9 and https://github.com/OP-TEE/optee_os/blob/master/ta/mk/ta_dev_kit.mk#L19. In fact to correct variable should be `$(out-dir)` IMO but it's not the purpose of this PR.

ta/os_test/Makefile
```diff
@@ -2,14 +2,29 @@ include $(TA_DEV_KIT_DIR)/mk/conf.mk
...
+OS_TEST_LIB = os_test_lib
 endif
 
 LDADD += -ldl
 
-include ../ta_common.mk
+# ensure os_test_lib is present
+.PHONY: os_test_lib
+os_test_lib:
+	@echo .. suspended while building $(subst os_test,os_test_lib,$(abspath $(@D)))/Makefile
```
jforissier: This line is not needed since make outputs messages already:
```
make[3]: Entering directory '<...>/optee_test/ta/os_test'
make[4]: Entering directory '<...>/optee_test/ta/os_test_lib'
[...]
make[4]: Entering directory '<...>/optee_test/ta/os_test_lib'
make[3]: Entering directory '<...>/optee_test/ta/os_test'

File: ta/os_test/Makefile [hunk 4]
```diff
@@ -2,14 +2,29 @@ include $(TA_DEV_KIT_DIR)/mk/conf.mk
...
+	@$(MAKE) -C $(subst os_test,os_test_lib,$(abspath $(@D))) all
+	@echo resuming ..
```
jforissier: Same for this line

ta/os_test/Makefile
wamserma: `link-out-dir`(or `out-dir`) is no longer available here. It was set by the previous `include ../ta_common.mk`, which now is called explicitly since we need to override the `all:` target.
There might be some escape-magic to pass `$(subst)` and `$(abspath)` in a way that evaluation is pushed down into the called sub-make, but that is beyond my skills.

jforissier:
Hmmm... Wait... We're trying to fix parallel builds, but what prevents two threads from trying to build `os_test_lib` simultaneously? Nothing I think.
IMO, this problem needs to be addressed one level up, i.e., in the buildroot makefile. Either create a dependency, or serialize the targets (as we do in the "regular" Makefile).

wamserma:
> Hmmm... Wait... We're trying to fix parallel builds, but what prevents two threads from trying to build `os_test_lib` simultaneously? 

If independent make runs are started: nothing. If they are launched as sub-makes, to my understanding, the jobs should be added to the same jobserver, so make should take care of that situation. Besides, the jobs are already serialized by the loop in the buildroot makefile. But there they jobs are generated by Glob-expansion, which means dependencies have to be resolved manually and encoded into directory names which always result into the same sequence after expansion, independent of locale etc. (prefixing a LANG=C is not enough).

To solve this cleanly, ta_dev_kit.mk needs to be extended with a method to pass dependencies/pre-requisites to the generic `all` target. Until then, this PS is the best we can do.

jforissier:
>  If they are launched as sub-makes, to my understanding, the jobs should be added to the same jobserver, so make should take care of that situation.

Correct.

I think I have a better idea. We have a makefile to build all the TAs already (`ta/Makefile`), Buildroot should use it instead of directly invoking `ta/*/Makefile`. This would solve the concurrency issue and also take care of another minor problem: Buildroot currently build `sdp_basic` even if `CFG_SECURE_DATA_PATH=n`.

Let me try that and I will post something.

jforissier:
https://github.com/OP-TEE/optee_test/pull/374 and https://github.com/OP-TEE/build/pull/394.

wamserma:
> #374 and [OP-TEE/build#394](https://github.com/OP-TEE/build/pull/394).

Tested against qemu-manifest 3.6.0. LGTM. (I had to remove `os_test_lib_dl` from the .gmake as this is not present in the 3.6.0 tests).

wamserma:
The issue targeted in this PR is (hopefully) set to be solved in https://github.com/OP-TEE/build/pull/353.

PR #374, opened 2019-10-16
ta: buildroot: preserve ta/Makefile as ta/Makefile.gmake

jforissier:
When using Buildroot to build optee_test, the file ta/CMakeList.txt is
converted into a Makefile which overwrites ta/Makefile. This is bad
because ta/Makefile contains configuration and dependency information,
such as: os_test_lib needs to be built before os_test, or sdp_basic
needs to be built only when CFG_SECURE_DATA_PATH=y, and so on. So,
instead of directly using ta/*/Makefile, Buildroot should use ta/Makefile.

This commit renames ta/Makefile to ta/Makefile.gmake so that Buildroot
may use it after it has overwriten ta/Makefile. A new ta/Makefile is
provided to preserve the plain GNU make build, which simply includes
Makefile.gmake.


etienne-lms:
+1 for the change.

Could optee_test main `CMakeList.txt` also drive build of the TAs, as proposed in https://github.com/OP-TEE/optee_test/pull/337?

jforissier:
@etienne-lms oh, I had completely forgotten about your PR! Sorry about that :-/

Yes, of course, it makes sense to use `CMakeList.txt`. Let's close this PR and continue the discussion on yours. Could you please rebase it?

etienne-lms:
Ok, I will. Yet, I though you could merge your P-R first (#374) as it seems usefull and is a prerequisite in #337 for building TAs from CMake build scripts.

I wonder if such changes  should be part of 3.7.0. I guess it not mandatory, especially since many guys already sent test tag for the rc1.

etienne-lms:
ok

PR #375, opened 2019-10-22
regression 1022: return TEE_ERROR_GENERIC on failure

jforissier:
When dlopen() or dlsym() fail in regression test 1022, the error status
is set to TEE_ERROR_OUT_OF_MEMORY which is inappropriate since the true
cause of the error is unknown. Use TEE_ERROR_GENERIC instead. This is
essentially cosmetic to avoid confusion when debugging.


jenswi-linaro:

PR #377, opened 2019-11-06
regression_4100.c: convert_from_string(): fix length of output buffer

jforissier:
convert_from_string() takes an hexadecimal character string and
converts it to binary. The size of the output buffer is computed from
the length of the input string, but the computation is wrong. Fix it.


jenswi-linaro:

PR #385, opened 2020-01-06
regression 4006: add test cases for SM2 DSA

jforissier:
Please review only commit "regression 4006: add test cases for SM2 DSA" and ignore the other commits, they are from the PKE PR. I have kept them for simplicity (avoid rebase and conflicts).


jenswi-linaro:
For the commit "regression 4006: add test cases for SM2 DSA":

PR #386, opened 2020-01-11
aes-perf, sha-perf: fix build with 64bit time_t on 32bit arches

andred:
| host/xtest/aes_perf.c: In function 'aes_perf_run_test':
| host/xtest/aes_perf.c:473:11: error: format '%lu' expects argument of type 'long unsigned int', but argument 2 has type 'time_t' {aka 'long long int'} [-Werror=format=]
|   473 |  vverbose("Clock resolution is %lu ns\n",
|       |           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
|   474 |      ts.tv_sec * 1000000000 + ts.tv_nsec);
|       |      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
|       |                             |
|       |                             time_t {aka long long int}
| host/xtest/crypto_common.h:56:11: note: in definition of macro '_verbose'
|    56 |    printf(__VA_ARGS__); \
|       |           ^~~~~~~~~~~
| host/xtest/aes_perf.c:473:2: note: in expansion of macro 'vverbose'
|   473 |  vverbose("Clock resolution is %lu ns\n",
|       |  ^~~~~~~~
| host/xtest/aes_perf.c:473:34: note: format string is defined here
|   473 |  vverbose("Clock resolution is %lu ns\n",
|       |                                ~~^
|       |                                  |
|       |                                  long unsigned int
|       |                                %llu

| host/xtest/sha_perf.c: In function 'sha_perf_run_test':
| host/xtest/sha_perf.c:323:11: error: format '%lu' expects argument of type 'long unsigned int', but argument 2 has type 'time_t' {aka 'long long int'} [-Werror=format=]
|   323 |  vverbose("Clock resolution is %lu ns\n", ts.tv_sec*1000000000 +
|       |           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~  ~~~~~~~~~~~~~~~~~~~~~~
|       |                                                                |
|       |                                                                time_t {aka long long int}
|   324 |   ts.tv_nsec);
|       |   ~~~~~~~~~~
| host/xtest/crypto_common.h:56:11: note: in definition of macro '_verbose'
|    56 |    printf(__VA_ARGS__); \
|       |           ^~~~~~~~~~~
| host/xtest/sha_perf.c:323:2: note: in expansion of macro 'vverbose'
|   323 |  vverbose("Clock resolution is %lu ns\n", ts.tv_sec*1000000000 +
|       |  ^~~~~~~~
| host/xtest/sha_perf.c:323:34: note: format string is defined here
|   323 |  vverbose("Clock resolution is %lu ns\n", ts.tv_sec*1000000000 +
|       |                                ~~^
|       |                                  |
|       |                                  long unsigned int
|       |                                %llu
| cc1: all warnings being treated as errors

The code has two problems:
1) it assumes time_t is unsigned
2) it assumes time_t is the same as unsigned long

1) rarely is true
2) is not true on 32bit architectures which have a 64bit time_t

Fix by appropriate casting and an updated printf format specifier.

Signed-off-by: André Draszik <andre.draszik@jci.com>

### DIFF
--- a/host/xtest/aes_perf.c
+++ b/host/xtest/aes_perf.c
@@ -470,8 +470,8 @@ void aes_perf_run_test(int mode, int keysize, int decrypt, size_t size, size_t u
 		perror("clock_getres");
 		return;
 	}
-	vverbose("Clock resolution is %lu ns\n",
-					ts.tv_sec * 1000000000 + ts.tv_nsec);
+	vverbose("Clock resolution is %jd ns\n",
+		 (intmax_t)ts.tv_sec * 1000000000 + ts.tv_nsec);
 
 	vverbose("input test buffer:  %s\n", buf_type_str(input_buffer));
 	vverbose("output test buffer: %s\n", buf_type_str(output_buffer));
--- a/host/xtest/sha_perf.c
+++ b/host/xtest/sha_perf.c
@@ -320,8 +320,8 @@ extern void sha_perf_run_test(int algo, size_t size, unsigned int n,
 		perror("clock_getres");
 		return;
 	}
-	vverbose("Clock resolution is %lu ns\n", ts.tv_sec*1000000000 +
-		ts.tv_nsec);
+	vverbose("Clock resolution is %jd ns\n",
+		 (intmax_t)ts.tv_sec * 1000000000 + ts.tv_nsec);
 
 	open_ta();
 	prepare_op(algo);


andred:
Well, the original problem is the incorrect format specifier for the different types of time_t out there (in linux (glibc / musl) -> long and int64_t

So, I was a bit torn on what to do, tbh.

Given this printf() outputs the resolution of the clock, and the resolution should never be 1s (I'd hope, at least), and never be negative either, an alternative could have been to not printf() the tv_sec part at all, or to just unconditionally cast to some other smaller type (i.e. not intmax_t).

> Thanks for the patch! In principle I think casting to `intmax_t` could fail if, say, `time_t` is `uinmax_t`.

One could just cast to signed int, and add extra checks to ensure tv_sec is not larger than INT_MAX (to catch the case time_t is unsigned and the clock is really really(!) low resolution) - but again, such a low res clock should not ever exist?

> There isn't even a guarantee that`time_t` is an integer type... OK these cases are probably pretty exceptional. So:

That's true (and the original code makes the (incorrect on linux) assumption that time_t is unsigned - so it'd fail in these cases, too). In practical terms, time_t is always signed on linux (glibc & musl) I think. Additionally, even coreutils has an assertion for time_t to be of type signed or unsigned int in its code base.

> (I've learned a new `printf` format specifier wink)

There are too many :-)


What would be useful though if optee_test in general was being compiled with more strict format warnings using -Wformat-signedness and -Wformat-truncation.

andred:
@jforissier, I've added your reviewed-by.

Thank you.

PR #387, opened 2020-01-15
Add regression 4014 (SM2 KEP)

jforissier:
Adds a test for the SM2 Key Exchange Protocol.


jforissier:
Test case for https://github.com/OP-TEE/optee_os/pull/3534.

jforissier:
Updated to include the algorithm availability test; commit "Add helper function ta_crypt_cmd_is_algo_supported()" is cherry-picked from https://github.com/OP-TEE/optee_test/pull/388.

jforissier:
Ping!
(I fixed a silly stack corruption bug -- `params[9]` instead of `params[4]`)

host/xtest/regression_4000.c
```diff
@@ -5543,4 +5543,335 @@ static void xtest_tee_test_4013(ADBG_Case_t *c)
...
+	TEE_ObjectHandle sv_handle = TEE_HANDLE_NULL;
+	TEE_Attribute params[9] = { };
+	size_t param_count = 0;
+	uint8_t out[128] = { };
+	size_t out_size = 0;
+	uint8_t conf_A[32];
+	uint8_t conf_B[32];
+
+	if (!ADBG_EXPECT_TEEC_SUCCESS(c,
+		xtest_teec_open_session(&session, &crypt_user_ta_uuid, NULL,
```
etienne-lms: prefer change indentation at lines 5565, 5582, 5589 5608 and so on.

host/xtest/regression_4000.c
```diff
@@ -5543,4 +5543,335 @@ static void xtest_tee_test_4013(ADBG_Case_t *c)
...
+	TEE_ObjectHandle eph_keyA = TEE_HANDLE_NULL;
+	TEE_ObjectHandle keyB = TEE_HANDLE_NULL;
+	TEE_ObjectHandle eph_keyB = TEE_HANDLE_NULL;
+	TEE_ObjectHandle sv_handle = TEE_HANDLE_NULL;
+	TEE_Attribute params[9] = { };
+	size_t param_count = 0;
+	uint8_t out[128] = { };
+	size_t out_size = 0;
+	uint8_t conf_A[32];
+	uint8_t conf_B[32];
```
etienne-lms: maybe `= { }` for both.

host/xtest/regression_4000.c
jforissier: With "proper" indentation (start after opening parenthesis), lines >80 characters cannot be avoided. And many tests use this style already. So I'm not sure which is best.

host/xtest/regression_4000.c
jforissier: OK

jforissier:
> I would have preferred some more test vector, but if that's all we have...
> `Acked-by: Jens Wiklander <jens.wiklander@linaro.org>`

There is another one in Part 3 of the spec, but no other that I know of.

host/xtest/regression_4000.c
jenswi-linaro: We could rewrite it as:
``` C
res = xtest_teec_open_session(...)
if (!ADBG_EXPECT_TEEC_SUCCESS(c, res))
         return;
```
However, I'm not sure that should be in the scope of this PR.

jforissier: @jenswi-linaro true, we have about 800 occurrences of this pattern... all are more or less formatted like the above.

jforissier:
Update.

PR #388, opened 2020-01-22
Detect SM2/SM3/SM4 support at run time

jforissier:
Needs ~~https://github.com/OP-TEE/optee_os/pull/3545~~ Edit: https://github.com/OP-TEE/optee_os/pull/3560 was merged instead.

### DIFF
--- a/host/xtest/regression_4000.c
+++ b/host/xtest/regression_4000.c
@@ -980,6 +980,13 @@ static void xtest_tee_test_4001(ADBG_Case_t *c)
 		uint8_t out[64] = { };
 		size_t out_size = 0;
 
+		if (hash_cases[n].algo == TEE_ALG_SM3 &&
+		    !ta_crypt_cmd_is_algo_supported(c, &session, TEE_ALG_SM3,
+						    TEE_CRYPTO_ELEMENT_NONE)) {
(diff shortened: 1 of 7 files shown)


jforissier:
Ping?

host/xtest/regression_4000.c
```diff
@@ -980,6 +980,13 @@ static void xtest_tee_test_4001(ADBG_Case_t *c)
 		uint8_t out[64] = { };
 		size_t out_size = 0;
 
+		if (hash_cases[n].algo == TEE_ALG_SM3 &&
```
jenswi-linaro: Why are we only checking this for `TEE_ALG_SM3`? Because it's safer this way in case you don't read the log carefully?

jforissier: Yes. Since other algorithms are mandatory (as per GP), we probably want to make sure they have not been disabled by mistake.
On the other hand, if anyone configures a special version of OP-TEE without support for some algorithms, he will likely want to test it with `xtest` too -- that would require changing the code a bit.

jenswi-linaro: Agree.

jenswi-linaro:

PR #389, opened 2020-01-23
regression 4110: add test vector

jforissier:
Adds a test vector for addmod/mulmod operations. These particular
values were found to fail the mulmod test when CFG_TA_MBEDTLS_MPI=n, in
other words: when the MPA library is used instead of Mbed TLS MPI.


jforissier:
It very much looks like there is a bug in `__mpa_div_q_r_internal()`.

PR #390, opened 2020-02-06
pkcs11 1000: test finalize aginst non-initiated library

etienne-lms:
Call `C_Finalize()` without prior `C_Initialize()`. Library shall return `CKR_CRYPTOKI_NOT_INITIALIZED`.


etienne-lms:
tag applied

PR #391, opened 2020-02-12
regression_1008: skip subtests if PTA is unavailable

jenswi-linaro:
Replaces the compile time configuration based on CFG_SECSTOR_TA_MGMT_PTA
with a runtime test if PTA_SECSTOR_TA_MGMT_UUID is available.

Skips the "Load corrupt TA" substests if PTA_SECSTOR_TA_MGMT_UUID isn't
available.


jenswi-linaro:
Tag applied

PR #392, opened 2020-02-14
GitHub: sync GitHub templates and actions from build.git

jbech-linaro:
Since "optee_os" and "build" both have reached an almost clean and up-to-date state, let's continue with adding templates and actions to the "optee_test" git also.

The GitHub templates and actions are useful to the "optee_test" git
also, therefore recursively sync up (copy) the .github folder from
build.git.

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,0 +1,21 @@
+<!--
+    General guidance when creating issues:
+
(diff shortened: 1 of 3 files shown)


jbech-linaro:
Squashed, rebased and tag(s) applied (should be) ready for merge! Thanks!

PR #394, opened 2020-02-18
Add regression 1023 (ELF initialization via .init_array)

jforissier:
Adds a test case for the ELF initialization functions (.init_array
section).

Signed-off-by: Jerome Forissier <jerome@forissier.org>

PR #398, opened 2020-02-20
Regression 1023: add tests cases for PTA_SYSTEM_GET_TPM_EVENT_LOG ser…

javieralso-arm:
…vice.

Adds test cases to exercise the PTA_SYSTEM_GET_TPM_EVENT_LOG service.

Signed-off-by: Javier Almansa Sobrino <javier.almansasobrino@arm.com>


javieralso-arm:
I wasn't sure where to add the tests for this so I added them to regression_1000.c (regression 1023). Let me know in case they should go somewhere else.

Cheers,
Javier

host/xtest/regression_1000.c
```diff
@@ -1744,3 +1746,43 @@ static void xtest_tee_test_1022(ADBG_Case_t *c)
...
+		if (!ADBG_EXPECT_TEEC_SUCCESS(c,
+		    xtest_teec_open_session(&session, &tpm_log_test_ta_uuid,
+					    NULL, &ret_orig)))
+			return;
+	}
+	Do_ADBG_EndSubCase(c, "TPM test create session");
+
+	Do_ADBG_BeginSubCase(c, "TPM test service invocation");
+	{
+		TEEC_Operation op = TEEC_OPERATION_INITIALIZER;
```
jforissier: If no operation is needed, you should pass NULL to `TEEC_InvokeCommand()`.

But, since we're driving everything from the client app (this code), wouldn't it be simpler to have the TA just forward the operation to the PTA? Basically, `TA_InvokeCommandEntryPoint()` below would do what `invoke_system_pta()` currently does.

host/xtest/regression_1000.c
```diff
@@ -1744,3 +1746,43 @@ static void xtest_tee_test_1022(ADBG_Case_t *c)
...
+		ADBG_EXPECT_TEEC_SUCCESS(c, TEEC_InvokeCommand(&session,
+					  TA_TPM_TEST_SHORT_BUF,
+					  &op, &ret_orig));
+	}
+	Do_ADBG_EndSubCase(c, "TPM test passing short buffer");
+
+	TEEC_CloseSession(&session);
+}
+
+ADBG_CASE_DEFINE(regression, 1023, xtest_tee_test_1023,
```
jforissier: Adding a new test case is fine. But, I have already used 1023 in https://github.com/OP-TEE/optee_test/pull/394.
We'll see what gets merged first.

ta/tpm_log_test/include/user_ta_header_defines.h
```diff
@@ -0,0 +1,17 @@
...
+ */
+
+#ifndef USER_TA_HEADER_DEFINES_H
+#define USER_TA_HEADER_DEFINES_H
+
+#include <ta_tpm_log.h>
+
+#define TA_UUID TA_TPM_LOG_TEST_UUID
+
+#define TA_FLAGS		(TA_FLAG_USER_MODE | TA_FLAG_EXEC_DDR)
```
jforissier: Both flags are deprecated, please use `0` instead (yes a lot of code would need cleaning up ;-)).

host/xtest/regression_1000.c
javieralso-arm: > But, since we're driving everything from the client app (this code), wouldn't it be simpler to have the TA just forward the operation to the PTA? Basically, TA_InvokeCommandEntryPoint() below would do what invoke_system_pta() currently does.

So you mean to move most of the logic to test the event log inside my implementation of TA_InvokeCommandEntryPoint() to the regression test and just forward the commands from the PA to the TPA, right?
I actually considered that option, but I rejected it as:

1. I wanted to keep the regression test as simple as possible as I noticed that most of the tests on regression_1000.c were relatively simple.

2. I decided that the logic to check that a TPM event log was properly read was best placed inside the TA. On a real fTPM implementation, the event log would be processed inside the TA and not forwarded to the client app, so I didn't see a reason why I should move that logic to the client and then forward the event log there.

Cheers,
Javier

ta/tpm_log_test/include/user_ta_header_defines.h
javieralso-arm: I noticed that TA_FLAGS is not used, should I just remove it instead of making it 0?

host/xtest/regression_1000.c
javieralso-arm: NP, I will leave it untouched and change it later if there were any conflict.

host/xtest/regression_1000.c
jforissier: Correct, that's what I meant.

I have no strong opinion either way, so it is enough for me to know that you considered both options and chose the one you liked most :)

jforissier: My comment about removing `op` if you don't need it still holds however.

host/xtest/regression_1000.c
jforissier: Fine.

ta/tpm_log_test/include/user_ta_header_defines.h
jforissier: It is used in [`user_ta_header.c`](https://github.com/OP-TEE/optee_os/blob/3.8.0/ta/arch/arm/user_ta_header.c#L83) which is imported as part of the TA build.

But my suggestion to use zero is not good, see [here](https://github.com/linaro-swg/optee_examples/pull/69). For consistency I'd like to see `TA_FLAG_EXEC_DDR` here.

host/xtest/regression_1000.c
```diff
@@ -34,7 +35,9 @@
 #include <ta_miss_test.h>
 #include <ta_sims_keepalive_test.h>
 #include <ta_concurrent.h>
+#include <ta_tpm_log_test.h>
 #include <sdp_basic.h>
+#include <conf.h>
```
jforissier: Why `<conf.h>`?

host/xtest/xtest_test.h
```diff
@@ -98,6 +98,9 @@ extern const unsigned int gp_tta_tcf_singleinstance_ta_size;
 extern const char gp_tta_tcf_multipleinstance_ta[];
 extern const unsigned int gp_tta_tcf_multipleinstance_ta_size;
 
+extern const char tpm_log_test_ta[];
+extern const unsigned int tpm_log_test_ta_size;
```
jforissier: I can't find why these are needed?

host/xtest/regression_1000.c
javieralso-arm: I wanted to add conditional compilation on CFG_CORE_TPM_EVENT_LOG to avoid IBART fails if OP-TEE wasn't built with support for the TPM event log.

host/xtest/xtest_test.h
javieralso-arm: Hmmm... Forgot to remove them whilst refactoring

host/xtest/regression_1000.c
jforissier: The `#ifdef CFG_CORE_TPM_EVENT_LOG` is fine, but `#include <conf.h>` is useless due to:
```
host/xtest/Makefile:159:CFLAGS += -include conf.h
```

ta/tpm_log_test/ta_entry.c
```diff
@@ -0,0 +1,161 @@
...
+	case TA_TPM_TEST_SHORT_BUF:
+	/*
+	 * Test case: Insufficient space on the buffer.
+	 */
+		params[0].memref.buffer = (void *)buf;
+		params[0].memref.size = 1;
+		memset(buf, 0, LOG_SIZE);
+
+		if (invoke_system_pta(PTA_SYSTEM_GET_TPM_EVENT_LOG,
+				      param_types, params) == TEE_SUCCESS) {
```
jforissier: Why not expect `== TEE_ERROR_SHORT_BUFFER`? Other error codes would indicate a problem so you don't want to report success in this case.

javieralso-arm: Comments fixed. I also renamed the test to 1024, as test 1023 was in use.

Thanks

ta/tpm_log_test/include/ta_tpm_log_test.h
```diff
@@ -0,0 +1,19 @@
+/* SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Copyright (c) 2020, ARM Limited. All rights reserved.
+ */
+
+#ifndef TA_TPM_LOG_TEST
```
etienne-lms: suggest adding `_H` as suffix

ta/tpm_log_test/include/user_ta_header_defines.h
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * Copyright (c) 2020, ARM Limited. All rights reserved.
+ */
+
+#ifndef USER_TA_HEADER_DEFINES_H
+#define USER_TA_HEADER_DEFINES_H
+
+#include <ta_tpm_log.h>
+
+#define TA_UUID TA_TPM_LOG_TEST_UUID
```
etienne-lms: minor: add a tabulation to indent as the few lines below.

ta/tpm_log_test/ta_entry.c
```diff
@@ -0,0 +1,162 @@
+// SPDX-License-Identifier: BSD-3-Clause
+/*
+ * Copyright (c) 2020, ARM Limited. All rights reserved.
+ */
+
+#include <ta_tpm_log.h>
+#include <string.h>
```
etienne-lms: more few lines below.

ta/tpm_log_test/ta_entry.c
```diff
@@ -0,0 +1,162 @@
...
+	const uint32_t param_types = TEE_PARAM_TYPES(
+						TEE_PARAM_TYPE_MEMREF_OUTPUT,
+						TEE_PARAM_TYPE_NONE,
+						TEE_PARAM_TYPE_NONE,
+						TEE_PARAM_TYPE_NONE);
+	TEE_Param params[TEE_NUM_PARAMS] = { 0 };
+
+	switch (nCommandID) {
+	case TA_TPM_TEST_GET_LOG:
+	/*
```
etienne-lms: suggest to move each command sequence in a local function.

ta/tpm_log_test/ta_entry.c
```diff
@@ -0,0 +1,162 @@
...
+			uint32_t field = 0;
+			unsigned char *digestptr = NULL;
+			unsigned int i;
+
+			MSG("Received event log with %i bytes",
+			    params[0].memref.size);
+			MSG("Parsing the event log header");
+			memcpy(&field, &buf[0], sizeof(uint32_t));
+
+			MSG("--> PCRIndex = 0x%" PRIx32, field);
```
etienne-lms: Replace `MSG()` with `DMSG()`.

ta/tpm_log_test/ta_entry.c
```diff
@@ -0,0 +1,162 @@
...
+ * Copyright (c) 2020, ARM Limited. All rights reserved.
+ */
+
+#include <ta_tpm_log.h>
+#include <string.h>
+#include <inttypes.h>
+#include <pta_system.h>
+#include <tee_internal_api.h>
+
+#define LOG_SIZE 1024
```
etienne-lms: prefer 1 or 2 tabs before the value.

ta/tpm_log_test/ta_entry.c
```diff
@@ -0,0 +1,162 @@
...
+{
+}
+
+/* Called when a command is invoked */
+TEE_Result TA_InvokeCommandEntryPoint(void *pSessionContext __unused,
+				      uint32_t nCommandID,
+				      uint32_t nParamTypes __unused,
+				      TEE_Param pParams[4] __unused)
+{
+	unsigned char buf[LOG_SIZE];
```
etienne-lms: ` = { };`

javieralso-arm: Comments fixed. Thanks

jforissier:
@etienne-lms do you want to give an Acked-by or Reviewed-by tag?

ta/tpm_log_test/include/ta_tpm_log.h
```diff
@@ -0,0 +1,12 @@
+/* SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Copyright (c) 2020, ARM Limited. All rights reserved.
+ */
+
+#ifndef TPM_LOG_TEST_H
+#define TPM_LOG_TEST_H
```
etienne-lms: prefer `TA_TMP_LOG_H` as per file name

ta/tpm_log_test/include/ta_tpm_log_test.h
```diff
@@ -0,0 +1,19 @@
+/* SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Copyright (c) 2020, ARM Limited. All rights reserved.
+ */
+
+#ifndef TA_TPM_LOG_TEST_H
+#define TA_TPM_LOG_TEST_H
+
+/*
+ *This UUID is generated with the ITU-T UUID generator at
```
etienne-lms: minor: add space before `This`.

ta/tpm_log_test/include/user_ta_header_defines.h
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * Copyright (c) 2020, ARM Limited. All rights reserved.
+ */
+
+#ifndef USER_TA_HEADER_DEFINES_H
+#define USER_TA_HEADER_DEFINES_H
+
+#include <ta_tpm_log.h>
+
+#define				TA_UUID TA_TPM_LOG_TEST_UUID
```
etienne-lms: fix indent, prefer "#define<space>NAME<tab>VALUE"

ta/tpm_log_test/include/user_ta_header_defines.h
```diff
@@ -0,0 +1,17 @@
...
+
+#define TA_FLAGS		TA_FLAG_EXEC_DDR
```
etienne-lms: `TA_FLAG_EXEC_DDR` refers to a deprecated flag. Prefer simply `#define TA_FLAGS	0`.

ta/tpm_log_test/ta_entry.c
etienne-lms: todo:
```c
#include <inttypes.h>
#include <pta_system.h>
#include <string.h>
#include <ta_tpm_log.h>
#include <tee_internal_api.h>

File: ta/tpm_log_test/ta_entry.c [hunk 18]
```diff
@@ -0,0 +1,177 @@
...
+	unsigned char *digestptr = NULL;
+
+	params[0].memref.buffer = (void *)buf;
+	params[0].memref.size = LOG_SIZE;
+
+	if (invoke_system_pta(PTA_SYSTEM_GET_TPM_EVENT_LOG,
+			      param_types, params) == TEE_SUCCESS) {
+
+		DMSG("Received event log with %i bytes",
+		    params[0].memref.size);
```
etienne-lms: checkpatch would complain on indentation. prefer fixing it.

ta/tpm_log_test/ta_entry.c
```diff
@@ -0,0 +1,177 @@
...
+		DMSG("Parsing the event log header");
+		memcpy(&field, &buf[0], sizeof(uint32_t));
+
+		DMSG("--> PCRIndex = 0x%" PRIx32, field);
```
etienne-lms: really need to keep these?
If so, maybe remove the `--> ` prefix and possibly remove the one at line 58, add an empty line before lines 74 and 97.

ta/tpm_log_test/ta_entry.c
```diff
@@ -0,0 +1,177 @@
...
+			return TEE_ERROR_GENERIC;
+		}
+
+		/*
+		 * Header digest must be formed of 20 bytes set to 0.
+		 * Ref. Section 5.3 of TCG EFI Protocol Specification.
+		 * Family 2.0 Level 00 Revision 00.12. March 30, 2016
+		 */
+		digestptr = &buf[8];
+		for (unsigned int i = 0; i < 20; i++) {
```
etienne-lms: define `unsigned int i = 0;` at function entry, or at current instruction block entry.

ta/tpm_log_test/ta_entry.c
```diff
@@ -0,0 +1,177 @@
...
+
+	params[0].memref.buffer = (void *)buf;
+	params[0].memref.size = 1;
+
+	if (invoke_system_pta(PTA_SYSTEM_GET_TPM_EVENT_LOG, param_types,
+			      params) != TEE_ERROR_SHORT_BUFFER) {
+		return TEE_ERROR_GENERIC;
+	}
+
+	for (unsigned int i = 0; i < LOG_SIZE; i++) {
```
etienne-lms: define `unsigned int i = 0;` at function entry.

ta/tpm_log_test/ta_entry.c
javieralso-arm: Sorry Etienne, I don't quite understand what do you want me to do there. Just rearrange the includes on that order?

Cheers,
Javier

etienne-lms: sorry, we prefer alphabetical sorting in includes, hence to proposed ordering.

javieralso-arm: Perfect, I get it now.

Thanks,
Javier

ta/tpm_log_test/ta_entry.c
javieralso-arm: All done. Thanks for the comments.
Javier

ta/tpm_log_test/include/ta_tpm_log.h
```diff
@@ -0,0 +1,12 @@
+/* SPDX-License-Identifier: BSD-3-Clause
+ *
```
etienne-lms: @jforissier, is this ok:
```c
/* SPDX-License-Identifier: BSD-3-Clause
 *
 * Copyright (...)
```
or prefer the below?
```c
/* SPDX-License-Identifier: BSD-3-Clause */
/*
 * Copyright (...)
```

ta/tpm_log_test/ta_entry.c
```diff
@@ -0,0 +1,181 @@
...
+	uint32_t field = 0;
+	unsigned char *digestptr = NULL;
+	unsigned int i = 0;
+
+	params[0].memref.buffer = (void *)buf;
+	params[0].memref.size = LOG_SIZE;
+
+	if (invoke_system_pta(PTA_SYSTEM_GET_TPM_EVENT_LOG,
+			      param_types, params) == TEE_SUCCESS) {
+
```
etienne-lms: remove empty line

ta/tpm_log_test/include/ta_tpm_log.h
jforissier: @etienne-lms good catch. Second form is preferred, in fact I don't think the first one passes checkpatch.

PR #399, opened 2020-03-02
pkcs11 1001: get info from slots

etienne-lms:
Add test to list slots implemented by the PKCS11 TA and get information from. This change tests 
 Cryptoki API functions `C_GetFunctionList()`, `C_GetInfo()`, `C_GetSlotList()` and `C_GetSlotInfo()`.

Depends on https://github.com/OP-TEE/optee_client/pull/190 and https://github.com/OP-TEE/optee_os/pull/3644.

### DIFF
--- a/host/xtest/pkcs11_1000.c
+++ b/host/xtest/pkcs11_1000.c
@@ -11,11 +11,12 @@
  * GNU General Public License for more details.
  */
 
-#include <stdio.h>
-#include <string.h>
+#include <ck_debug.h>
 #include <inttypes.h>
-
 #include <pkcs11.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
 
 #include "xtest_test.h"
 #include "xtest_helpers.h"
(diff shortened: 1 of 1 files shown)


host/xtest/pkcs11_1000.c
```diff
@@ -48,3 +49,69 @@ static void xtest_tee_test_1000(ADBG_Case_t *c)
...
+	rv = C_GetSlotList(1, NULL, &present_slot_count);
+	if (!ADBG_EXPECT_CK_OK(c, rv))
+		goto out;
+
+	if (!ADBG_EXPECT_COMPARE_UNSIGNED(c, slot_count, ==,
+					  present_slot_count))
+		goto out;
+
+	slot_ids = calloc(slot_count, sizeof(CK_SLOT_ID));
+	if (!ADBG_EXPECT_TRUE(c, !slot_count || slot_ids))
```
jenswi-linaro: If you split this into two ADBG_EXPECT_XXX calls we could actually tell what's the error in the log without further debugging.

etienne-lms: Ok. Actually, we expect at least 1 slot/token, otherwise, there will be nothing to test. So I think `slot_count != 0` could be tested from line 84 above.

host/xtest/pkcs11_1000.c
```diff
@@ -48,3 +49,81 @@ static void xtest_tee_test_1000(ADBG_Case_t *c)
...
+	slot_ids = calloc(slot_count, sizeof(CK_SLOT_ID));
+	if (!ADBG_EXPECT_NOT_NULL(c, slot_ids))
+		goto out;
+
+	rv = C_GetSlotList(1, slot_ids, &slot_count);
+	if (!ADBG_EXPECT_CK_OK(c, rv))
+		goto out;
+
+	for (i = 0; i < slot_count; i++) {
+		CK_SLOT_ID slot = *(slot_ids + i);
```
jenswi-linaro: I'm starting to get used to people using `&ptr[idx]` instead `(ptr + idx)` as a way of hiding the pointer arithmetics. This is the actually the opposite using pointer arithmetics to avoid array indexing.

How about writing `slot_idx[i]` instead since we're actually dereferencing too.

etienne-lms: right :)  sorry.

jenswi-linaro: No problem. :-)

etienne-lms:
Fixup commit squashed. Tag applied.

PR #400, opened 2020-03-04
pkcs11 1001: test C_GetSlotList() for zero-count non-NULL buffer case

etienne-lms:
Check C_GetSlotList() returns CKR_BUFFER_TOO_SMALL, not CKR_OK when
called with a non-NULL ID buffer reference and a count info to ID.
Prior this change was this specific test not covered in case TA
implements more than 1 token.


etienne-lms:
review tag applied

PR #402, opened 2020-03-09
os_test: call_longjmp(): add __noreturn attribute

jforissier:
call_longjmp() does not return so it should have a __noreturn attribute.
Fixes a Clang warning:

 $ make COMPILER=clang
 ...
   CC      out/os_test.o
 os_test.c:667:1: warning: function 'call_longjmp' could be declared with attribute 'noreturn' [-Wmissing-noreturn]
 {
 ^

### DIFF
--- a/ta/os_test/os_test.c
+++ b/ta/os_test/os_test.c
@@ -663,7 +663,7 @@ static TEE_Result test_float(void)
 }
 #endif /*CFG_TA_FLOAT_SUPPORT*/
 
-static __noinline void call_longjmp(jmp_buf env)
+static __noinline __noreturn void call_longjmp(jmp_buf env)
 {
 	DMSG("Calling longjmp");
 	longjmp(env, 1);


etienne-lms:
Since this change, I get a build warning with GCC 8.3.0 
Looks like i should have tested it ;-)
```
os_test.c: In function ‘call_longjmp’:
os_test.c:671:1: warning: ‘noreturn’ function does return
 }
 ^
```

Maybe add `while (1) { TEE_Panic(0); }` at function end to comply with gcc and clang expectations?

jforissier:
Doh!
I'd rather we add `__noreturn` to the `setjmp()` prototype. It looks like Clang knows about it in a way or another, while GCC doesn't.

jenswi-linaro:
+1 to `__noreturn` attribute for `longjmp()` prototype.

jforissier:
https://github.com/OP-TEE/optee_os/pull/3669

etienne-lms:
thanks

PR #404, opened 2020-03-13
pkcs11 1002: test session API functions

etienne-lms:
Add test of C_OpenSession(), C_GetSessionInfo(), C_CloseSession() and
C_CloseAllSessions() from PKCS11 TA test token in pkcs11 regression
test 1002.

Test token over which test is made is the last slot found if the slot
list reported by the Cryptoki library. Helper function
init_lib_and_find_token_slot() takes care of Cryptoki pre-session setup.


etienne-lms:
tag applied

PR #406, opened 2020-03-26
OP-TEE session login checks for client UUID generation

vesajaaskelainen:
These changes adds testing for handling of open session client UUID generation within Linux kernel.

This is related to: https://github.com/OP-TEE/optee_os/issues/3642


host/xtest/Makefile
```diff
@@ -73,11 +73,13 @@ srcs +=	adbg/src/adbg_case.c \
 	regression_7000.c \
 	regression_8000.c \
 	regression_8100.c \
+	session_1000.c \
```
jenswi-linaro: Please move the content of this file into `regression_1000.c` instead and make it part of the regression suite instead continuing from 1024 or whatever is the next free number in the series.

vesajaaskelainen: Ok.

In this case is it OK to rebase/rework the branch as it makes reviewing a bit easier? (your guidance was to add new commits and then squash/fixup at the end)

jenswi-linaro: Please go ahead and rebase and rework as needed.

vesajaaskelainen: Moved, rebased and force pushed a new version

jforissier:
One more thing: I think it would be good to accept UID 1001 and GID 1001, too. Those are created by our buildroot environment: 
https://github.com/OP-TEE/build/blob/3.8.0/br-ext/package/optee_client/optee_client.mk#L26-L31
```
$ id test
uid=1001(test) gid=1003(test) groups=1003(test),1001(teeclnt),1002(ion)
```
I rarely run `xtest` as root, I normally use this `test` login.

Note, in this case perhaps we should also update `optee_client.mk` and assign hard-coded UID and GID values (1001) instead of letting buildroot choose (-1).

host/xtest/regression_1000.c
```diff
@@ -1983,3 +1984,175 @@ static void xtest_tee_test_1025(ADBG_Case_t *c)
...
+/*
+ * TEE Client UUID name space identifier (UUIDv4):
+ * Linux: 58ac9ca0-2086-4683-a1b8-ec4bc08e01b6
+ */
+
+/* TEEC_LOGIN_PUBLIC's Client UUID is NIL UUID */
+static TEEC_UUID client_uuid_public = { };
+
+/* UUID for user == root(0) */
+// $ uuidgen -n 58ac9ca0-2086-4683-a1b8-ec4bc08e01b6 --sha1 -N "user=0"
```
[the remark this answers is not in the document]
vesajaaskelainen: Changing to user->uid

host/xtest/regression_1000.c
```diff
@@ -1983,3 +1984,175 @@ static void xtest_tee_test_1025(ADBG_Case_t *c)
...
+static const char *client_uuid_user_root = "1893ce12-1205-59ed-af8a-85e152ab8240";
+
+/* UUID for group == root(0) */
+// $ uuidgen -n 58ac9ca0-2086-4683-a1b8-ec4bc08e01b6 --sha1 -N "group=0"
```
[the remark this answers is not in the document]
vesajaaskelainen: Changing to group->gid

vesajaaskelainen:
> One more thing: I think it would be good to accept UID 1001 and GID 1001, too. Those are created by our buildroot environment:
> https://github.com/OP-TEE/build/blob/3.8.0/br-ext/package/optee_client/optee_client.mk#L26-L31
> 
> ```
> $ id test
> uid=1001(test) gid=1003(test) groups=1003(test),1001(teeclnt),1002(ion)
> ```
> 
> I rarely run `xtest` as root, I normally use this `test` login.
> 
> Note, in this case perhaps we should also update `optee_client.mk` and assign hard-coded UID and GID values (1001) instead of letting buildroot choose (-1).

I can change it to use uid and gid dynamically aka. calculate UUIDv5 on the fly.

jforissier:
> I can change it to use uid and gid dynamically aka. calculate UUIDv5 on the fly.

Sounds good!

host/xtest/regression_1000.c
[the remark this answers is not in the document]
vesajaaskelainen: // line was removed due to dynamic uid handling.

host/xtest/regression_1000.c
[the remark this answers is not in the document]
vesajaaskelainen: // line was removed due to dynamic gid handling.

vesajaaskelainen:
@jforissier I have now updated the PR with UUIDv5 calculation. There was this endianess thing that took a while. It would have been nice if big endian format (byte array) would have been used for UUID structure like what it was in Linux kernel.

I added your ack/reviews for older commits and for those that were changed afterwards don't have those.

I have also now updated uid/gid scheme in Linux kernel driver PR.

jforissier:
@vesajaaskelainen thanks for the update, this looks pretty good to me -- I will review more closely tomorrow.

I realize the code introduced in "xtest: add UUIDv5 calculation helper" is non-trivial and has to be done on the client side in any application that wants to use the user- or group-based login methods. Could we make things easier by adding a function to `libteec`? The annoying things is, SHA-1 is needed and I'd rather avoid introducing dependencies on e.g. OpenSSL etc.
Perhaps we could use a simple SHA-1 implementation?

vesajaaskelainen:
> @vesajaaskelainen thanks for the update, this looks pretty good to me -- I will review more closely tomorrow.
> 
> I realize the code introduced in "xtest: add UUIDv5 calculation helper" is non-trivial and has to be done on the client side in any application that wants to use the user- or group-based login methods. Could we make things easier by adding a function to `libteec`? The annoying things is, SHA-1 is needed and I'd rather avoid introducing dependencies on e.g. OpenSSL etc.
> Perhaps we could use a simple SHA-1 implementation?

Now there is a confusion here.

Kernel is expected to form the Client UUID so that it has any trustworthiness.

The code for that is in Linaro's Linux PR:
https://github.com/linaro-swg/linux/pull/74

So there is no need to do this operation in user space at all. Only in group based login user space is expected to provide group ID otherwise the connection data is NULL.

The test here is so that we query the Client ID from TA, eg we want to see that the kernel generated the right thing. And in order to verify it in case where test executor can change the logic is copied from kernel code so that _verification_ can be done.

So there is no need to have SHA-1 in client library at all. At least for this operation. It is only needed for running xtest.

Earlier version was expecting that the xtest was executed as root and this now makes it as dynamic.

host/xtest/regression_1000.c
```diff
@@ -2036,3 +2037,190 @@ static void xtest_tee_test_1025(ADBG_Case_t *c)
...
+		return;
+
+	sprintf(uuid_name, "gid=%x", group);
+
+	result = xtest_uuid_v5(&expected_client_uuid, &uuid_ns, uuid_name,
+			       strlen(uuid_name));
+	if (!ADBG_EXPECT_TEEC_SUCCESS(c, result))
+		return;
+
+	result = TEEC_InitializeContext(NULL, &context);
```
jenswi-linaro, line 2198: It's OK to rely on `xtest_teec_ctx` being properly initialized here.

host/xtest/regression_1000.c
```diff
@@ -2036,3 +2037,190 @@ static void xtest_tee_test_1025(ADBG_Case_t *c)
...
+		return;
+
+	sprintf(uuid_name, "uid=%x", geteuid());
+
+	result = xtest_uuid_v5(&expected_client_uuid, &uuid_ns, uuid_name,
+			       strlen(uuid_name));
+	if (!ADBG_EXPECT_TEEC_SUCCESS(c, result))
+		return;
+
+	result = TEEC_InitializeContext(NULL, &context);
```
jenswi-linaro, line 2141: It's OK to rely on `xtest_teec_ctx` being properly initialized here.

jforissier:
@vesajaaskelainen 

> Now there is a confusion here.
> 
> Kernel is expected to form the Client UUID so that it has any trustworthiness.
> 
> The code for that is in Linaro's Linux PR:
> [linaro-swg/linux#74](https://github.com/linaro-swg/linux/pull/74)
> 
> So there is no need to do this operation in user space at all. Only in group based login user space is expected to provide group ID otherwise the connection data is NULL.

Ah, I see. Makes sense now. I should have waited until properly reading the code before commenting ;-). Sorry for the confusion.

host/xtest/regression_1000.c
```diff
@@ -2036,3 +2037,190 @@ static void xtest_tee_test_1025(ADBG_Case_t *c)
...
+	TEEC_Session session = { };
+	uint32_t login = (uint32_t)-1;
+	TEEC_UUID client_uuid = { };
+
+	result = TEEC_InitializeContext(NULL, &context);
+	if (!ADBG_EXPECT_TEEC_SUCCESS(c, result))
+		return;
+
+	result = TEEC_OpenSession(&context, &session, &os_test_ta_uuid,
+				  TEEC_LOGIN_PUBLIC, NULL, NULL, &ret_orig);
```
etienne-lms, line 2092: As mentioned by @jenswi-linaro below, remove this and use `xtest_teec_ctx`  at line 2091.

```c
       result = TEEC_OpenSession(&xtest_teec_ctx, &session,
				 &pta_invoke_tests_ta_uuid,
				 NULL, &ret_orig);
	if (!ADBG_EXPECT_TEEC_SUCCESS(c, result))
		return;

	result = get_client_identity(&session, &login, &client_uuid);
	if (!ADBG_EXPECT_TEEC_SUCCESS(c, result))
		goto out;

	ADBG_EXPECT_COMPARE_UNSIGNED(c, login, ==, TEEC_LOGIN_PUBLIC);

	ADBG_EXPECT_EQUAL(c, &client_uuid_public, &client_uuid,
			  sizeof(TEEC_UUID));

out:
	TEEC_CloseSession(&session);
}
```

And same inside `xtest_tee_test_1027()` and `xtest_tee_test_1028()` below.

host/xtest/regression_1000.c
```diff
@@ -2036,3 +2037,190 @@ static void xtest_tee_test_1025(ADBG_Case_t *c)
...
+{
+	TEEC_Result result = TEEC_ERROR_GENERIC;
+	uint32_t ret_orig = 0;
+	TEEC_Context context = { };
+	TEEC_Session session = { };
+	uint32_t login = (uint32_t)-1;
+	TEEC_UUID client_uuid = { };
+	TEEC_UUID expected_client_uuid = { };
+	TEEC_UUID uuid_ns = { };
+	char uuid_name[TEE_UUID_NS_NAME_SIZE];
```
etienne-lms, line 2181: ` = { 0 };` or ` = { };`

host/xtest/xtest_uuid_helpers.c
```diff
@@ -0,0 +1,149 @@
...
+	TEEC_UUID u = { };
+	const char *p = s;
+	size_t i = 0;
+
+	if (!p || strnlen(p, 37) != 36)
+		return TEEC_ERROR_BAD_FORMAT;
+	if (p[8] != '-' || p[13] != '-' || p[18] != '-' || p[23] != '-')
+		return TEEC_ERROR_BAD_FORMAT;
+
+	u.timeLow = parse_hex(p, 8, &res);
```
etienne-lms, line 57: suggest:
```
static TEEC_Result advance_dash(const char **s)
{
	if (**s != '-')
		return TEEC_ERROR_BAD_FORMAT;

	(**s)++;

	return TEEC_SUCCESS;
}

static TEEC_Result parse_hex32_and_adance(const char **s, size_t nchars, uint32_t *dst)
{
	uint32_t v = 0;
	size_t n = 0;
	int c = 0;

	for (n = 0; n < nchars; n++) {
		c = hex((*s)[n]);
		if (c == -1)
			return TEE_ERROR_BAD_FORMAT;

		v = (v << 4) + c;
	}

	*dest = v;
	*s += nchars;

	return TEE_SUCCESS;
}

TEEC_Result xtest_uuid_from_str(TEEC_UUID *uuid, const char *s)
{	TEEC_UUID u = { };
	const char *p = s;
	size_t i = 0;

	if (get_hex32_and_advance(&p, 8, &u.timeLow)
	    advance_dash(&p) ||
	    get_hex32_and_advance(&p, 8, &u.timeMid)
	    advance_dash(&p) ||
	    get_hex32_and_advance(&p, 4, &u.timeHiAndVersion)
	    advance_dash(&p))
		return TEEC_ERROR_BAD_FORMAT;

	for (i = 0; i < 8; i++) {
		if (get_hex32_and_advance(&p, 2, u.clockSeqAndNode + i))
			return TEEC_ERROR_BAD_FORMAT;

		if (i < 7 && advance_dash(&p))
			return TEEC_ERROR_BAD_FORMAT;
	}

	*uuid = u;

	return TEE_SUCCESS;
}

File: host/xtest/xtest_uuid_helpers.h [hunk 9]
```diff
@@ -0,0 +1,23 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2014, STMicroelectronics International N.V.
```
etienne-lms, line 3: Maybe ` * Copyright (c) 2019, Linaro Limited` as per new source file **xtest_uuid_helpers.c**.

host/xtest/xtest_uuid_helpers.c
jforissier: Is it easier to read? I don't think so. And, the commit simply imports existing code, there is no reason to change it IMO.

host/xtest/xtest_uuid_helpers.h
vesajaaskelainen: As this is a copy from same project other git tree I would not like to change it. But in here I can put what ever is recommended action from project team. Probably should be own commit after the copy or so.

host/xtest/regression_1000.c
vesajaaskelainen: ok

host/xtest/regression_1000.c
```diff
@@ -2036,3 +2037,190 @@ static void xtest_tee_test_1025(ADBG_Case_t *c)
...
+	return TEEC_SUCCESS;
+}
+
+static void xtest_tee_test_1026(ADBG_Case_t *c)
+{
+	TEEC_Result result = TEEC_ERROR_GENERIC;
+	uint32_t ret_orig = 0;
+	TEEC_Context context = { };
+	TEEC_Session session = { };
+	uint32_t login = (uint32_t)-1;
```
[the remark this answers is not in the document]
vesajaaskelainen: I am fine with either way. Which ever is recommended approach in here. If there are counter recommendations I will go for UINT32_MAX as that removes the cast at least.

host/xtest/regression_1000.c
vesajaaskelainen: ok

host/xtest/regression_1000.c
vesajaaskelainen: ok

host/xtest/regression_1000.c
```diff
@@ -2036,3 +2037,190 @@ static void xtest_tee_test_1025(ADBG_Case_t *c)
...
+ * Value here is random UUID that is allocated as name space identifier for
+ * forming Client UUID's for TEE environment using UUIDv5 scheme.
+ */
+static const char *client_uuid_linux_ns = "58ac9ca0-2086-4683-a1b8-ec4bc08e01b6";
+
+/* TEEC_LOGIN_PUBLIC's Client UUID is NIL UUID */
+static TEEC_UUID client_uuid_public = { };
+
+static TEEC_Result get_client_identity(TEEC_Session *session, uint32_t *login,
+				       TEEC_UUID *client_uuid)
```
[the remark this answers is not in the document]
vesajaaskelainen: If moved there I suppose `ta_os_test_cmd_client_identity()` would be more matching?

There is already `ta_crypt_cmd_*` variants and this would fit to that scheme.

host/xtest/xtest_uuid_helpers.h
jforissier: When copying an existing file I don't change the copyrights, I simply add mine if I modify the code.

jforissier:
I'm restarting IBART. It should pass now that https://github.com/OP-TEE/optee_client/pull/195 is merged.

host/xtest/regression_1000.c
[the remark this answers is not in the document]
vesajaaskelainen: Updated

host/xtest/regression_1000.c
vesajaaskelainen: updated

host/xtest/regression_1000.c
vesajaaskelainen: updated

jforissier:
For the two commits: "xtest: regression_1000: add test for session login as user" and "xtest: regression_1000: add test for session login as group":

host/xtest/regression_1000.c
vesajaaskelainen: Updated.

host/xtest/xtest_uuid_helpers.c
vesajaaskelainen: Left as is.

vesajaaskelainen:
> For the two commits: "xtest: regression_1000: add test for session login as user" and "xtest: regression_1000: add test for session login as group":
> 

Thanks for the review!

Updated the PR with those included.

vesajaaskelainen:
I have now tested this with PATCH v2 for kernel changes that is on mailing list. 

Note: What is in Linaro tree there is no practical change in the code.

I do not have access to full test suite so cannot say if there are any effects for that. Hopefully not. Only issue might be that if application based logins are tested and they do not pass as kernel says not supported.

jforissier:
> I do not have access to full test suite so cannot say if there are any effects for that. Hopefully not. 

I have run it on QEMU, all is good. In fact there are no tests with login methods other than `TEEC_LOGIN_PUBLIC`.
```
root@Vexpress:/ xtest
Run test suite with level=0

#
# regression+gp
#

[snip]

747 test cases of which 0 failed
0 test cases were skipped
TEE test application done!
root@Vexpress:/
```

vesajaaskelainen:
> > I do not have access to full test suite so cannot say if there are any effects for that. Hopefully not.
> 
> I have run it on QEMU, all is good. In fact there are no tests with login methods other than `TEEC_LOGIN_PUBLIC`.

Thanks for testing it out!

jforissier:
@etienne-lms @jenswi-linaro gentle reminder in case you'd like to ack before I merge this, thanks.

host/xtest/xtest_uuid_helpers.c
```diff
@@ -6,6 +6,8 @@
 #include <ctype.h>
 #include <stdint.h>
 #include <string.h>
+#include <endian.h>
+#include <openssl/evp.h>
```
etienne-lms, line 10: fix order

vesajaaskelainen:
> @etienne-lms @jenswi-linaro gentle reminder in case you'd like to ack before I merge this, thanks.

I can fix @etienne-lms's comment later today and will include those acks and update the PR.

host/xtest/xtest_uuid_helpers.c
vesajaaskelainen: ok

vesajaaskelainen: @etienne-lms 
Hmm... I was kinda looking for guidance on how it should be done but best that I could find was reference to clang-format within Linux kernel. And that didn't modify include order.

So are you looking order like this:
```
#include <ctype.h>
#include <endian.h>
#include <openssl/evp.h>
#include <stdint.h>
#include <string.h>
#include <tee_api_types.h>
#include <tee_client_api.h>
#include "xtest_uuid_helpers.h"
```

etienne-lms: Yes: alphabetical ordering with `<...>` includes first, then alphabetical ordering with `"..."` includes. Thanks.

vesajaaskelainen:
@jforissier updated the PR with acks and #include reordering.

host/xtest/xtest_uuid_helpers.c
vesajaaskelainen: Fixed.

jforissier:
@vesajaaskelainen cool, I'm merging this, thanks.

etienne-lms:
Now that this is merged... I am wondering if we won't face the same issue as for https://github.com/OP-TEE/optee_test/issues/415: one testing latest OP-TEE (os/client/test) with latest mainline Linux kernel will get its xtest regression tests to fail on these new tests since mainline Linux kernel does not integrate https://github.com/linaro-swg/linux/pull/74.
:(

jforissier:
> one testing latest OP-TEE (os/client/test) with latest mainline Linux kernel will get its xtest regression tests to fail on these new tests

Yes, but it is just a matter of time until the code lands upstream, and it is one of the reasons we have linaro-swg/linux branch optee.
However... now that you're mentioning it... we do have platforms in our manifest that are **not** based on linaro-swg optee...

```
$ git remote -v | grep origin
origin	https://github.com/OP-TEE/manifest.git (fetch)
origin	https://github.com/OP-TEE/manifest.git (push)
$ $ git grep linux | grep -v 'revision="optee'
am43xx.xml:        <project path="linux"          name="ti-linux-kernel/ti-linux-kernel.git" revision="00a3beacce33dc14fa301eb5f3fb5a341212e9b4" remote="ti" />
am57xx.xml:        <project path="linux"          name="ti-linux-kernel/ti-linux-kernel.git" revision="00a3beacce33dc14fa301eb5f3fb5a341212e9b4" remote="ti" />
dra7xx.xml:        <project path="linux"          name="ti-linux-kernel/ti-linux-kernel.git" revision="00a3beacce33dc14fa301eb5f3fb5a341212e9b4" remote="ti" />
poplar.xml:        <project path="linux"                name="96boards-poplar/linux.git"             revision="poplar-4.9" clone-depth="1" />
rpi3.xml:        <project path="linux"                name="linaro-swg/linux.git"                  revision="rpi3-optee-4.14" clone-depth="1" />
verdin.xml:        <project path="linux"                name="igoropaniuk/linux"               revision="mainline_verdin" clone-depth="1" />
```

I will create PRs for Poplar and RPi3. The TI boards are stuck to a fixed revision so not sure there is a branch to update for them. Ping @OP-TEE/plat-ti.
For Verdin: ping @igoropaniuk.

=> The commits to cherry-pick from https://github.com/linaro-swg/linux/commits/optee are:
1. Commit baa151f2924f ("tee: add support for session's client UUID generation")
2. Commit ad19acdcdbc5 ("tee: optee: Add support for session login client UUID generation")

vesajaaskelainen:
PATCH v2 has been sent for those Linux kernel patches in the mailing list about a week ago. Now kernel is in rc4 and in weekend that would be rc5. So probably the earliest we have code in the kernel is 3-4 weeks (for 5.8-rc1) assuming that there are no blocking issues found.

Best would be if we could cherry-pick changes from upstream kernel for those vendor trees.

But I suppose you were planning to make op-tee release in about 3 week timeline which is kinda problematic.

Issue #410, opened 2020-04-08
xtest regression_1018 failing with CFG_CORE_DYN_SHM=n


b49020:
Test-case added via [commit](https://github.com/OP-TEE/optee_test/commit/84382b31d3a3758fcd52f2751bb52ed4138a4281) fails with `CFG_CORE_DYN_SHM=n`.

The reason seems to be allocation from dma_buf shared memory to be page aligned (see [PAGE_SHIFT](https://github.com/linaro-swg/linux/blob/optee/drivers/tee/optee/core.c#L504) as min_alloc_order). This leads to actual allocation of 8192 bytes instead of expected 6001 bytes leading to out of bounds issue being undetected.

So I would like to hear views about how to fix it? Maybe set `min_alloc_order=1 byte`?

jenswi-linaro:
There's two problems here.
The test depends on a certain implementation to get the `TEEC_ORIGIN_COMMS` error origin, this error doesn't have to be caught at this level.

When we're using the reserved memory pool instead we're hitting this code path in Client API:
https://github.com/OP-TEE/optee_client/blob/abc05e61872a9ae2cd08c885c7b523df5ba1a26b/libteec/src/tee_client_api.c#L274-L281

There's a check missing that we're not out of bounds, if there would be such a check the error origin should be `TEEC_ORIGIN_API`.

So the test need to be updated to accept `TEEC_ORIGIN_API` as a valid error origin and Client API need to be update with the missing check.

etienne-lms:
Agree regarding the error origin. I think the test should only ensure the origin is _not_ the TA itself, whatever it being the API, the comms or the TEE.

b49020:
@jenswi-linaro 

> When we're using the reserved memory pool instead we're hitting this code path in Client API:

Its true when we use `TEEC_RegisterSharedMemory` but not hit when we use `TEEC_AllocateSharedMemory` with reserved memory pool.

So in general, I agree with the check for out of bounds in `optee-client` but I think it should be common check [here](https://github.com/OP-TEE/optee_client/blob/abc05e61872a9ae2cd08c885c7b523df5ba1a26b/libteec/src/tee_client_api.c#L273).

jenswi-linaro:
Yeah, and here too: https://github.com/OP-TEE/optee_client/blob/abc05e61872a9ae2cd08c885c7b523df5ba1a26b/libteec/src/tee_client_api.c#L232

b49020:
[...]

PR #411, opened 2020-04-08
regression 1018: out of bounds memref tests ensure TA is not reached

etienne-lms:
Change regression test 1018 to allow any layer if the TEE invocation
channel to report outbound memory reference as long as it is the the
TA itself since outbound memory reference should be caught before the
TA is invoked by the TEE.

This change makes xtest regression test 1018 more flexible towards
changes in the OP-TEE components that can catch invalid memory reference
in invocation parameters depending on the OP-TEE configuration.

Fixes: 84382b31d3a3 ("regression 1018: add 1byte out of bounds reference test sub cases
")
Reported-by: Sumit Garg <sumit.garg@linaro.org>

### DIFF
--- a/host/xtest/regression_1000.c
+++ b/host/xtest/regression_1000.c
@@ -1388,7 +1388,7 @@ static void invoke_1byte_out_of_bounds(ADBG_Case_t *c, TEEC_Session *session,
 	ret = TEEC_InvokeCommand(session, TA_OS_TEST_CMD_PARAMS,
 				 &op, &ret_orig);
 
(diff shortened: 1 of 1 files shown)


etienne-lms:
Note: i did not mention  in commit message the failing test traces reported by https://github.com/OP-TEE/optee_test/issues/410 as this change does not fix the issue, it rather fix regression-1018 when https://github.com/OP-TEE/optee_test/issues/410 is fixed through https://github.com/OP-TEE/optee_client.

etienne-lms:
fixed line break typo in commit message.

jforissier:

etienne-lms:
tags applied

PR #412, opened 2020-04-08
regression 1025: test null memref on pta invocation

etienne-lms:
Add sub test in regression test 1025 for invocation of a PTA with
a NULL memory reference.

### DIFF
--- a/host/xtest/regression_1000.c
+++ b/host/xtest/regression_1000.c
@@ -1917,6 +1917,7 @@ static void xtest_tee_test_1025(ADBG_Case_t *c)
 	uint32_t ret_orig = 0;
 	uint8_t *empty_buf = NULL;
 	TEEC_SharedMemory shm = { };
+	TEEC_Result res = TEEC_ERROR_GENERIC;
 
 	Do_ADBG_BeginSubCase(c, "Invalid NULL buffer memref registration");
 
@@ -1952,7 +1953,7 @@ static void xtest_tee_test_1025(ADBG_Case_t *c)
 	empty_buf = malloc(1);
 	if (!empty_buf) {
 		(void)ADBG_EXPECT_TEEC_SUCCESS(c, TEEC_ERROR_OUT_OF_MEMORY);
-		goto out;
+		goto out_session;
 	}
 
 	op.paramTypes = TEEC_PARAM_TYPES(TEEC_MEMREF_TEMP_INPUT,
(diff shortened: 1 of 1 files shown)


etienne-lms:
@cneveux and all: propose these extra test to check PTA can also be invoked with a NULL memref buffer.
Depends on https://github.com/OP-TEE/optee_os/pull/3733.

Feel free to reject both if considered useless.

host/xtest/regression_1000.c
```diff
@@ -1975,10 +1976,36 @@ static void xtest_tee_test_1025(ADBG_Case_t *c)
 				       TA_OS_TEST_CMD_NULL_MEMREF_PARAMS, &op,
 				       &ret_orig));
 
+	TEEC_CloseSession(&session);
+
 	Do_ADBG_EndSubCase(c, "Input MEMREF Buffer NULL - Size non 0 bytes");
 
-out:
+	Do_ADBG_BeginSubCase(c, "Input MEMREF Buffer NULL over PTA invocation");
```
cneveux, line 1983: Why not starting subtest case if the pta session return success? As you say this is optional, hence if PTA not found there is no subcase to be log, isn't it?

etienne-lms: I use the same convention used for other test PTA cases: test is attempted and a log is emitted if  test PTA is not embedded. Being done from a test or or sub test does not really matter I think. I wanted all null memref test to be from the same test case. Yet no strong opinion on that being in a test or subtest.

cneveux: Fine, but my concern is that: subcase is counted but not run. 
At the end, there is no indication about sub tests not run and so we consider that everything is OK even if not tested.

jforissier: @cneveux I think you have a valid point, that being said, `xtest` reports a summary of the number of "cases" and "subtests", but not the number of "subcases".
- A "case" is a whole test case like `xtest 1025`
- A "subtest" is the slightest thing checked with `ADBG_EXPECT*`. This number is quite meaningless IMO because we have tests running in big loops with concurrency impacts so the numbers are not reproducible.
- A "subcase" is what we're discussing here (`Do_ADBG_BeginSubCase()`).
What I'm saying is, it would be nice to report the number of subcases in the `xtest` summary and do what you recommend (not run subcases that do not apply to a given configuration).

Perhaps a topic for a separate PR?

jenswi-linaro:
> Perhaps a topic for a separate PR?

+1

host/xtest/regression_1000.c
cneveux: Approved.

etienne-lms:
@cneveux tag applied.

xtest fails as this depends on OP-TEE/optee_os#3733.

PR #413, opened 2020-04-08
GP: expect non zero digestLength

jforissier:
OP-TEE commit 2e5e6460eb36 ("libutee: Set digestLength value in
TEE_OperationInfo structure") causes a regression in the GP test suite,
namely: regression tests gp_9052 and gp_9115. These tests expect an
output digestLength of zero, which is surprising actually. I don't have
access to updated test suites from GP some I'm assuming it is a bug in
the test spec. This commit adds new defines for the SHA256 and MD5
digest lengths and updates the patch applied to the test XML.

Tested-by: Jerome Forissier <jerome@forissier.org> (QEMU, GP)


jforissier:
@etienne-lms if you have access to the latest GP test suite, could you check if this makes sense or if I'm missing something?

etienne-lms:
sure. your change makes sense indeed. i wanted to test it before given some feedback.

jforissier:
Update. I forgot to patch a patch, did not notice because I had edited the final file (`gp_9000.c`) manually in my environment but this update is clearly needed for this to work from a clean state.

etienne-lms:
This patch files works fine. Thanks. I confirm gp 9052 and 9115 were failing prior this change and with do successfully pass after this change.
`Tested-by: Etienne Carriere <etienne.carriere@linaro.org> (QEMU, GP)`

I've looked at GPD test suite 2.0.0 (which I unfortunately cannot run): the macro names are `SIZE_DIGEST_MD5_16` and `SIZE_DIGEST_SHA256_32` and it seems the digest sizes are properly expected in these 2 tests. Up to you to uses these macro names or the one proposed here.

jforissier:
Excellent, thanks Étienne, I will use the GPD 2.0.0 names since there's no reason not to.

jforissier:
Update

etienne-lms:
Successfully tested again.

PR #416, opened 2020-04-13
regression 1025: skip test if memref null capability not supported

etienne-lms:
Skip regression test 1025 when TEE context reports that memref null
capability is not supported by the OP-TEE ecosystem. This change
prevents xtest suite to fail on regression 1025 when one tests
the latest optee_os/optee_client/optee_test that supports this recent
capability with a mainline or older Linux kernel that does not yet
support this feature.

Reported-by: Masami Hiramatsu <masami.hiramatsu@linaro.org>

### DIFF
--- a/host/xtest/regression_1000.c
+++ b/host/xtest/regression_1000.c
@@ -1917,6 +1917,11 @@ static void xtest_tee_test_1025(ADBG_Case_t *c)
 	uint32_t ret_orig = 0;
 	uint8_t *empty_buf = NULL;
 
+	if (!xtest_teec_ctx.memref_null) {
+		Do_ADBG_Log("Skip test: MEMREF_NULL capability not supported");
+		return;
+	}
+
 	if (!ADBG_EXPECT_TEEC_SUCCESS(c,
 				      xtest_teec_open_session(&session,
 							      &os_test_ta_uuid,


host/xtest/regression_1000.c
```diff
@@ -1917,6 +1917,11 @@ static void xtest_tee_test_1025(ADBG_Case_t *c)
 	uint32_t ret_orig = 0;
 	uint8_t *empty_buf = NULL;
 
+	if (!xtest_teec_ctx.memref_null) {
```
jenswi-linaro, line 1920: This field is implementation defined and may be changed.
We could have a helper function/macro in `tee_client_api.h` to report this status instead. It is an extension and probably only useful here in xtest, but it will allow us to make changes to `TEEC_Context` without synchronizing with xtest.

jforissier:
In principle `xtest` is not supposed to work with older versions of OP-TEE -- we always expect all Gits are in sync. The rationale being simplicity, there is already enough to take care of with configurable feature not to add the burden of older configurations.
Now, this is also about the kernel version, which might be a bit harder to keep updated for various reasons. So I'm not against this.

jenswi-linaro:

jforissier:
Forgot to add:

etienne-lms:
So far so good. I guess this is the simplest way. Thanks for the feedback.

etienne-lms:
tags applied.

PR #417, opened 2020-04-14
regression_1025: test registering of NULL buffer as shared memory

etienne-lms:
Add a conformance test to check `TEEC_RegisterSharedMemory()` refuses to register
a NULL buffer as a shared memory reference as per GPD TEE specification.

This test currently fails as optee_client needs to be fixed. Fix proposal in https://github.com/OP-TEE/optee_client/pull/204.

### DIFF
--- a/host/xtest/regression_1000.c
+++ b/host/xtest/regression_1000.c
@@ -1916,9 +1916,30 @@ static void xtest_tee_test_1025(ADBG_Case_t *c)
 	TEEC_Operation op = TEEC_OPERATION_INITIALIZER;
 	uint32_t ret_orig = 0;
 	uint8_t *empty_buf = NULL;
+	TEEC_SharedMemory shm = { };
+
+	Do_ADBG_BeginSubCase(c, "Invalid NULL buffer memref registration");
+
+	memset(&shm, 0, sizeof(shm));
+	shm.flags = TEEC_MEM_INPUT;
+	shm.buffer = NULL;
+	shm.size = 0;
+
+	ADBG_EXPECT(c, TEEC_ERROR_BAD_PARAMETERS,
+		    TEEC_RegisterSharedMemory(&xtest_teec_ctx, &shm));
+
+	memset(&shm, 0, sizeof(shm));
+	shm.flags = TEEC_MEM_OUTPUT;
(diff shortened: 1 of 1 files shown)


host/xtest/regression_1000.c
```diff
@@ -1916,9 +1916,30 @@ static void xtest_tee_test_1025(ADBG_Case_t *c)
...
+	shm.size = 0;
+
+	ADBG_EXPECT(c, TEEC_ERROR_BAD_PARAMETERS,
+		    TEEC_RegisterSharedMemory(&xtest_teec_ctx, &shm));
+
+	Do_ADBG_EndSubCase(c, "Invalid NULL buffer memref registration");
 
 	if (!xtest_teec_ctx.memref_null) {
-		Do_ADBG_Log("Skip test: MEMREF_NULL capability not supported");
+		Do_ADBG_Log("Skip sub-tests: MEMREF_NULL capability not supported");
```
jforissier, line 1942: Unrelated change, and should be "subcases" not "sub-tests".

etienne-lms: I think it is related since test case is not skipped, only the next subcases as this change adds a subcases to this test case that is always executed. No?

jforissier: Ah, yes. Then just `s/test/subcases`.

etienne-lms:
s/test/subcases/ fixed.
tag applied.

Tested with OP-TEE/optee_client#204.
With memref_null support:
```
* regression_1025 Test memref NULL and/or 0 bytes size
o regression_1025.1 Invalid NULL buffer memref registration
  regression_1025.1 OK
o regression_1025.2 Input/Output MEMREF Buffer NULL - Size 0 bytes
  regression_1025.2 OK
o regression_1025.3 Input MEMREF Buffer NULL - Size non 0 bytes
  regression_1025.3 OK
  regression_1025 OK
```
Without memref_null support:
```
* regression_1025 Test memref NULL and/or 0 bytes size
o regression_1025.1 Invalid NULL buffer memref registration
  regression_1025.1 OK
Skip subcases: MEMREF_NULL capability not supported
  regression_1025 OK
```

PR #419, opened 2020-04-15
regression 4007_dsa: run all tests at level 0

jforissier:
Some DSA tests are currently run at level 1 or higher (xtest -l 1).
But, since optee_os commit xxxxx, the DSA key generation is much faster
so there is no reason to exclude test cases from the default level.


jforissier:
I'll fill in the commit reference in the description once https://github.com/OP-TEE/optee_os/pull/3754 is merged.

jforissier:
Ping?

jenswi-linaro:

PR #422, opened 2020-05-13
xtest: disable tests 1027 and 1028 if no OpenSSL

vchong:
Tests 1027 and 1028 depend on xtest_uuid_v5() which in turn depends on OpenSSL
APIs so they should just return if OPENSSL_FOUND is not defined.

### DIFF
--- a/host/xtest/regression_1000.c
+++ b/host/xtest/regression_1000.c
@@ -2085,6 +2085,7 @@ ADBG_CASE_DEFINE(regression, 1026, xtest_tee_test_1026,
 
 static void xtest_tee_test_1027(ADBG_Case_t *c)
 {
+#ifdef OPENSSL_FOUND
 	TEEC_Result result = TEEC_ERROR_GENERIC;
 	uint32_t ret_orig = 0;
 	TEEC_Session session = { };
@@ -2125,6 +2126,11 @@ static void xtest_tee_test_1027(ADBG_Case_t *c)
 
 out:
 	TEEC_CloseSession(&session);
+#else /*!OPENSSL_FOUND*/
+	UNUSED(c);
+	/* xtest_uuid_v5() depends on OpenSSL */
+	Do_ADBG_Log("OpenSSL not available, skipping test 1027");
+#endif
 }
 
 ADBG_CASE_DEFINE(regression, 1027, xtest_tee_test_1027,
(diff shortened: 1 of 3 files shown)


host/xtest/regression_1000.c
```diff
@@ -2125,13 +2126,18 @@ static void xtest_tee_test_1027(ADBG_Case_t *c)
 
 out:
 	TEEC_CloseSession(&session);
+#else /*!OPENSSL_FOUND*/
+	UNUSED(c);
```
jforissier, line 2130: I suggest adding a comment `/* xtest_uuid_v5() depends on OpenSSL */` because it's not totally obvious which part of the test depends on it. And same below.

host/xtest/xtest_uuid_helpers.c
```diff
@@ -5,12 +5,15 @@
...
 #include <endian.h>
+#ifdef OPENSSL_FOUND
 #include <openssl/evp.h>
+#endif
 #include <stdint.h>
 #include <string.h>
 #include <tee_api_types.h>
 #include <tee_client_api.h>
 #include "xtest_uuid_helpers.h"
+#include <adbg.h>
```
jenswi-linaro, line 16: This might not be needed.

vchong: Right. This was from a overwritten commit. Will remove.

host/xtest/regression_1000.c
vchong: Will add.

vchong:
> Is there a way we could catch this in CI, by forcing a build without OPENSSL_FOUND somehow?

This doesn't seem so straightforward. :( We have `$(call force,CFG_TA_MBEDTLS,y)` in `optee_os/mk/config.mk`, which in turn forces `CFLAGS += -DOPENSSL_FOUND=1` in `optee_test/host/xtest/Makefile`, which just means it's always enabled anyway in Linux builds. I only found this out because there's no `CFLAGS += -DOPENSSL_FOUND=1` in `Android.mk`.

vchong:
All comments addressed, tags added and forced push. Thanks.

jforissier:
> We have `$(call force,CFG_TA_MBEDTLS,y)` in `optee_os/mk/config.mk`, which in turn forces `CFLAGS += -DOPENSSL_FOUND=1` in `optee_test/host/xtest/Makefile`

Just a quick idea, we could add a condition:
```
WITH_OPENSSL ?= y
ifeq ($(WITH_OPENSSL),y)
... all the OpenSSL CFLAGS / LDFLAGS stuff ...
endif
```
...and then build once normally and once with `make WITH_OPENSSL=n`.
Anyway that would be for a separate PR.

vchong:
> ...and then build once normally and once with make WITH_OPENSSL=n.

I've thought of a local optee_test flag too. I suppose it's ok then to remove the dependency of `OPENSSL_FOUND` on `CFG_GP_PACKAGE_PATH` and `CFG_TA_MBEDTLS`? But then, how do we trigger the build though? In ibart?

PR #423, opened 2020-05-13
xtest: Makefile: create flag to enable OpenSSL

vchong:
As more tests use or depend on OpenSSL, it does not make sense anymore
to enable OpenSSL, i.e. define OPENSSL_FOUND, based on one or several or
all of the 'imported' CFG* flags. Instead, decouple the definition of
OPENSSL_FOUND from CFG* flags and use a local WITH_OPENSSL flag instead.
This will allow CI testing of builds without OPENSSL_FOUND.

Suggested-by: Jerome Forissier <jerome@forissier.org>

### DIFF
--- a/host/xtest/Makefile
+++ b/host/xtest/Makefile
@@ -33,8 +33,12 @@ define cc-bits
 $(if $(filter arm, $(1)),32,$(if $(filter aarch64, $(1)),64,unknown-arch))
 endef
 
-# OpenSSL is used by GP tests series 8500 and Mbed TLS test 8103
-ifneq ($(CFG_GP_PACKAGE_PATH)$(filter y,$(CFG_TA_MBEDTLS)),)
+# OpenSSL is used by:
+# - GP tests series 8500
+# - Mbed TLS test 8103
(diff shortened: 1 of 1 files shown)


host/xtest/Makefile
```diff
@@ -33,8 +33,12 @@ define cc-bits
 $(if $(filter arm, $(1)),32,$(if $(filter aarch64, $(1)),64,unknown-arch))
 endef
 
-# OpenSSL is used by GP tests series 8500 and Mbed TLS test 8103
-ifneq ($(CFG_GP_PACKAGE_PATH)$(filter y,$(CFG_TA_MBEDTLS)),)
+# OpenSSL is used by:
+# - GP tests series 8500
+# - Mbed TLS test 8103
+# - UUID tests 1027 and 1028
```
jforissier, line 39: Minor: "User/group login tests 1027 and 1028"

vchong: Done.

vchong:
Comment addressed and tag added. Thanks!

PR #424, opened 2020-05-16
[RFC] Kernel version restrictions on xtest test cases

etienne-lms:
Change `xtest` to take a new optional argument `-k` to skip specific tests that mandate mainline Linux patches.
Config switch `CFG_XTEST_KERNEL_RESTRICTIONS=y` enforces `-k` feature at built time.

Apply to regression 1027 and 1028: skipped if kernel restriction enable.

Below are the logs for `xtest 1028` and `xtest -k 1028` on qemu_virt booted on a recent linux-next head (5.7-rc4), which does not implement OPTEE client UUID arguments:
```bach
# xtest 1028
Test ID: 1028
Run test suite with level=0

#
# regression
#
 
* regression_1028 Session: group login for current user's effective group
regression_1000.c:2195: memcmp(&expected_client_uuid, &client_uuid, sizeof(TEEC_UUID)) has an unexpected value: 0x28, expected 0x0
  regression_1028 FAILED
Result of testsuite regression filtered by "1028":
regression_1028 FAILED first error at regression_1000.c:2195
6 subtests of which 1 failed
1 test case of which 1 failed
101 test cases were skipped
TEE test application done!
```
Now with argument `-k`:
```bach
# xtest -k 1028
Test restricted to mainline kernel 5.7           <------ note here the new trace message
Test ID: 1028
Run test suite with level=0

#
# regression
#
 
* regression_1028 Session: group login for current user's effective group
- Skip the test case, needs a patched Linux kernel.          <------  here the new trace message
  At least up to Linux kernel v5.7, client UUID is
  not implemented.

  regression_1028 OK
Result of testsuite regression filtered by "1028":
regression_1028 OK
0 subtests of which 0 failed
1 test case of which 0 failed
101 test cases were skipped
TEE test application done!
# 
```

### DIFF
--- a/host/xtest/xtest_main.c
+++ b/host/xtest/xtest_main.c
@@ -18,6 +18,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <sys/utsname.h>
 #include <unistd.h>
 
 #ifdef OPENSSL_FOUND
@@ -48,6 +49,10 @@ ADBG_SUITE_DEFINE(regression);
 char *_device = NULL;
 unsigned int level = 0;
 static const char glevel[] = "0";
+
+/* By default bypass kernel version using UINT_MAX */
+unsigned int xtest_kernel_major = UINT_MAX;
+unsigned int xtest_kernel_minor = UINT_MAX;
 
 #ifdef WITH_GP_TESTS
 #define GP_SUITE	"+gp"
@@ -85,6 +90,11 @@ void usage(char *program)
 	printf("\t                   separated by a '+')\n");
 	printf("\t                   Default value: '%s'\n", gsuitename);
 	printf("\t-h                 Show usage\n");
+	printf("\t-k                 Skip test cases that Linux kernel may not\n");
+	printf("\t                   support. Checks kernel version with uname.\n");
+#ifdef CFG_XTEST_KERNEL_RESTRICTIONS
+	printf("\t                   Force enabled by configuration switch!\n");
+#endif
 	printf("applets:\n");
 	printf("\t--sha-perf [opts]  SHA performance testing tool (-h for usage)\n");
 	printf("\t--aes-perf [opts]  AES performance testing tool (-h for usage)\n");
@@ -108,6 +118,48 @@ static void init_ossl(void)
 #endif
 }
 
+static int set_kernel_restriction(void)
+{
+	struct utsname os_name = { };
+	int major = 0;
+	int minor = 0;
+
+	if (uname(&os_name)) {
+		printf("Error: uname() failed\n");
+		return -1;
+	}
+
+	if (strcmp(os_name.sysname, "Linux")) {
+		fprintf(stderr, "Error: unrecognized uname.sysname %s\n",
+			os_name.sysname);
+		return -1;
+	}
+
+	if (sscanf(os_name.release, "%u.%u.", &major, &minor) != 2) {
+		fprintf(stderr, "Error: unrecognized uname.release %s\n",
+			os_name.release);
+		return -1;
+	}
+
+	if (cmp_kernel_version(major, minor,
+			       KERNEL_MAJOR_MIN, KERNEL_MINOR_MIN) < 0)
+		printf("Warning: version %u.%u too early for OP-TEE\n",
+		       major, minor);
+
+	if (cmp_kernel_version(major, minor,
+			       KERNEL_MAJOR_NEXT, KERNEL_MINOR_NEXT) >= 0)
+		printf("Warning: unknown next kernel %u.%u.\n",
+		       major, minor);
+
+	xtest_kernel_major = major;
+	xtest_kernel_minor = minor;
+
+	printf("Test restricted to mainline kernel %u.%u\n",
+	       xtest_kernel_major, xtest_kernel_minor);
+
+	return 0;
+}
+
 int main(int argc, char *argv[])
 {
 	int opt = 0;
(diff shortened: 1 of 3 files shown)


etienne-lms:
Commit log for 1st commit is fuzzy:
```patch
 For example, consider regression tests 1027 and 1028 that needs
 specific Linux patches for client UUID management, not upstream
-up to 5.7. The test can implement:
+up to 5.7. Assuming v5.8 implements clienr UUIDs, regression
+1027/1028 would implement:

-    if (!xtest_kernel_is_supported(5,7))
+    if (!xtest_kernel_is_supported(5,8))
 	return; /* skip the test */
```
+s/It happen/It happens/

host/xtest/xtest_main.c
```diff
@@ -121,6 +173,7 @@ int main(int argc, char *argv[])
 		.SuiteID_p = NULL,
 		.cases = TAILQ_HEAD_INITIALIZER(all.cases),
 	};
+	int kernel_restrictions = 0;
```
jenswi-linaro, line 176: `s/int/bool/`

etienne-lms:
> I think we can skip CFG_XTEST_KERNEL_RESTRICTIONS, a runtime switch should be enough.

Cmdline argument `-k` is not backward compatible since prior 3.9.0, unsupported arg `-k` makes `xtest` to fail.
Test envs would still need to adapt pre/post `-k` support.

The builtin configuration make it up to the build env to set the support. As for with `CFG_TEE_CORE_EMBED_INTERNAL_TESTS=y` in the Core. It is handy.

jenswi-linaro:
So take it via an environment variable also, `XTEST_KERNEL_RESTRICTIONS` for instance.
You could when run xtest as:
``` bash
env XTEST_KERNEL_RESTRICTIONS=y xtest
```

etienne-lms:
Ok, fair, I'll update thie P-R as you are suggesting.

etienne-lms:
By the way, I don't really like `kernel restriction`.
Maybe s/`kernel_restriction`/`conservative_kernel`/?

jforissier:
Frankly, I don't like this. I find it somewhat complex and confusing (what tests are selected/ignored exactly?), and more importantly, testing the kernel version to detect features is wrong as mentioned again recently by Linus: https://lkml.org/lkml/2020/5/9/475.

If downstream projects have special restrictions like they're running an older kernel, they can deal with that. In this specific case, if I were using an older kernel without REE login support, I would very much prefer to run xtest and ignore 1027 and 1028 specifically. Adding an "exclude" option to `xtest` to reverse the pattern matching would be a better idea IMO. For example: `xtest -x 1027 -x 1028` would run everything but tests that contain the strings `1027` or `1028`.

etienne-lms:
https://github.com/OP-TEE/optee_test/pull/192?

b49020:
How about to follow similar pattern to how Linux test project (LTP) handles such cases with `.min_kver` option? See corresponding documentation [here](https://github.com/linux-test-project/ltp/blob/master/doc/test-writing-guidelines.txt#L632).

b49020:
BTW, thanks @etienne-lms for looking into this.

etienne-lms:
> Adding an "exclude" option to xtest to reverse the pattern matching would be a better idea IMO. For example: xtest -x 1027 -x 1028 would run everything but tests that contain the strings 1027 or 1028.

I've created https://github.com/OP-TEE/optee_test/pull/427 (based on #192). Addresses the issue of excluding test cases from xtest suites based on test IDs and shall wildcard pattern syntax.

jforissier:
In my opinion, excluding tests and pattern matching (instead of substring matching) are two separate topics. They'd better be introduced separately. Adding `-x <substring>` based on substring matching should be very easy.

etienne-lms:
You prefer a basic string search for `-x` rather than a wildcard pattern search?

jforissier:
> You prefer a basic string search for `-x` rather than a wildcard pattern search?

Yes. And we can always add pattern matching later (`xtest -w *2??? -x -w *201?` for example).

etienne-lms:
> And we can always add pattern matching later (xtest -w *2??? -x -w *201? for example).

This example looks overcomplex to me :)
In the meantime, I'll chaneg this P-R as you suggest.

etienne-lms:
updated

jforissier:
> This example looks overcomplex to me :)

Perhaps not a good example ;-) but my point is I prefer giving orthogonal meaning to `-x` ("exclude what follows") and `-w` ("what follows is a pattern, not a substring).

etienne-lms:
Oops, I did not update this P-R, see https://github.com/OP-TEE/optee_test/pull/427.

jforissier:
> How about to follow similar pattern to how Linux test project (LTP) handles such cases with `.min_kver` option? See corresponding documentation [here](https://github.com/linux-test-project/ltp/blob/master/doc/test-writing-guidelines.txt#L632).

That's a nice way of specifying things, but the fundamental issue I think is we should not bind tests  to kernel versions since downstream users can backport features into older kernels.

Consider for instance our official Rpi3 kernel branch: https://github.com/linaro-swg/linux/tree/rpi3-optee-4.14. It is based on 4.14 although it has the REE login capability.

Similarly, in the future we might have configurable features in the OP-TEE driver, or other drivers depending on OP-TEE which could be optional via `CONFIG_` settings. This all makes kernel version checking more or less irrelevant.

etienne-lms:
I'll close this P-R. Initial issue (excluding xtest tests) should be addressed by https://github.com/OP-TEE/optee_test/pull/427.

b49020:
@jforissier Yeah I agree with you that backports could be problematic to handle with `.min_kver`.

PR #427, opened 2020-05-18
xtest: exclude tests with -x option

etienne-lms:
This P-R is based on https://github.com/OP-TEE/optee_test/pull/192.

Introduces **~~wildcards and~~** test exclusions in xtest command line.

### DIFF
--- a/host/xtest/adbg/src/adbg_run.c
+++ b/host/xtest/adbg/src/adbg_run.c
@@ -129,21 +129,27 @@ static int ADBG_RunSuite(
 	Do_ADBG_Log("######################################################");
 
 	TAILQ_FOREACH(case_def, &Runner_p->Suite_p->cases, link) {
-		if (argc > 0) {
-			bool HaveMatch = false;
-			int i = 0;
+		/* Default match if by default all tests are to be run */
+		bool HaveMatch = !argc || !strcmp(argv[0], "-x");
+		int i = 0;
 
-			for (i = 0; i < argc; i++) {
-
-				if (strstr(case_def->TestID_p, argv[i])) {
-					HaveMatch = true;
-					break;
+		for (i = 0; i < argc; i++) {
+			if (!strcmp(argv[i], "-x")) {
+				i++;
+				if (i >= argc) {
+					Do_ADBG_Log("Error: -x <test-id>");
+					return 1;
 				}
+				if (strstr(case_def->TestID_p, argv[i]))
+					HaveMatch = false;
+			} else {
+				if (strstr(case_def->TestID_p, argv[i]))
+					HaveMatch = true;
 			}
-			if (!HaveMatch) {
-				NumSkippedTestCases++;
-				continue;
-			}
+		}
+		if (!HaveMatch) {
+			NumSkippedTestCases++;
+			continue;
 		}
 
 		Case_p = ADBG_Case_New(case_def);
@@ -191,12 +197,20 @@ static int ADBG_RunSuite(
 
 	Do_ADBG_Log("+-----------------------------------------------------");
 	if (argc > 0) {
+		bool test_exclusion = false;
 		int i = 0;
 
-		for (i = 0; i < argc; i++)
+		for (i = 0; i < argc; i++) {
+			if (!strcmp(argv[i], "-x")) {
+				test_exclusion = true;
+				continue;
+			}
 			Do_ADBG_Log(
-				"Result of testsuite %s filtered by \"%s\":",
-				Runner_p->Suite_p->SuiteID_p, argv[i]);
+				"Result of testsuite %s filtered by \"%s%s\":",
+				Runner_p->Suite_p->SuiteID_p,
+				test_exclusion ? "-x " : "", argv[i]);
+			test_exclusion = false;
+		}
 	} else {
 		Do_ADBG_Log("Result of testsuite %s:",
 			    Runner_p->Suite_p->SuiteID_p);
--- a/host/xtest/xtest_main.c
+++ b/host/xtest/xtest_main.c
@@ -67,7 +67,7 @@ void usage(char *program);
 
 void usage(char *program)
 {
-	printf("Usage: %s <options> <test_id>\n", program);
+	printf("Usage: %s <options> [[-x] <test-id>]...]\n", program);
 	printf("\n");
 	printf("options:\n");
 	printf("\t-d <device-type>   TEE device path. Default not set (use any)\n");
@@ -85,6 +85,13 @@ void usage(char *program)
 	printf("\t                   separated by a '+')\n");
 	printf("\t                   Default value: '%s'\n", gsuitename);
 	printf("\t-h                 Show usage\n");
+	printf("\t<test-id>          Add <test-id> to the list of tests to be run.\n");
+	printf("\t                   A substring match is performed. May be specified\n");
+	printf("\t                   several times. If no tests are given, all the\n");
+	printf("\t                   tests are added.\n");
+	printf("\t-x <test-id>       Exclude <test-id> from the list of tests to be\n");
+	printf("\t                   run. A substring match is performed. May be\n");
+	printf("\t                   specified several times.\n");
 	printf("applets:\n");
 	printf("\t--sha-perf [opts]  SHA performance testing tool (-h for usage)\n");
 	printf("\t--aes-perf [opts]  AES performance testing tool (-h for usage)\n");
@@ -96,6 +103,12 @@ void usage(char *program)
 	printf("\t--sdp-basic [opts] Basic Secure Data Path test setup ('-h' for usage)\n");
 #endif
 	printf("\t--stats [opts]     Various statistics ('-h' for usage)\n");
+	printf("\n");
+	printf("Examples:\n");
+	printf("\txtest -t regression 4001 4003\n");
+	printf("\t                   run regression tests 4001 and 4003\n");
+	printf("\txtest -t regression -x 1027 -x 1028\n");
+	printf("\t                   run all regression tests but 1027 and 1028\n");
 	printf("\n");
 }
 
@@ -121,6 +134,8 @@ int main(int argc, char *argv[])
 		.SuiteID_p = NULL,
 		.cases = TAILQ_HEAD_INITIALIZER(all.cases),
 	};
+	bool exclusion = false;
+	size_t last_gen_option = 1;
 
 	opterr = 0;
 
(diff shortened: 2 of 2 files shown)


etienne-lms:
updated: fixed @jforissier address.

jforissier:
Doesn't work? `xtest -w '102?'` runs nothing. `xtest -x '102?'` runs a bunch of tests including 1020 and 1021.

etienne-lms:
> Doesn't work? xtest -w '102?' runs nothing. xtest -x '102?' runs a bunch of tests including 1020 and 1021.

You need to prefix string with a `*`:
`xtest -w '*102?'`

etienne-lms:
`xtest -x '102?'`  excludes nothing, no match.
`xtest -x '*102?'` will exclude `regression_102?`

jforissier:
Doh! You're right, sorry.

etienne-lms:
updated: wildcard pattern removed from `-x`; option `-w` fully removed.

host/xtest/xtest_main.c
```diff
@@ -97,6 +102,12 @@ void usage(char *program)
 #endif
 	printf("\t--stats [opts]     Various statistics ('-h' for usage)\n");
 	printf("\n");
+	printf("Examples:\n");
+	printf("\txtest -t regression 4001 4003\n");
+	printf("\t                   run regression tests 4001 and 4003\n");
+	printf("\txtest -t regression -x 1017 -x 1018\n");
```
jforissier, line 108: s/1017/1027/
s/1018/1028/

Perhaps remove `-t regression` since it is the default? (and above also).

host/xtest/xtest_main.c
```diff
@@ -147,27 +160,51 @@ int main(int argc, char *argv[])
 	else if (argc > 1 && !strcmp(argv[1], "--stats"))
 		return stats_runner_cmd_parser(argc - 1, &argv[1]);
 
-	while ((opt = getopt(argc, argv, "d:l:t:h")) != -1)
+	while ((opt = getopt(argc, argv, "d:l:t:h:k")) != -1) {
```
jforissier, line 163: Extra `:k`

host/xtest/xtest_main.c
```diff
@@ -147,27 +160,51 @@ int main(int argc, char *argv[])
...
 			test_suite = optarg;
+			last_gen_option = optind;
 			break;
 		case 'h':
 			usage(argv[0]);
 			return 0;
+		case '?':
+			if (optopt == 'x') {
+				/*
+				 * -w and -x are not an option processed here,
```
jforissier, line 183: "The -x option is not processed here, it is part of the test IDs"

host/xtest/xtest_main.c
```diff
@@ -67,7 +67,8 @@ void usage(char *program);
 
 void usage(char *program)
 {
-	printf("Usage: %s <options> <test_id>\n", program);
+	printf("Usage: %s <options> [[-x] <test-id>]...]\n", program);
+	printf("       %s <applet> [options...]\n", program);
```
jforissier, line 71: Unrelated change

host/xtest/adbg/src/adbg_run.c
```diff
@@ -130,17 +130,28 @@ static int ADBG_RunSuite(
...
+						match_exclusion = true;
+				} else {
+					check_inclusion = true;
+					if (strstr(case_def->TestID_p, argv[i]))
+						match_inclusion = true;
 				}
 			}
-			if (!HaveMatch) {
+			if (match_exclusion ||
+			    (check_inclusion && !match_inclusion)) {
```
jforissier, line 154: You don't need several booleans. How about this instead?
```diff
diff --git a/host/xtest/adbg/src/adbg_run.c b/host/xtest/adbg/src/adbg_run.c
index 8d4e31f..17adbbf 100644
--- a/host/xtest/adbg/src/adbg_run.c
+++ b/host/xtest/adbg/src/adbg_run.c
@@ -129,21 +129,29 @@ static int ADBG_RunSuite(
        Do_ADBG_Log("######################################################");

        TAILQ_FOREACH(case_def, &Runner_p->Suite_p->cases, link) {
+               bool HaveMatch = false;
+
                if (argc > 0) {
-                       bool HaveMatch = false;
                        int i = 0;

                        for (i = 0; i < argc; i++) {
-
-                               if (strstr(case_def->TestID_p, argv[i])) {
-                                       HaveMatch = true;
-                                       break;
+                               if (!strcmp(argv[i], "-x")) {
+                                       i++;
+                                       if (i >= argc) {
+                                               Do_ADBG_Log("Error: -x ARG");
+                                               return 1;
+                                       }
+                                       if (strstr(case_def->TestID_p, argv[i]))
+                                               HaveMatch = false;
+                               } else {
+                                       if (strstr(case_def->TestID_p, argv[i]))
+                                               HaveMatch = true;
                                }
                        }
-                       if (!HaveMatch) {
-                               NumSkippedTestCases++;
-                               continue;
-                       }
+               }
+               if (!HaveMatch) {
+                       NumSkippedTestCases++;
+                       continue;
                }

                Case_p = ADBG_Case_New(case_def);
```

WIth the advantage that order becomes significant which gives more possibilities to include and exclude tests.

jforissier:
The title of the PR is now wrong

host/xtest/adbg/src/adbg_run.c
etienne-lms: I understand your sequence but it fails when only `-x ID` are provided or when no ID is passed.
I'll see how to fix that.

host/xtest/xtest_main.c
etienne-lms: ok, will discard.

host/xtest/xtest_main.c
```diff
@@ -85,6 +86,10 @@ void usage(char *program)
 	printf("\t                   separated by a '+')\n");
 	printf("\t                   Default value: '%s'\n", gsuitename);
 	printf("\t-h                 Show usage\n");
+	printf("\t<test-id>          The name(s) of the tests to be run. A substring match is\n");
+	printf("\t                   performed.\n");
+	printf("\t-x <test-id>       The name(s) of the tests to not be run. A substring\n");
```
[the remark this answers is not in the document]
etienne-lms: ok

host/xtest/xtest_main.c
```diff
@@ -85,6 +86,10 @@ void usage(char *program)
 	printf("\t                   separated by a '+')\n");
 	printf("\t                   Default value: '%s'\n", gsuitename);
 	printf("\t-h                 Show usage\n");
+	printf("\t<test-id>          The name(s) of the tests to be run. A substring match is\n");
```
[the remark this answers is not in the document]
etienne-lms: ok

host/xtest/xtest_main.c
etienne-lms: for the example, we need `-t regression` to restrict added tests to the regression test suite only.

host/xtest/xtest_main.c
etienne-lms: oops. thanks.

host/xtest/xtest_main.c
etienne-lms: ok

host/xtest/adbg/src/adbg_run.c
jforissier: Ah yes. That should fix the problematic cases:
```diff
diff --git a/host/xtest/adbg/src/adbg_run.c b/host/xtest/adbg/src/adbg_run.c
index 17adbbf..9ef99ce 100644
--- a/host/xtest/adbg/src/adbg_run.c
+++ b/host/xtest/adbg/src/adbg_run.c
@@ -129,7 +129,7 @@ static int ADBG_RunSuite(
 	Do_ADBG_Log("######################################################");
 
 	TAILQ_FOREACH(case_def, &Runner_p->Suite_p->cases, link) {
-		bool HaveMatch = false;
+		bool HaveMatch = !argc || !strcmp(argv[0], "-x");
 
 		if (argc > 0) {
 			int i = 0;
```

etienne-lms:
Updated.

host/xtest/adbg/src/adbg_run.c
etienne-lms: I just saw your latest proposal...  P-R updated accordingly.

etienne-lms:
Fixup commits squashed.
Tag applied.
Rephrased the commit message.

b49020:
It sounds like a good option to exclude test-cases. So FWIW,

`Acked-by: Sumit Garg <sumit.garg@linaro.org>`

In order to make this option usable, do we have a mapping where user can easily detect which test-case belongs to which kernel feature (or patches present in [upstream-tee-subsys-patches.txt](https://github.com/linaro-swg/linux/blob/optee/upstream-tee-subsys-patches.txt))?

jforissier:
> In order to make this option usable, do we have a mapping where user can easily detect which test-case belongs to which kernel feature (or patches present in [upstream-tee-subsys-patches.txt](https://github.com/linaro-swg/linux/blob/optee/upstream-tee-subsys-patches.txt))?

Would a comment near the test case be good enough? For example:
```C
/*
 * regression_1025
 * Depends on kernel commit "driver: tee: Handle NULL pointer indication from client"
 * Linaro tree: https://github.com/linaro-swg/linux/commit/1149e7d06a00
 * Upstream: <put sha-1 here when known>
 */
```

b49020:
> Would a comment near the test case be good enough?

Yes, I think that could be a good starting point to make user aware of kernel dependency patch which could be the reason for its failure. Although, the example you mentioned (`regression_1025`) passes even with kernel patch missing, probably something to do with the test-case itself.

jforissier:
Passes or is ignored? 1027 and 1028 are perhaps better examples.

b49020:
> Passes or is ignored?

I got `regression_1025 OK` while my kernel tree was missing `driver: tee: Handle NULL pointer indication from client` patch.

jforissier:
@b49020 I have this:
```
* regression_1025 Test memref NULL and/or 0 bytes size
o regression_1025.1 Invalid NULL buffer memref registration
  regression_1025.1 OK
Skip subcases: MEMREF_NULL capability not supported
  regression_1025 OK
```

b49020:
@jforissier yes you are correct, `regression_1025` is skipped in case kernel patch is missing. I was just confused by the end result.

@etienne-lms yes I agree with you regarding no feedback available for regression 1027 and 1028.

So we can go ahead with this PR and comments as suggested by @jforissier.

etienne-lms:
tags applied.

etienne-lms:
> So we can go ahead with this PR and comments as suggested by @jforissier.

I think you refer to jerome's comments https://github.com/OP-TEE/optee_test/pull/427#issuecomment-630180482.
Please see https://github.com/OP-TEE/optee_test/pull/429.

jenswi-linaro:
I think this should be included in the release.

jforissier:
@jenswi-linaro agreed. I have a small list of PRs I'd like to propose for the release, too. I will write them to the changelog PR. In the mean time since we agree on this one, I'm merging it.

Thanks.

PR #429, opened 2020-05-19
xtest: inline description of test dependencies

etienne-lms:

### DIFF
--- a/host/xtest/regression_1000.c
+++ b/host/xtest/regression_1000.c
@@ -2083,6 +2083,15 @@ out:
 ADBG_CASE_DEFINE(regression, 1026, xtest_tee_test_1026,
 		 "Session: public login");
 
+/*
+ * regression_1027
+ * Depends on OpenSSL
+ * Depends on kernel commit "tee: optee: Add support for session login client UUID generation"
+ * Linaro tree: https://github.com/linaro-swg/linux/commit/ad19acdcdbc5
+ * Upstream: <put sha-1 here when known>
+ *
+ * xtest skips the test when not built with OpenSSL.
+ */
 static void xtest_tee_test_1027(ADBG_Case_t *c)
 {
 #ifdef OPENSSL_FOUND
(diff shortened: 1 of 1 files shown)


etienne-lms:
1st commit is related to discussion thread https://github.com/OP-TEE/optee_test/pull/427#issuecomment-630180482.
The 2nd commit is an kind of extension for all other tests that can be skipped based on unsatisfied dependencies from TEE resources.

b49020:
`Acked-by: Sumit Garg <sumit.garg@linaro.org>`

etienne-lms:
Tag applied.

host/xtest/regression_4000.c
```diff
@@ -5549,6 +5549,11 @@ static void xtest_tee_test_4013(ADBG_Case_t *c)
 ADBG_CASE_DEFINE(regression, 4013, xtest_tee_test_4013,
 		"Test generation of device unique TA keys");
 
+/*
+ * regression_4014
+ * Depends on SM2 KEP support in OP-TEE OS.
```
jforissier, line 5554: True, and sub-cases in 4006 depend on SM2 PKE or SM2 DSA yet you're not mentioning that... Similarly 4001 will test SM3 if available.

In fact I don't see a reason to add all these comments except for the kernel-related ones to help identify which patches exactly are needed.

etienne-lms: I agree :) that's why I split into 2 commits.
I propose to discard the 2nd commit and keep only the 1st one. Ok?

jforissier: > I agree :) that's why I split into 2 commits.

Aha true I missed that :) it's good.

> I propose to discard the 2nd commit and keep only the 1st one. Ok?

Sounds good.

jenswi-linaro:
Same here for "xtest: regression 1027, 1028: add a comment on test dependencies ":

etienne-lms:
Rephrased commit message: 
```patch
-Add an inline comment on dependencies on regression tests 1027 and 1028.
-When these dependencies are not satisfied, xtest fails.
+Add an inline comments on kernel patch dependencies on regression tests
+1027 and 1028. When these dependencies are not satisfied, xtest fails.

```
Tags applied.

PR #426, opened 2020-05-16
xtest: regression 1027: fix build warning when openssl not used

etienne-lms:
Fix build warning reported with trace message like below:

.../optee-test-custom/host/xtest/regression_1000.c:2049:20: error: ‘client_uuid_linux_ns’ defined but not used [-Werror=unused-variable]
 static const char *client_uuid_linux_ns = "58ac9ca0-2086-4683-a1b8-ec4bc08e01b6";
                    ^~~~~~~~~~~~~~~~~~~~

Fixes: 8e070bc41e1d ("xtest: disable tests 1027 and 1028 if no OpenSSL")
Signed-off-by: Etienne Carriere <etienne.carriere@linaro.org>

PR #430, opened 2020-05-19
xtest: fix unused variable warning

vchong:
The client_uuid_linux_ns variable is only used in tests 1027 and 1028,
which are compiled only if OPENSSL_FOUND is defined, so the variable
itself needs to be enclosed too with the OPENSSL_FOUND ifdef to avoid
unused variable warnings.

Reported-by: Dmitry Shmidt <dimitrysh@google.com>


etienne-lms:
Addressed through https://github.com/OP-TEE/optee_test/pull/426.

vchong:
> Addressed through #426.

Oops, sorry I missed that. Thanks!

PR #431, opened 2020-05-25
xtest: fix generic argument handling

etienne-lms:
Fix position of last generic argument badly handled that makes -d, -l or
-l argument being treated also as if it was the first test ID to consider
in the filtered test IDs list.

Fixes: 66542a1bb4e0 ("xtest: exclude tests with -x")


jforissier:
BTW the `-d` option doesn't seem to work:
```
$ xtest -d /dev/tee0
Run test suite with level=0

TEE test application started with device [/dev/tee0]
Failed to open TEE context: 0xffff0008
```

etienne-lms:
> BTW the -d option doesn't seem to work:

Looking backward, it seems to me this was not supported since OP-TEE 2.0.0. Since this tag, argument `-d NAME` is expected to be used only with `NAME` set to `optee-tz`. Actually, `-d` does not define a device file but a TEE identifier, as per GPC TEE Client API spec. I think we should fix `xtest` usage to state that. It is optee_client libteec that defines valid TEE identifiers.

jforissier:
OK thanks for looking that up. Would anything of value be lost if we just removed the argument? :wink:
If we have no way to test it, it does not really make sense to keep it, does it?

etienne-lms:
> OK thanks for looking that up. Would anything of value be lost if we just removed the argument? wink
If we have no way to test it, it does not really make sense to keep it, does it?

As the argument is a valid argument to `TEEC_InitializeContext()`, I think we should keep this `xtest` argument.

I've applied your tag on the reviewed commit.
I've appended another commit for the `-d` case. If you prefer I create a specific P-R for it, tell me.

jforissier:
> As the argument is a valid argument to `TEEC_InitializeContext()`, I think we should keep this `xtest` argument.

OK, I agree. Perhaps `_device` should be changed to `tee_name;` in the second patch, but anyway it is less confusing now so for this patch also:

Thanks.

etienne-lms:
> OK, I agree. Perhaps `_device` should be changed to `tee_name` in the second patch...

You're right. I prefer to address this now despite you acked the change, so that we're done with that.

etienne-lms:
Tag applied on 2nd commit.
I appended another commit for renaming `_device` to `xtest_tee_name`. It impacts GPD test suite resources so I preferred it as a specific commit.

jforissier:
Commit "xtest: rename _device to xtest_tee_name to avoid confusion" is:

`Tested-by: Jerome Forissier <jerome@forissier.org> (QEMU, GP)`

etienne-lms:
Tags applied of last commit. Thanks for cross-testing the changes.
All commits tagged.

PR #432, opened 2020-05-26
xtest regression 1006: add tests on binary block properties

etienne-lms:
Added tests related to https://github.com/OP-TEE/optee_os/pull/3886.

### DIFF
--- a/ta/os_test/os_test.c
+++ b/ta/os_test/os_test.c
@@ -51,6 +51,54 @@ static TEE_Result check_returned_prop(
 	return TEE_SUCCESS;
 }
 
+static TEE_Result check_binprop_ones(size_t size, uint8_t *bbuf, size_t bblen)
+{
+	uint8_t ones[4] = { 0xff, 0xff, 0xff, 0xff };
+
+	if (size > 4 || bblen != size) {
+		EMSG("Size error (size=%zu, bblen=%zu)", size, bblen);
+		return TEE_ERROR_GENERIC;
+	}
+	if (strncmp(bbuf, ones, bblen)) {
+		EMSG("Unexpected content");
+		DHEXDUMP(bbuf, bblen);
+		return TEE_ERROR_GENERIC;
+	}
+	return TEE_SUCCESS;
+}
+
+static TEE_Result get_binblock_property(TEE_PropSetHandle h,
+					char *nbuf,
+					uint8_t **bbuf,
+					size_t *bblen)
+{
+	TEE_Result res = TEE_ERROR_GENERIC;
+
+	*bbuf = NULL;
+	*bblen = 0;
+	res = TEE_GetPropertyAsBinaryBlock(h, NULL, *bbuf, bblen);
+
+	if (res == TEE_SUCCESS && !*bblen)
+		return TEE_SUCCESS;
+
+	if (res != TEE_ERROR_SHORT_BUFFER) {
+		EMSG("TEE_GetPropertyAsBinaryBlock() size query returned 0x%x",
+		     (unsigned int)res);
+		return res ? res : TEE_ERROR_GENERIC;
+	}
+
+	*bbuf = TEE_Malloc(*bblen, TEE_MALLOC_FILL_ZERO);
+	if (!bbuf)
+		return TEE_ERROR_OUT_OF_MEMORY;
+
+	res = TEE_GetPropertyAsBinaryBlock(h, NULL, *bbuf, bblen);
+	if (res != TEE_SUCCESS)
+		EMSG("TEE_GetPropertyAsBinaryBlock(\"%s\") returned 0x%x",
+		     nbuf, (unsigned int)res);
+
+	return res;
+}
+
 static TEE_Result print_properties(TEE_PropSetHandle h,
 				   TEE_PropSetHandle prop_set,
 				   struct p_attr *p_attrs, size_t num_p_attrs)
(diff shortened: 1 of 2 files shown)


etienne-lms:
I expect IBART to fail as this P-R depends on https://github.com/OP-TEE/optee_os/pull/3886.

ta/os_test/os_test.c
```diff
@@ -51,6 +51,90 @@ static TEE_Result check_returned_prop(
...
+
+	*bbuf = NULL;
+	*bblen = 0;
+	res = TEE_GetPropertyAsBinaryBlock(h, NULL, *bbuf, bblen);
+
+	if (res == TEE_SUCCESS && !*bblen)
+		return TEE_SUCCESS;
+
+	if (res != TEE_ERROR_SHORT_BUFFER) {
+		EMSG("TEE_GetPropertyAsBinaryBlock() szie query returned 0x%x\n",
```
jforissier, line 129: size

ta/os_test/os_test.c
```diff
@@ -51,6 +51,90 @@ static TEE_Result check_returned_prop(
...
+		     (unsigned int)res);
+		return TEE_ERROR_GENERIC;
```
jforissier, line 131: return res?

ta/os_test/os_test.c
```diff
@@ -51,6 +51,90 @@ static TEE_Result check_returned_prop(
...
+					size_t *bblen)
+{
+	TEE_Result res = TEE_ERROR_GENERIC;
+
+	*bbuf = NULL;
+	*bblen = 0;
+	res = TEE_GetPropertyAsBinaryBlock(h, NULL, *bbuf, bblen);
+
+	if (res == TEE_SUCCESS && !*bblen)
+		return TEE_SUCCESS;
```
jforissier, line 126: This case would be an existing, zero-sized property, right? It is not currently tested, should it be added?

ta/os_test/os_test.c
```diff
@@ -234,40 +320,47 @@ while (true) {
...
-				if (res != TEE_SUCCESS) {
-					EMSG(
-					"TEE_GetPropertyAsBinaryBlock(\"%s\") returned 0x%x\n",
-					nbuf, (unsigned int)res);
-					return res;
+			res = get_binblock_property(h, nbuf, &bbuf, &bblen);
+			if (res)
+				return res;
+
+			if (strcmp("myprop.binaryblock", nbuf) == 0) {
```
jforissier, line 335: Minor: `!strcmp())`

ta/os_test/os_test.c
```diff
@@ -51,6 +51,90 @@ static TEE_Result check_returned_prop(
...
+	case 4:
+		EMSG("\t%#"PRIx8" %#"PRIx8" %#"PRIx8" %#"PRIx8,
+		     bbuf[0], bbuf[1], bbuf[2], bbuf[3]);
+		break;
+	default:
+		TEE_Panic(0);
+	}
+
+	return TEE_ERROR_GENERIC;
+}
```
jforissier, line 112: This is quite complex... how about:
```C
static TEE_Result check_binprop_ones(size_t size, uint8_t *bbuf, size_t bblen)  
{                                                                               
	uint8_t ones[4] = { 0xff, 0xff, 0xff, 0xff };                           

	if (size > 4 || bblen != size) {                                                    
		EMSG("Size error (size=%zu, bblen=%zu)", size, bblen);
		return TEE_ERROR_GENERIC;                                       
	}                                                                       
	if (strncmp(bbuf, ones, bblen)) {                                       
		EMSG("Unexpected content");                                     
		DHEXDUMP(bbuf, bblen);                                          
		return TEE_ERROR_GENERIC;                                       
	}                                                                       
	return TEE_SUCCESS;                                                     
}                                                                               
```

etienne-lms: +1. I'll update the P-R accordingly.

ta/os_test/os_test.c
etienne-lms: This is (should be) tested by the 3 properties here: https://github.com/OP-TEE/optee_test/pull/432/files#diff-a518b40775b076c81be0dbf0be040d2bR41-R46.

ta/os_test/os_test.c
etienne-lms: thanks

ta/os_test/os_test.c
etienne-lms: ok

ta/os_test/os_test.c
jforissier: Duh, I reviewed only the first patch ;-) that is fine.

ta/os_test/os_test.c
etienne-lms: If `res == TEE_SUCCESS`, we must report another return code, hence i think `TEE_ERROR_GENERIC` would be fine. 
Maybe `return res ? res : TEE_ERROR_GENERIC;` ?

jforissier: Ah you're right, no don't bother -- the original code is fine.

ta/os_test/os_test.c
```diff
@@ -36,6 +36,14 @@ static TEE_Result check_returned_prop(
 		TEE_Result return_res, TEE_Result expected_res,
 		uint32_t return_len, uint32_t expected_len)
 {
+	/*
+	 * If expecting TEE_ERROR_SHORT_BUFFER but returned size matches
+	 * expected size then TEE_SUCCESS is a valid return code.
+	 */
+	if (expected_res == TEE_ERROR_SHORT_BUFFER &&
+	    return_res == TEE_SUCCESS && return_len <= expected_len)
+		return TEE_SUCCESS;
```
jforissier, line 45: Why not change the caller instead? The caller knows if the size is big enough or not.

etienne-lms: Ok, I agree i would be better.

etienne-lms:
Updated.
IBART should succeed (OP-TEE/optee_os#3886 is merged).

PR #439, opened 2020-08-14
os_test: C++: disable profiling (-pg)

jforissier:
os_test includes C++ exception tests which are incompatible with ftrace
by design [1]. This commit makes sure the C++ tests are not compiled
with the profiling flag (-pg), which is the flag that one would use to
enable function tracing in TAs.

Link: [1] https://github.com/OP-TEE/optee_os/issues/4022


jforissier:
Depends on https://github.com/OP-TEE/optee_os/pull/4034.

PR #442, opened 2020-08-19
Disable TLS/C++ tests with some unsupported compilers

jforissier:
Addresses https://github.com/OP-TEE/optee_os/pull/4008#issuecomment-675512429.

### DIFF
--- a/host/xtest/regression_1000.c
+++ b/host/xtest/regression_1000.c
@@ -2202,6 +2202,7 @@ ADBG_CASE_DEFINE(regression, 1028, xtest_tee_test_1028,
 
 static void xtest_tee_test_1029(ADBG_Case_t *c)
 {
+	TEEC_Result res = TEEC_SUCCESS;
 	TEEC_Session session = { 0 };
 	uint32_t ret_orig = 0;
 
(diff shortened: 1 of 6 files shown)


jenswi-linaro:

host/xtest/regression_1000.c
```diff
@@ -2211,15 +2212,21 @@ static void xtest_tee_test_1029(ADBG_Case_t *c)
 		return;
 
 	Do_ADBG_BeginSubCase(c, "TLS variables (main program)");
-	ADBG_EXPECT_TEEC_SUCCESS(c,
-		TEEC_InvokeCommand(&session, TA_OS_TEST_CMD_TLS_TEST_MAIN, NULL,
-				   &ret_orig));
+	res = TEEC_InvokeCommand(&session, TA_OS_TEST_CMD_TLS_TEST_MAIN, NULL,
+				 &ret_orig);
+	if (res == TEEC_ERROR_NOT_SUPPORTED)
+		printf("TA returned TEEC_ERROR_NOT_SUPPORTED\n");
```
etienne-lms, line 2218: `Do_ADBG_Log()` instead of `printf()`.

For consistency with other skipped tested:
`Do_ADBG_Log(" - 1029 -   skip test, Thread Local Storage test not supported");`

jforissier: OK, I will use `Do_ADBG_Log()` and adjust the format for more consistency. But I'd rather not repeat "Thread Local Storage" which is printed immediately before. It's better to inform the user that the cause is with the TA I think.

jforissier:
Update

etienne-lms:

PR #443, opened 2020-08-25
Revert "os_test: arm64: disable C++ and TLS tests depending on compiler"

jforissier:
This reverts commit 30efcbeaf8864d0f2a5c4be593a5411001fab31b.

Now that R_AARCH64_TLSDESC relocations are implemented in OP-TEE [1]
there is no need to make an exception for older GCC versions.

[1] https://github.com/OP-TEE/optee_os/commit/7bc927fa5192f4af4f59072f38822e0082e4bddb


jenswi-linaro:

PR #444, opened 2020-08-26
regression 6020: fix fs_open() flags

jenswi-linaro:
The flag TEE_DATA_FLAG_OVERWRITE is not permitted with the
TEE_OpenPersistentObject() functions. So remove the flag where needed in
the 6020 case.


jforissier (approved):

jenswi-linaro:
Tag applied

PR #445, opened 2020-09-04
regression: add 1032 Register read-only shared memory

jenswi-linaro:
Adds a case 1032 for registering read-only shared memory. The Linux
kernel normally refuses to register memory that is read-only so TEE
Client API has to have a fallback option to handle this. This test case
verifies that this fallback option is in place.


host/xtest/regression_1000.c
```diff
@@ -2316,3 +2316,26 @@ static void xtest_tee_test_1031(ADBG_Case_t *c)
 ADBG_CASE_DEFINE(regression, 1031, xtest_tee_test_1031,
 		 "Test C++ features");
 #endif
+
+static void xtest_tee_test_1032(ADBG_Case_t *c)
+{
+	TEEC_Result res = TEEC_SUCCESS;
+	TEEC_Context ctx = { };
+	TEEC_SharedMemory shm = {
+		.buffer = xtest_tee_test_1032,
```
jforissier, line 2325: This is read-only but also executable. How about also testing an initialized static const buffer (thus in the `.rodata` section)?

jenswi-linaro: OK

jenswi-linaro:
Fixed up the commit with the suggested test. Tag applied.

etienne-lms:
@jenswi-linaro, I think you forgot to update the test :)

jenswi-linaro:
> @jenswi-linaro, I think you forgot to update the test :)

Sorry, which part?

etienne-lms:
Sorry, I'm wrong.

jenswi-linaro:
No worries.
Tag applied.

PR #451, opened 2020-10-08
Fix compilation of regression_7000.c + include xml/

clementfaure:


jforissier:
At some point and to simplify maintenance we might be able to remove the regular Makefiles, and use only `Android.mk` (for AOSP builds) and `CMakeLists.txt` (for the Buildroot environment). Anyways,

clementfaure:
@jforissier Thanks for the info
Tags applied!

clementfaure:
Tag applied

PR #454, opened 2020-10-27
xtest: remove __unused and do not include <compiler.h>

jforissier:
xtest (the 'host' part, not the TAs) uses __unused from OP-TEE's
<compiler.h>. This does not comply with the following statement from
the C specification:

 In addition to the names documented in this manual, reserved names
 include all external identifiers (global functions and variables)
 that begin with an underscore (‘_’) and all identifiers regardless of
 use that begin with either two underscores or an underscore followed
 by a capital letter are reserved names. This is so that the library
 and header files can define functions, variables, and macros for
 internal purposes without risk of conflict with names in user
 programs.

This is not a hypothetical problem, __unused is used by musl and as
optee-test can't be built with musl.

<compiler.h> and __unused should only be used in TAs (in which case we
know the C library is OP-TEE's libutils). Remove them from xtest and add
-Wno-unused-param to the C flags instead.

Reported-by: Ross Burton <ross.burton@arm.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,6 +20,7 @@ add_compile_options (
 	-Wswitch-default -Wunsafe-loop-optimizations
 	-Wwrite-strings -Werror -fPIC
  	-Wno-missing-field-initializers
+	-Wno-unused-parameter
 )
 
 find_program(CCACHE_FOUND ccache)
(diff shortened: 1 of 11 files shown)


jforissier:
@rossburton feedback welcome!
I took the liberty of adding a `Reported-by:` tag with your name and email address, if you prefer not to be cited just let me know.

jenswi-linaro:

rossburton:
Looks good to me.

PR #455, opened 2020-10-29
xtest: remove some warnings to fix builds with Clang

### LINKED ISSUE #452: Fails to build with clang
Clang doesn't support `-Wmissing-noreturn` or `-Wno-unsafe-loop-optimizations` but these are both passed unconditionally in `host/xtest/Makefile`.

rossburton:
Two compiler flags are GCC-specific and cause build failures with Clang.

-Wno-unsafe-loop-optimizations was added by 5a83a50 specifically for GCC 7.x.
As GCC 9 is common now, this can be removed.

-Wmissing-noreturn is of negliable benefit and doesn't show actual problems
in the code.

Fixes #452.

Signed-off-by: Ross Burton <ross.burton@arm.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,7 @@ add_compile_options (
 	-Wfloat-equal -Wformat-nonliteral -Wformat-security
 	-Wformat=2 -Winit-self -Wmissing-declarations
 	-Wmissing-format-attribute -Wmissing-include-dirs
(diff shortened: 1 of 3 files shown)


jenswi-linaro:

jforissier:
The flags are in `CMakelists.txt` and `host/xtest/CMakeLists.txt`, too. These files should be updated as well.

jforissier:
LGTM, thanks @rossburton.

jforissier:
Merged, thanks.

PR #456, opened 2020-11-01
dummy: do not review, do not merge, just for testing IBART

jbech-linaro:
This is just a dummy to try to trigger the IBART GP enablement.


jbech-linaro:
IBART with GP (for Hikey) seems to run just fine! I'll keep this open in case I need to push any other IBART changes.

PR #462, opened 2020-11-06
xtest: sort cases to use a well defined order

jenswi-linaro:
Prior to this patch was the order of the cases depending on link order
and in which order constructors are called inside an object file. To use
a well defined order, sort the cases while registering them.


jenswi-linaro:
I'm not sure what happened in `regression_6018.3`.

etienne-lms:

jforissier:

jenswi-linaro:
Tags applied

PR #463, opened 2020-11-10
gp: remove leftovers from GP test suite v1.1.0.4 support

jforissier:
Since f7c37e7c29eb ("gp: update compliance suite to version 2.0.0.2") it
is not possible to build the GlobalPlatform TEE Initial Configuration
Test Suite v1.1.0.4. The test suite was upgraded and the build system
switched from GNU make to CMake.

This commit removes some leftovers in the Makefiles from the previous
configuration.


jforissier:
@jenswi-linaro am I missing something here?

host/xtest/Makefile
```diff
@@ -100,16 +100,6 @@ CFLAGS += -I$(OPTEE_PKCS11_TA_HEADERS)
...
-ifdef CFG_GP_PACKAGE_PATH
-CFLAGS += -DWITH_GP_TESTS
-
-srcs  += gp_7000.c \
-	 gp_7500.c \
-	 gp_8000.c \
-	 gp_8500.c \
-	 gp_9000.c
-endif
-
```
jenswi-linaro, line 112: The other `ifdef CFG_GP_PACKAGE_PATH` and the `ifndef CFG_GP_PACKAGE_PATH` below can also be removed.

jforissier: How could I miss them...? Fixed now.

jenswi-linaro:
Thanks for cleaning up after me. :-)

Issue #458, opened 2020-11-02
Fails to link with musl


rossburton:
optee-test 3.11.0 when built with the musl C library instead of glibc:
```
aarch64-poky-linux-musl-ld.bfd -e__ta_entry -pie -T [...]/build/ta/os_test/ta.lds -Map=[...]/build/ta/os_test/5b9e0e40-2636-11e1-ad9e-0002a5d5c51b.map --sort-section=alignment -z max-page-size=4096  --as-needed   --dynamic-list [...]/build/ta/os_test/dyn_list --eh-frame-hdr  [...]/build/ta/os_test/init.o [...]/build/ta/os_test/os_test.o [...]/build/ta/os_test/ta_entry.o [...]/build/ta/os_test/test_float_subj.o [...]/build/ta/os_test/cxx_tests.o [...]/build/ta/os_test/cxx_tests_c.o [...]/build/ta/os_test/user_ta_header.o -L[...]/build/ta/os_test_lib -los_test -ldl -L[...]/recipe-sysroot/usr/include/optee/export-user_ta/lib --start-group -lutils -lutee -lmbedtls -ldl [...]/recipe-sysroot/usr/lib/libstdc++.a [...]/recipe-sysroot/usr/lib/aarch64-poky-linux-musl/10.2.0/libgcc_eh.a --end-group [...]/recipe-sysroot/usr/lib/aarch64-poky-linux-musl/10.2.0/libgcc.a -lutils -o [...]/build/ta/os_test/5b9e0e40-2636-11e1-ad9e-0002a5d5c51b.elf

aarch64-poky-linux-musl-ld.bfd: [...]/libstdc++.a(eh_alloc.o): in function `(anonymous namespace)::pool::free(void*) [clone .constprop.0]':
[...]/libstdc++-v3/include/aarch64-poky-linux-musl/bits/gthr-default.h:749: undefined reference to `pthread_mutex_lock'
[...]/libstdc++-v3/include/aarch64-poky-linux-musl/bits/gthr-default.h:779: undefined reference to `pthread_mutex_unlock'
```

(repeated in various places)

jenswi-linaro:
I looks like the musl C library doesn't support pthreads. Anything special needed to enable that with musl?

rossburton:
musl does support pthreads, it just integrates the functions directly into libc instead of having a separate library.

jenswi-linaro:
Why the undefined references then?

rossburton:
[...]

PR #464, opened 2020-11-10
Optional C++ and TLS tests

jforissier:
Addressing https://github.com/OP-TEE/optee_test/issues/458.

### DIFF
--- a/ta/os_test/sub.mk
+++ b/ta/os_test/sub.mk
@@ -1,5 +1,6 @@
 cppflags-$(CFG_TA_FLOAT_SUPPORT) += -DCFG_TA_FLOAT_SUPPORT=1
 cppflags-$(WITH_TLS_TESTS) += -DWITH_TLS_TESTS=1
+cppflags-$(WITH_CXX_TESTS) += -DWITH_CXX_TESTS=1
 
 global-incdirs-y += include
 global-incdirs-y += ../crypt/include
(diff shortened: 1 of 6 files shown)


jenswi-linaro:

jforissier:
I have `s/CPLUSPLUS/CXX/` in the second patch (see https://github.com/OP-TEE/build/pull/441). Tag applied to first patch only.

ta/os_test/sub.mk
```diff
@@ -8,7 +10,7 @@ srcs-y += init.c
 srcs-y += os_test.c
 srcs-y += ta_entry.c
 srcs-$(CFG_TA_FLOAT_SUPPORT) += test_float_subj.c
-ifneq ($(COMPILER),clang)
+ifeq ($(WITH_CXX_TESTS),1)
```
etienne-lms, line 13: s/1/y/ ?

etienne-lms (commented):
In commit long for 1st commit "Introduce WITH_TLS_TESTS": missing `[1]` in  `Link:`.

+ comment in 2nd commit.

ta/os_test/sub.mk
jforissier: Good catch, fixed and tested properly with both `WITH_CXX_TESTS` set to `y` and `n` this time.

jforissier:
@jenswi-linaro may I apply your R-b tag to "Introduce WITH_CXX_TESTS"?

jenswi-linaro:
@jforissier, sure. Looks good.

jforissier:
IBART reports a failure but the log looks OK, although it appears to be truncated. @jbech-linaro did the test time out perhaps? Should I just try to restart the test?

PR #466, opened 2020-11-26
xtest: mark unused params

vchong:
Mark unused params to avoid build warnings/errors.


vchong:
Not sure why these weren't detected before by CI.

jforissier:
> Not sure why these weren't detected before by CI.

This is the "host" part of `xtest`, not the one that is cross-compiled so it is built with whichever compile comes by default with the Linux distribution. Many different GCC versions, could even be Clang. That is maybe why.

Anyway:

vchong:
Tag applied. Thanks!

PR #467, opened 2020-11-26
regression 1008: only test open session failure once

jforissier:
Test 1008 runs a negative test for TEEC_OpenSession() 100 times in a
loop. A comment says it is to check for memory leaks, but it is a bit
pointless because there is no guarantee that such a leak would be
detected this way. xtest is a functional test suite, not a stress test
tool so this kind of test is not really appropriate. It increases the
execution time and the amount of logs produced for very little benefit.

Therefore, remove the loop and test the open session failure only once.

### DIFF
--- a/host/xtest/regression_1000.c
+++ b/host/xtest/regression_1000.c
@@ -749,19 +749,15 @@ static void xtest_tee_test_1008(ADBG_Case_t *c)
 	{
 		size_t n = 0;
 
-		(void)ADBG_EXPECT_TEEC_RESULT(c, TEEC_ERROR_GENERIC,
-			xtest_teec_open_session(&session_crypt,
-						&create_fail_test_ta_uuid, NULL,
-						&ret_orig));
-		/*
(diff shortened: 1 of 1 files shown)


etienne-lms:

jenswi-linaro:
We have caught a few bugs with this in the past, but I agree that 100 times might be a bit excessive today. Perhaps we can let the `-l` affect the number of loops takes where the default value `1` (?) would result in only one loop, while any larger value would keep the old `100`. We could of course add even more granularity, but I doubt it's worth it.
The point with a number larger than 1 or 2 is that it leaves a bit better mark in case you track memory.

xtest is primarily a functional test, but not only we have a few other stress tests for concurrency which I also find quite useful.

jforissier:
@jenswi-linaro OK, `level` is already used to reduce the number of iterations in concurrency test 1013 so it makes sense. Updated.

jenswi-linaro:

PR #468, opened 2020-11-26
xtest: add applogin tests

parport0:
Add testcases applogin_1000, applogin_1001, and applogin_1002
for testing application-based login methods for use in case
of path-based application ID. Compile them conditionally (CFG_APPLOGIN).

Depends on kernel commit "tee: add support for application-based session login methods"
Mailing list archives: https://lists.trustedfirmware.org/pipermail/op-tee/2020-October/000211.html

Signed-off-by: Elvira Khabirova <e.khabirova@omprussia.ru>


jenswi-linaro:
These tests belongs in the regression series 1xxx suite.

parport0:
Moved the tests to `regression_1000.c`. The tests are now compiled unconditionally.

host/xtest/regression_1000.c
```diff
@@ -2349,3 +2349,214 @@ static void xtest_tee_test_1032(ADBG_Case_t *c)
...
+{
+#ifdef OPENSSL_FOUND
+	TEEC_Result result = TEEC_ERROR_GENERIC;
+	uint32_t ret_orig = 0;
+	TEEC_Session session = { };
+	uint32_t login = UINT32_MAX;
+	TEEC_UUID client_uuid = { };
+	TEEC_UUID expected_client_uuid = { };
+	TEEC_UUID uuid_ns = { };
+	char uuid_name[TEE_UUID_NS_NAME_SIZE + 4] = { };
```
jenswi-linaro, line 2371: Since you really mean `PATH_MAX + 4` why not use that instead and leave `TEE_UUID_NS_NAME_SIZE` alone?

host/xtest/regression_1000.c
```diff
@@ -2349,3 +2349,214 @@ static void xtest_tee_test_1032(ADBG_Case_t *c)
...
+	if (!ADBG_EXPECT_TEEC_SUCCESS(c, result))
+		return;
+
+	if (readlink("/proc/self/exe", app, PATH_MAX) < 0) {
+		Do_ADBG_Log(" - 1033 -   skip test, "
+			    "/proc/self/exe not readable");
+		return;
+	}
+
+	sprintf(uuid_name, "app=%s", app);
```
jenswi-linaro, line 2385: Please use `snprintf()` instead.

host/xtest/regression_1000.c
```diff
@@ -2349,3 +2349,214 @@ static void xtest_tee_test_1032(ADBG_Case_t *c)
...
+	uuid_name[TEE_UUID_NS_NAME_SIZE] = 0;
```
jenswi-linaro, line 2386: This isn't needed with `snprintf()` and if it's needed with `sprintf()` you're already in deep trouble.

jenswi-linaro (commented):
The comments for 1033 applies for 1034 too.

host/xtest/regression_1000.c
parport0: Because the kernel patch changes `TEE_UUID_NS_NAME_SIZE` to `PATH_MAX` and that would introduce a discrepancy.

host/xtest/regression_1000.c
parport0: The kernel buffer is `PATH_MAX` long, because having a buffer longer than that is more problematic to allocate. But since a path with added `app=` and `uid=` and stuff can be longer, the kernel truncates the buffer to `PATH_MAX`. This is considered safe since the TA checks whether the hash is an expected one or not, and the TA authors should not allow for collision attacks based on the application ID being truncated.

The kernel implementation in fact uses `snprintf()` for this. However, this is not possible in optee_test, since GCC complains about the string being truncated:
```
regression_1000.c: In function ‘xtest_tee_test_1033’:
regression_1000.c:2381:50: error: ‘%s’ directive output may be truncated writing up to 4095 bytes into a region of size 4092 [-Werror=format-truncation=]
  snprintf(uuid_name, TEE_UUID_NS_NAME_SIZE, "app=%s", app);
                                                  ^~   ~~~
regression_1000.c:2381:2: note: ‘snprintf’ output between 5 and 4100 bytes into a destination of size 4096
  snprintf(uuid_name, TEE_UUID_NS_NAME_SIZE, "app=%s", app);
  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
regression_1000.c: In function ‘xtest_tee_test_1034’:
regression_1000.c:2449:57: error: ‘%s’ directive output may be truncated writing up to 4095 bytes into a region of size between 4079 and 4086 [-Werror=format-truncation=]
  snprintf(uuid_name, TEE_UUID_NS_NAME_SIZE, "uid=%x:app=%s", geteuid(), app);
                                                         ^~              ~~~
regression_1000.c:2449:2: note: ‘snprintf’ output between 11 and 4113 bytes into a destination of size 4096
  snprintf(uuid_name, TEE_UUID_NS_NAME_SIZE, "uid=%x:app=%s", geteuid(), app);
  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
regression_1000.c: In function ‘xtest_tee_test_1035’:
regression_1000.c:2520:57: error: ‘%s’ directive output may be truncated writing up to 4095 bytes into a region of size between 4079 and 4086 [-Werror=format-truncation=]
  snprintf(uuid_name, TEE_UUID_NS_NAME_SIZE, "gid=%x:app=%s", group, app);
                                                         ^~          ~~~
regression_1000.c:2520:2: note: ‘snprintf’ output between 11 and 4113 bytes into a destination of size 4096
  snprintf(uuid_name, TEE_UUID_NS_NAME_SIZE, "gid=%x:app=%s", group, app);
  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
Hence the truncation is performed manually here. Though I will replace `sprintf()` with `snprintf()` anyway for hardening purposes, but with an actual buffer size.

parport0:
Addressed the comments.

host/xtest/regression_1000.c
jenswi-linaro: I'm sorry, I don't follow.

parport0: https://lists.trustedfirmware.org/pipermail/op-tee/2020-October/000211.html includes the following lines:
```
-#define TEE_UUID_NS_NAME_SIZE	128
+#define TEE_UUID_NS_NAME_SIZE	PATH_MAX
```
I could leave `TEE_UUID_NS_NAME_SIZE` at 128 in the tests, while the kernel patch changes it to `PATH_MAX`, but I'm not sure that's right.

jenswi-linaro: OK, a comment somewhere in the function explaining that we truncate the buffer at the same place as the kernel would help.

host/xtest/regression_1000.c
jenswi-linaro: Something like (might need some tweaking in the formatting)
``` C
snprintf(uuid_name, sizeof(uuid_name), "app=%*s", TEE_UUID_NS_NAME_SIZE, app);
```
should do it.

host/xtest/regression_1000.c
```diff
@@ -2349,3 +2349,214 @@ static void xtest_tee_test_1032(ADBG_Case_t *c)
...
+	if (!ADBG_EXPECT_TEEC_SUCCESS(c, result))
+		return;
+
+	if (readlink("/proc/self/exe", app, PATH_MAX) < 0) {
+		Do_ADBG_Log(" - 1033 -   skip test, "
+			    "/proc/self/exe not readable");
+		return;
+	}
+
+	snprintf(uuid_name, TEE_UUID_NS_NAME_SIZE + 4, "app=%s", app);
```
etienne-lms, line 2385: s/TEE_UUID_NS_NAME_SIZE + 16/sizeof(uuid_name)/

Also check truncation, I.e.:
```c
	len = snprintf(uuid_name, sizeof(uuid_name),  "app=%s", app);
	if (ADBG_EXPECT_TRUE(c, len > 0 && len < sizeof(uuid_name)))
		return;
```

Last, can remove line below (`uuid_name[TEE_UUID_NS_NAME_SIZE] = 0;`) as stated by @jenswi-linaro .

host/xtest/regression_1000.c
```diff
@@ -2349,3 +2349,214 @@ static void xtest_tee_test_1032(ADBG_Case_t *c)
...
+	ADBG_EXPECT_COMPARE_UNSIGNED(c, login, ==,
+				     TEEC_LOGIN_GROUP_APPLICATION);
+
+	ADBG_EXPECT_EQUAL(c, &expected_client_uuid, &client_uuid,
+			  sizeof(TEEC_UUID));
+
+out:
+	TEEC_CloseSession(&session);
+#else /*!OPENSSL_FOUND*/
+	/* xtest_uuid_v5() depends on OpenSSL */
```
etienne-lms, line 2555: `UNUSED(c);`

host/xtest/regression_1000.c
parport0: [A previous comment](https://github.com/OP-TEE/optee_test/pull/468#discussion_r533156066) explains why we need a manual truncation at `TEE_UUID_NS_NAME_SIZE`.

The return value of snprintf cannot be equal to or more than `sizeof(uuid_name)` in this case since `sizeof(uuid_name)` is defined as length of `app` plus four. I would not trust my own judgement on this, but the compiler agrees, since if the truncation *was* possible, the compiler would refuse to work with the `‘snprintf’ output between 5 and 4100 bytes into a destination of size 4096`-type errors as shown [above](https://github.com/OP-TEE/optee_test/pull/468#discussion_r533156066).

I will replace hard-coded lengths with `sizeof(uuid_name)`. I don't know cases in which `snprintf` can return a negative number that don't include your system catching on fire but I will add a check for this.

host/xtest/regression_1000.c
parport0: `"app=%.*s"` works.

But then the next parameter should supply not `TEE_UUID_NS_NAME_SIZE`, but `TEE_UUID_NS_NAME_SIZE - x`, where `x` is the amount of other characters in the string ("app="), since this parameter limits the length of the `%s` only. 

While this may work in the case of "app=%s", this wont in case of "uid=XXX:app=", since we don't know the length of the printed uid without calculating it. And at this point, manually setting a byte to `\0` is much easier.

parport0:
Addressed some of the comments.

jenswi-linaro (commented):
Looks good:

Now we only need to wait for the kernel patch to gain some acceptance.

host/xtest/regression_1000.c
```diff
@@ -2349,3 +2349,239 @@ static void xtest_tee_test_1032(ADBG_Case_t *c)
...
+	TEEC_Result result = TEEC_ERROR_GENERIC;
+	uint32_t ret_orig = 0;
+	TEEC_Session session = { };
+	uint32_t login = UINT32_MAX;
+	TEEC_UUID client_uuid = { };
+	TEEC_UUID expected_client_uuid = { };
+	TEEC_UUID uuid_ns = { };
+	char uuid_name[TEE_UUID_NS_NAME_SIZE + 4] = { };
+	char app[PATH_MAX] = { };
+	int len;
```
etienne-lms, line 2373: `= 0;`

etienne-lms (commented):
Thanks @parport0 for the explanations and updates.
LGTM with 2 minor remaining comments.

parport0:
Addressed the comments.

host/xtest/regression_1000.c
jenswi-linaro: OK

vesajaaskelainen: I just commented on the kernel patch about the length of uuid_name.

So if we try to get rid of the limit altogether then we don't need to care about it anymore.

host/xtest/regression_1000.c
```diff
@@ -2349,3 +2349,239 @@ static void xtest_tee_test_1032(ADBG_Case_t *c)
...
+	char uuid_name[TEE_UUID_NS_NAME_SIZE + 4] = { };
+	char app[PATH_MAX] = { };
+	int len = 0;
+
+	res = xtest_uuid_from_str(&uuid_ns, client_uuid_linux_ns);
+
+	if (!ADBG_EXPECT_TEEC_SUCCESS(c, res))
+		return;
+
+	if (readlink("/proc/self/exe", app, PATH_MAX) < 0) {
```
vesajaaskelainen, line 2380: In the kernel there are two methods for matching -- this requires that the kernel is configured for CONFIG_TEE_APPID_PATH.

If this is configured differently then the test will fail which is of course the then what we want.

In theory we could get it from /proc/config.gz IF that is available on running system -- but that might not be the case. 

Any ideas how we should handle this?

etienne-lms: We already had similar cases where optee components config/version could affect xtest result.
For those, we rather suggested in discussions (I don't remember which) to disable the test case (i.e. here `xtest -x 1033`) when running latest optee_test xtest with older optee components.

PR #471, opened 2020-12-11
xtest: pkcs11: add key generation tests

ruchi393:
Add 1010 for C_GenerateKey() with following mechanisms
supported:
PKCS11_CKM_GENERIC_SECRET_KEY_GEN
PKCS11_CKM_AES_KEY_GEN

### DIFF
--- a/host/xtest/pkcs11_1000.c
+++ b/host/xtest/pkcs11_1000.c
@@ -61,6 +61,12 @@ static CK_MECHANISM cktest_hmac_sha384_mechanism = {
 static CK_MECHANISM cktest_hmac_sha512_mechanism = {
 	CKM_SHA512_HMAC, NULL, 0,
 };
+static CK_MECHANISM cktest_gensecret_keygen_mechanism = {
+	CKM_GENERIC_SECRET_KEY_GEN, NULL, 0,
+};
+static CK_MECHANISM cktest_aes_keygen_mechanism = {
+	CKM_AES_KEY_GEN, NULL, 0,
+};
 
 /*
  * Util to find a slot on which to open a session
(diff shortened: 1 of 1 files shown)


host/xtest/pkcs11_1000.c
```diff
@@ -1887,3 +1893,210 @@ static void xtest_pkcs11_test_1009(ADBG_Case_t *c)
...
+	Do_ADBG_BeginSubCase(c, "Generate AES Key - Try Encrypting");
+
+	rv = C_GenerateKey(session, &cktest_aes_keygen_mechanism,
+			   cktest_generate_aes_object,
+			   ARRAY_SIZE(cktest_generate_aes_object),
+			   &key_handle);
+	if (!ADBG_EXPECT_CK_OK(c, rv))
+		goto err;
+
+	rv = C_EncryptInit(session, &cktest_aes_cbc_mechanism, key_handle);
```
etienne-lms, line 2080: this will leave a processing initiated, hence active in the token.
It should be terminated with a C_Entrypt() of C_EncryptFina().

ruchi393: Fixed with C_EncryptFinal()

host/xtest/pkcs11_1000.c
```diff
@@ -1887,3 +1893,215 @@ static void xtest_pkcs11_test_1009(ADBG_Case_t *c)
...
+/* Valid template to generate a generic secret */
+static CK_ATTRIBUTE cktest_generate_gensecret_object_valid1[] = {
+	{ CKA_CLASS, &(CK_OBJECT_CLASS){CKO_SECRET_KEY},
+						sizeof(CK_OBJECT_CLASS) },
+	{ CKA_KEY_TYPE, &(CK_KEY_TYPE){CKK_GENERIC_SECRET},
+						sizeof(CK_KEY_TYPE) },
+	{ CKA_ENCRYPT, &(CK_BBOOL){CK_TRUE}, sizeof(CK_BBOOL) },
+	{ CKA_DECRYPT, &(CK_BBOOL){CK_TRUE}, sizeof(CK_BBOOL) },
+	{ CKA_SIGN, &(CK_BBOOL){CK_TRUE}, sizeof(CK_BBOOL) },
+	{ CKA_VERIFY, &(CK_BBOOL){CK_TRUE}, sizeof(CK_BBOOL) },
```
vesajaaskelainen, line 1936: About `CKA_ENCRYPT`, `CKA_DECRYPT`, `CKA_SIGN` and `CKA_VERIFY` for `CKK_GENERIC_SECRET`.

In `PKCS #11 Cryptographic Token Interface Current Mechanisms Specification Version 2.40 Plus Errata 01` and `2.6 Generic secret key`'s table 44 there are no checks for those operations.

And in the text:
`Generic secret key objects (object class CKO_SECRET_KEY, key type CKK_GENERIC_SECRET) hold generic secret keys. These keys do not support encryption or decryption; however, other keys can be derived from them and they can be used in HMAC operations.`

This would mean in my understanding that those would be invalid.

vesajaaskelainen: Then again HMAC can be considered needing `CKA_SIGN` and `CKA_VERIFY`.

vesajaaskelainen: So probably one would need to remove
```
	{ CKA_ENCRYPT, &(CK_BBOOL){CK_TRUE}, sizeof(CK_BBOOL) },
	{ CKA_DECRYPT, &(CK_BBOOL){CK_TRUE}, sizeof(CK_BBOOL) },
```
Or set them as `CK_FALSE`.

ruchi393: We don't allow the encrypt and decrypt operations even if they are set in the template for Generic keys. There is a testcase which uses this template and if a C_Encrypt operation is attempted with a CKK_GENERIC_SECRET key  even with C_Encrypt set, error CKR_KEY_FUNCTION_NOT_PERMITTED is returned. 

In current code , if someone tries to set CKA_ENCRYPT or DECRYPT to true for a generic secret , we don't check it there. Should we be checking and not allowing these properties to be set ? Specification is not very clear on this. I guess just preventing the encrypt/decrypt operation even if the property is set should be ok

ruchi393: For CKA_SIGN and CKA_VERIFY and Generic secret keys, should there be any restriction if someone attempts to use the generic secret key directly ? The current code allows these operations on a key of type CKK_GENERIC_SECRET

As you have mentioned, specification states :
Generic secret key objects (object class CKO_SECRET_KEY, key type CKK_GENERIC_SECRET) hold generic secret keys. These keys do not support encryption or decryption; however, other keys can be derived from them and they can be used in HMAC operations.

So ideally, only keys derived from CKK_GENERIC_SECRET should be allowed in sign/verify.  When deriving key from a CKK_GENERIC_SECRET key does the key type of the derived key change ? I am not sure in this scenario how we would be able to distinguish between a generate/imported CKK_GENERIC_SECRET and a derived CKK_GENERIC_SECRET.

Specification mentions that CKA_LOCAL property will be false for a derived key. That can be one way of distinguishing between generated and derived key. But the property would be false even for a key imported and created using C_CreateObject

etienne-lms: > Should we be checking and not allowing these properties to be set ? 

Would that rely help?

> When deriving key from a CKK_GENERIC_SECRET key does the key type of the derived key change ?

Regarding key derivation, the caller must provide a template for the key created from a derivation. So this generated key will not inherit attributes from the key used for the derivation.

Actually, there is this ID `CKA_DERIVE_TEMPLATE` defined in  the CryptoKi API. Neither PKCS#11 v2.40 nor v3.0 clearly specify this attribute however I think it would be good we kind of support this extension, allowing a key to restrict its use when used to generate a derived key. But that's another story.

> ```x
>	{ CKA_ENCRYPT, &(CK_BBOOL){CK_TRUE}, sizeof(CK_BBOOL) },
>	{ CKA_DECRYPT, &(CK_BBOOL){CK_TRUE}, sizeof(CK_BBOOL) },
> ```

As for here, since defining a `CKK_GENERIC_SECRET` with `CKA_ENCRYPT`/`CKA_DECRYPT` true is meaningless in PKCS11 as you pointed out, I think these 2 attributes should simply be removed here.
Maybe added as a invalid test case but I think it's a weak test case and I'm not sure it's worth adding such.

vesajaaskelainen: > As you have mentioned, specification states :
> Generic secret key objects (object class CKO_SECRET_KEY, key type CKK_GENERIC_SECRET) hold generic secret keys. These keys do not support encryption or decryption; however, other keys can be derived from them and they can be used in HMAC operations.
> 
> So ideally, only keys derived from CKK_GENERIC_SECRET should be allowed in sign/verify. When deriving key from a CKK_GENERIC_SECRET key does the key type of the derived key change ? I am not sure in this scenario how we would be able to distinguish between a generate/imported CKK_GENERIC_SECRET and a derived CKK_GENERIC_SECRET.

I would interpret that as that you can use them in HMAC without deriving them first.

vesajaaskelainen: > We don't allow the encrypt and decrypt operations even if they are set in the template for Generic keys. There is a testcase which uses this template and if a C_Encrypt operation is attempted with a CKK_GENERIC_SECRET key even with C_Encrypt set, error CKR_KEY_FUNCTION_NOT_PERMITTED is returned.

I believe this is OK and good test.

ruchi393: > > Should we be checking and not allowing these properties to be set ?
> 
> Would that rely help?

No. These attributes are anyways checked when doing some operation using the key. Specification also doesn't mandate it so my take would be to skip checking these attributes during generation/creation of object.

> 
> > When deriving key from a CKK_GENERIC_SECRET key does the key type of the derived key change ?
> 
> Regarding key derivation, the caller must provide a template for the key created from a derivation. So this generated key will not inherit attributes from the key used for the derivation.
> 
> Actually, there is this ID `CKA_DERIVE_TEMPLATE` defined in the CryptoKi API. Neither PKCS#11 v2.40 nor v3.0 clearly specify this attribute however I think it would be good we kind of support this extension, allowing a key to restrict its use when used to generate a derived key. But that's another story.
I will check this out when adding support for derived keys.
> 
> > ```
> >   { CKA_ENCRYPT, &(CK_BBOOL){CK_TRUE}, sizeof(CK_BBOOL) },
> >   { CKA_DECRYPT, &(CK_BBOOL){CK_TRUE}, sizeof(CK_BBOOL) },
> > ```
> 
> As for here, since defining a `CKK_GENERIC_SECRET` with `CKA_ENCRYPT`/`CKA_DECRYPT` true is meaningless in PKCS11 as you pointed out, I think these 2 attributes should simply be removed here.
I will remove these attributes when creating GENERIC_SECRET_KEY
> Maybe added as a invalid test case but I think it's a weak test case and I'm not sure it's worth adding such.

vesajaaskelainen:
I believe this looks OK and once squashed as one commit:
`Reviewed-by: Vesa Jääskeläinen <vesa.jaaskelainen@vaisala.com>`

ruchi393:
Squashed and tags applied

PR #478, opened 2021-01-14
xtest: pkcs11: Modify testcase 1011 for changes in C_Logout()

ruchi393:
PKCS#11 specification states that :
When C_Logout successfully executes, any of the application’s
handles to private objects should become invalid (even if a user
is later logged back into the token, those handles remain invalid).
In addition, all private session objects from sessions belonging
to the application should also be destroyed.

One of the find sub-test case was expecting that after logging back
in a session , private session object would be visible which was
a wrong assumption. This has been modified.

Another subcase assumed the handle of the token object returned
would be same as that when the token object was created by that
session. This assumption would hold good only during the first
log in into this session. If a user logs out and then logs in
again, the token object handles would change. So this sub-case
has been moved to be tested during the first login to the session.

### DIFF
--- a/host/xtest/pkcs11_1000.c
+++ b/host/xtest/pkcs11_1000.c
@@ -2339,10 +2339,32 @@ static void xtest_pkcs11_test_1011(ADBG_Case_t *c)
 	if (!ADBG_EXPECT_CK_OK(c, rv))
 		goto out;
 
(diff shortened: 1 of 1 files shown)


etienne-lms (commented):

ruchi393:
Tag applied

PR #485, opened 2021-02-06
Fix failing build

vesajaaskelainen:
With integration of 01f6f06 ("regression: add tests for the supplicant plugin framework") builds started failing in Yocto.

This PR fixes the issue and also silences older warning.

### DIFF
--- a/host/supp_plugin/Makefile
+++ b/host/supp_plugin/Makefile
@@ -6,16 +6,18 @@ PLUGIN_OBJ		= $(patsubst %.c, $(O)/supp_plugin/%.o, $(PLUGIN_SRS))
 PLUGIN_INCLUDES_DIR	= $(CURDIR)/include $(OPTEE_CLIENT_EXPORT)/include
 
 PLUGIN_INCLUDES		= $(addprefix -I, $(PLUGIN_INCLUDES_DIR))
-PLUGIN_CCFLAGS		= -Wall -fPIC
-PLUGIN_LDFLAGS		= -shared
+PLUGIN_CCFLAGS		= $(CFLAGS) -Wall -fPIC
+PLUGIN_LDFLAGS		= $(LDFLAGS) -shared
+
+CC			?= $(CROSS_COMPILE)gcc
 
 $(O)/supp_plugin/$(PLUGIN): $(PLUGIN_OBJ)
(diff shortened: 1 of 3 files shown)


jforissier:
Using CFLAGS and LDFLAGS when cross-compilation is involved makes me a bit nervous. They might be set assuming host compilation and break cross-builds. Anyways we can always fix that later if problems happens...

vesajaaskelainen:
> Using CFLAGS and LDFLAGS when cross-compilation is involved makes me a bit nervous. They might be set assuming host compilation and break cross-builds. Anyways we can always fix that later if problems happen...
> 

Here is the examples from `xtest` compilation:

`CC`:
https://github.com/OP-TEE/optee_test/blob/master/host/xtest/Makefile#L18

`CFLAGS`:
https://github.com/OP-TEE/optee_test/blob/master/host/xtest/Makefile#L105
https://github.com/OP-TEE/optee_test/blob/master/host/xtest/Makefile#L138

`LDFLAGS`:
https://github.com/OP-TEE/optee_test/blob/master/host/xtest/Makefile#L151

Why it works in `xtest` is that those `xFLAGS` is because they are being added `+=` and not set. 

And `CC` is conditionally set.

compilation:
https://github.com/OP-TEE/optee_test/blob/master/host/xtest/Makefile#L164

linking:
https://github.com/OP-TEE/optee_test/blob/master/host/xtest/Makefile#L159

modifying this plugin's Makefile to be more like `xtest`'s `Makefile` in usage of variables and style. I do not see the need to have `PLUGIN_` prefix in most of those variables.

Applied your tags.

vesajaaskelainen:
Note: `make install` (https://github.com/OP-TEE/optee_test/blob/master/Makefile#L76) for these is also a bit broken as there is missing default for `CFG_TEE_PLUGIN_LOAD_PATH` (and has also issues with `TA_DIR` and `bindir`) but our Yocto build weren't using that and then we also added additional permission configs so we have bitbake recipe with bit more customized `do_install`.

jenswi-linaro:
I think we'd like to have those patches too. We can take that in another PR to avoid broaden the scope here.

jenswi-linaro:

vesajaaskelainen:
> I think we'd like to have those patches too. We can take that in another PR to avoid broaden the scope here.

Let's do new PR for that as that is not so time critical.

vesajaaskelainen:
> `Acked-by: Jens Wiklander <jens.wiklander@linaro.org>`

Thansk. Applied tags.

PR #491, opened 2021-02-21
xtest: pkcs11: Add tests for digest operations

vesajaaskelainen:
Add test cases to test calling of:

- C_DigestInit
- C_Digest
- C_DigestUpdate
- C_DigestKey
- C_DigestFinal

### DIFF
--- a/host/xtest/pkcs11_1000.c
+++ b/host/xtest/pkcs11_1000.c
@@ -5,6 +5,9 @@
 
 #include <ck_debug.h>
 #include <inttypes.h>
+#ifdef OPENSSL_FOUND
+#include <openssl/evp.h>
+#endif
 #include <pkcs11.h>
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


host/xtest/pkcs11_1000.c
```diff
@@ -3741,3 +3742,863 @@ static void xtest_pkcs11_test_1016(ADBG_Case_t *c)
...
+		ADBG_EXPECT_CK_OK(c, C_DestroyObject(session, key_handle));
+		key_handle = CK_INVALID_HANDLE;
+	}
+
+	ADBG_EXPECT_CK_OK(c, C_CloseSession(session));
+
+close_lib:
+	ADBG_EXPECT_CK_OK(c, close_lib());
+}
+ADBG_CASE_DEFINE(pkcs11, 1020, xtest_pkcs11_test_1020,
```
vesajaaskelainen, line 4603: number here needs to be updated before merge.

host/xtest/pkcs11_1000.c
```diff
@@ -3741,3 +3742,863 @@ static void xtest_pkcs11_test_1016(ADBG_Case_t *c)
...
+	CK_ULONG digest_size;
+} digest_test_patterns[] = {
+	{
+		.test_name = "CKM_MD5/empty",
+		.mecha = CKM_MD5,
+		.data = digest_test_pattern_empty,
+		.data_size = sizeof(digest_test_pattern_empty) - 1,
+		.digest = digest_test_pattern_empty_md5,
+		.digest_size = sizeof(digest_test_pattern_empty_md5),
+	},
```
etienne-lms, line 3842: Maybe a nice macro to implement something like:
```c
	DIGEST_TEST{"CKM_MD5/empty", patern_empty, md5),
	DIGEST_TEST{"CKM_MD5/test pattern", patern, md5),
	DIGEST_TEST{"CKM_SHA_1/empty", patern_empty, sha1),
	DIGEST_TEST{"CKM_SHA_1/test pattern", patern, sha1),
	...

File: host/xtest/pkcs11_1000.c [hunk 3]
```diff
@@ -3741,3 +3742,863 @@ static void xtest_pkcs11_test_1016(ADBG_Case_t *c)
...
+		.mecha = CKM_SHA512,
+		.data = digest_test_pattern,
+		.data_size = sizeof(digest_test_pattern) - 1,
+		.digest = digest_test_pattern_sha512,
+		.digest_size = sizeof(digest_test_pattern_sha512),
+	},
+};
+
+static CK_ATTRIBUTE digest_generate_aes_object[] = {
+	{ CKA_CLASS, &(CK_OBJECT_CLASS) { CKO_SECRET_KEY },
```
etienne-lms, line 3934: no space char between cast and cast'ed:
`	{ CKA_CLASS, &(CK_OBJECT_CLASS){ CKO_SECRET_KEY },`

host/xtest/pkcs11_1000.c
```diff
@@ -3741,3 +3742,863 @@ static void xtest_pkcs11_test_1016(ADBG_Case_t *c)
...
+	CK_SESSION_HANDLE session = CK_INVALID_HANDLE;
+	CK_FLAGS session_flags = CKF_SERIAL_SESSION | CKF_RW_SESSION;
+	bool logged_in = false;
+	uint8_t data[128] = { 0 };
+	CK_ULONG data_size = 0;
+	uint8_t digest[64] = { 0 };
+	CK_ULONG digest_size = 0;
+	const uint8_t *expect_digest = NULL;
+	CK_ULONG expect_digest_size = 0;
+	CK_MECHANISM mechanism = {
```
etienne-lms, line 3976: i think `CK_MECHANISM mechanism = { CKM_MD5, NULL, 0 };` is fine.

host/xtest/pkcs11_1000.c
```diff
@@ -3741,3 +3742,863 @@ static void xtest_pkcs11_test_1016(ADBG_Case_t *c)
...
+	if (!ADBG_EXPECT_CK_OK(c, rv))
+		goto out;
+
+	rv = C_DigestFinal(session, digest, &digest_size);
+	if (!ADBG_EXPECT_CK_OK(c, rv))
+		goto out;
+
+	/* Verify digest with openssl */
+	mdctx = EVP_MD_CTX_create();
+	if (!mdctx)
```
etienne-lms, line 4147: it would be nice to trace the unlikely failures. Maybe change here to:
```c
	mdctx = EVP_MD_CTX_create();
	if (!ADBG_EXPECT_NOT_NULL(mdctx))
		goto out;
```
and at `out:`:
```
out:
	if (!ADBG_EXPECT_NULL(mdctx)) {
		Do_ADBG_Log("Unexpected failure in openssl functions: %d", ret);
		EVP_MD_CTX_destroy(mdctx);
	}
```

etienne-lms (commented):
thanks for those tests. 
This change brings openssl in the test. Is it really needed? Maybe running without openssl should at least not run other tests. I guess we could implement the expected hashes. But as adding thing in the TA, one day or another, more complex crypto will likely bring openssl back.

vesajaaskelainen:
> thanks for those tests.
> ... and thanks for your patience, 11 days...

No worries -- I am pretty overloaded myself too.

> This change brings openssl in the test. Is it really needed? Maybe running without openssl should at least not run other tests. I guess we could implement the expected hashes. But as adding thing in the TA, one day or another, more complex crypto will likely bring openssl back.

I was thinking that perhaps it is a good idea to discuss this now.

I believe it is a good idea to verify that operation does what it is expected to do -- for digest it is all about that result is correct. I trust that actual crypto library in OP-TEE side works nicely -- so we don't need to test that but what we need to test is that calculations actually work as expected by PKCS#11 -- there are multi step operations and thus it would be a very good to verify that in failure cases the crypto state in in expected state. Also good idea to verify that data passes correct -- and we do not get regressions if someone goes and changes something in the pipe.

For most cases in digest tests the real problem is update key -- as they key needs to be generated in OP-TEE side we need to read it back and then do live calculation -- thus you need to some crypto library.

Also next PR after those will be ECDSA and that starts to need use of some crypto library. Duplicating in example DER code, verifying things and such is not very good idea. I always recommend to use real already tested and good crypto library.

If we have too parametrized test suite then we start to miss some tests in some situations and that could hide problems.

Thus I would recommend that we select a crypto library that we use in the tests and just require that to be there.

I am happy to use mbedtls or something else too but in this case I believe as openssl is already there and does the needed job -- lets use that.

host/xtest/pkcs11_1000.c
vesajaaskelainen: OK

host/xtest/pkcs11_1000.c
vesajaaskelainen: OK

vesajaaskelainen: Just a note -- I have trigged my local copy to have more checkpatch stuff... and it complains
```
ERROR: space required before the open brace '{'
#24: FILE: host/xtest/pkcs11_1000.c:3934:
+	{ CKA_CLASS, &(CK_OBJECT_CLASS){ CKO_SECRET_KEY },
```
Did the change nevertheless.

host/xtest/pkcs11_1000.c
vesajaaskelainen: I am not big fan of macro programming but perhaps in this particular case it is OK.

host/xtest/pkcs11_1000.c
vesajaaskelainen: OK.

vesajaaskelainen:
Updated the PR based on feedback.

host/xtest/pkcs11_1000.c
```diff
@@ -3741,3 +3742,802 @@ static void xtest_pkcs11_test_1016(ADBG_Case_t *c)
...
+
+		if (!ADBG_EXPECT_BUFFER(c, expect_digest, expect_digest_size,
+					digest,	digest_size))
+			goto out;
+
+		/* Verify that end of buffer is still 0xCC */
+		for (i = expect_digest_size; i < sizeof(digest); i++)
+			if (!ADBG_EXPECT_COMPARE_UNSIGNED(c, digest[i], ==,
+							  0xCC))
+				break;
```
etienne-lms, line 3980: `goto out;`

host/xtest/pkcs11_1000.c
```diff
@@ -3741,3 +3742,802 @@ static void xtest_pkcs11_test_1016(ADBG_Case_t *c)
...
```
etienne-lms, line 4018: `goto out;`

host/xtest/pkcs11_1000.c
```diff
@@ -3741,3 +3742,802 @@ static void xtest_pkcs11_test_1016(ADBG_Case_t *c)
...
+	if (!ADBG_EXPECT_COMPARE_SIGNED(c, ret, ==, 1))
+		goto out;
+	ret = EVP_DigestFinal_ex(mdctx, hash, &md_len);
+	if (!ADBG_EXPECT_COMPARE_SIGNED(c, ret, ==, 1))
+		goto out;
+	EVP_MD_CTX_destroy(mdctx);
+	mdctx = NULL;
+
+	if (!ADBG_EXPECT_BUFFER(c, hash, md_len,
+				digest,	digest_size))
```
etienne-lms, line 4101: fits in 1 line

etienne-lms (commented):
optee_test can be enabled in build environments that may not support openssl.
We need optee_test build not to failed in such cases.
I see 2 options:
1. condition pkcs11_1000.c build upon `OPENSSL_FOUND`/`WITH_OPENSSL` in makefiles.
2. condition implementation of out data verification upon `OPENSSL_FOUND` inside **pkcs11_1000.c** .

I prefer the 2nd one.

host/xtest/pkcs11_1000.c
vesajaaskelainen: OK

host/xtest/pkcs11_1000.c
vesajaaskelainen: OK

host/xtest/pkcs11_1000.c
vesajaaskelainen: OK

vesajaaskelainen:
Updated the PR based on feedback.

host/xtest/pkcs11_1000.c
```diff
@@ -3741,3 +3744,810 @@ static void xtest_pkcs11_test_1016(ADBG_Case_t *c)
...
+		goto out;
+	EVP_MD_CTX_destroy(mdctx);
+	mdctx = NULL;
+
+	if (!ADBG_EXPECT_BUFFER(c, hash, md_len, digest, digest_size))
+		goto out;
+#else
+	Do_ADBG_Log("OpenSSL not available, skipping C_DigestKey verification");
+#endif
+	ADBG_EXPECT_CK_OK(c, C_DestroyObject(session, key_handle));
```
etienne-lms, line 4110: maybe an empty line above.

host/xtest/pkcs11_1000.c
```diff
@@ -3741,3 +3744,810 @@ static void xtest_pkcs11_test_1016(ADBG_Case_t *c)
...
```
etienne-lms, line 4188: ditto

etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` (with 2 minor comments)

host/xtest/pkcs11_1000.c
vesajaaskelainen: Sure.

host/xtest/pkcs11_1000.c
vesajaaskelainen: Sure.

vesajaaskelainen:
Updated the PR with feedback.

Applied tag. Thanks!

NOTE: test number need to be updated so that this will be last in the list.

host/xtest/pkcs11_1000.c
vesajaaskelainen: Updated.

vesajaaskelainen:
Rebased the PR now that key derivation support was merged.

Used next test ID for this PR.

etienne-lms:
Thanks @vesajaaskelainen for the whole digest support!

PR #497, opened 2021-03-10
xtest: pkcs11: Correct template for secret key

ruchi393:
When creating object of type CKO_GENERIC_SECRET using
C_CreateObject(), CKA_VALUE is mandatory attribute in the
template and CKA_VALUE_LENGTH should not be provided. This
check was added in the implementation recently so the template
need to be corrected.

Also add an additional check to confirm that CKA_VALUE_LENGTH
gets populated correctly by the TA when such objects are created.

This depends on 
https://github.com/OP-TEE/optee_os/pull/4465


etienne-lms (commented):

ruchi393:
Tag applied.

PR #504, opened 2021-04-12
xtest: CMake: enable CFG_PKCS11_TA when building with CFG_PKCS11_TA=y

etienne-lms:
Define CFG_PKCS11_TA macro in xtest source compilation when
CFG_PKCS11_TA=y unless what build fails when optee_test is built with
CFG_PKCS11_TA=y while optee_os was built without.

Fixes: https://github.com/OP-TEE/optee_os/issues/4550

### DIFF
--- a/host/xtest/CMakeLists.txt
+++ b/host/xtest/CMakeLists.txt
@@ -103,6 +103,7 @@ if (CFG_SECURE_DATA_PATH)
 endif()
 
 if (CFG_PKCS11_TA)
+	add_compile_options(-DCFG_PKCS11_TA)
 	list (APPEND SRC pkcs11_1000.c)
 endif()
 


jforissier:

jenswi-linaro:
This error seems to be a mismatch between flags used when TA dev kit was compiled and later when xtest was built.
This approach makes sense in this particular situation, but we don't want to end up in a situation where every sing `CFG_*` is added as a compiler flag one by in this `CMakList.txt`. By the way, don't we need something similar in `Makefile`?

etienne-lms:
There's already a equivalent in **host/xtest/Makefile**, see https://github.com/OP-TEE/optee_test/blob/3.12.0/host/xtest/Makefile#L98

jenswi-linaro:

etienne-lms:
tags applied

Issue #510, opened 2021-04-27
gp_40178 random failures


jforissier:
```
# QEMUv8
$ make GP_PACKAGE=/path/to/TEE_Initial_Configuration-Test_Suite_v2_0_0_2-2017_06_09.7z run
```
```
Welcome to Buildroot, type root or test to login
buildroot login: test
$ xtest gp_40178
Test ID: gp_40178
Run test suite with level=0

TEE test application started over default TEE instance
######################################################
#
# regression+gp+pkcs11
#
######################################################

* gp_40178 ce-7b-17
Segmentation fault
$
```

See also https://github.com/OP-TEE/optee_os/pull/4552#issuecomment-827563612.

jforissier:
The problematic line in `xtest_tee_40178()`:
```C
    Test_Tool_Erase_Persistent_Time(c, SESSION01, CMD_TEE_SetTAPersistentTime_and_GetTAPersistentTime);
```
At this point, SESSION01 is not open.

In fact, the whole test is not applicable because OP-TEE provides no external tool to erase the persistent time of a TA.
Who knows how to exclude a test from the GP test suite? @jenswi-linaro @etienne-lms

etienne-lms:
Thanks for the investigation. I agree we should disable this test.

PR #511, opened 2021-04-27
gp: disable test ce-7b-17 (gp_40178)

jforissier:
Add a patch to exclude ce-7b-17 from the test suite since OP-TEE does
not provide the necessary external command to erase TA persistent time.

Note that skipping a test will shift the numbering in the "gp" suite,
so after this patch gp_40178 will be the same as gp_40179 prior to the
patch.

Fixes: https://github.com/OP-TEE/optee_test/issues/510


jenswi-linaro:
Since this is a bug in OP-TEE we should keep track of this so it can be fixed eventually.

jenswi-linaro:

jforissier:
> Since this is a bug in OP-TEE we should keep track of this so it can be fixed eventually.

Not exactly a bug, because the spec does not mention there should be any mean to discard the persistent time once it has been set (see the state machine in the description of `TEE_GetTAPersistentTime()`). It is only needed for testability. That would be a special "test mode" enabled at compile time perhaps.

Anyway, I will create an issue.

PR #513, opened 2021-04-29
regression 4002: add CMAC test case with a 16 bytes increment

clementfaure:
Hello,

@jforissier
The test to add is in the end very simple ... just one `XTEST_MAC_CMAC_CASE` :)
Should I specify in the commit message, that it's related to the CAAM? I don't know if I should since this test is generic and the CAAM platform specific ..

Please refer to https://github.com/OP-TEE/optee_os/pull/4583

Thanks

### DIFF
--- a/host/xtest/regression_4000.c
+++ b/host/xtest/regression_4000.c
@@ -1160,6 +1160,7 @@ static const struct xtest_mac_case mac_cases[] = {
 	XTEST_MAC_CMAC_CASE(vect10, 9),
 	XTEST_MAC_CMAC_CASE(vect11, 9),
 	XTEST_MAC_CMAC_CASE(vect12, 9),
+	XTEST_MAC_CMAC_CASE(vect12, 16),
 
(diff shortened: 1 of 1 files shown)


jenswi-linaro:
If you're targeting a bug that is supposed to be fixed please mention that in the commit message. A link to https://github.com/OP-TEE/optee_os/pull/4583 in the commit message should be useful.

jforissier (approved):
Please add:
```
Link: https://github.com/OP-TEE/optee_os/pull/4583
Signed-off-by: ...
```
...to the commit description as suggested by Jens. With that:

Thanks!

clementfaure:
@jenswi-linaro @jforissier Commit reworded!

jenswi-linaro:

jforissier:
"does not properly handle"? That was the problem that was fixed, right? So you can't say that. Instead "This is a test case for a bug in the CAAM driver which would not handle... [1]" and `Link: [1] https://...`.

clementfaure:
Indeed, that's not well written 🤦 .
Let me fix that

PR #519, opened 2021-06-21
xtest : disable large TA tests with virtualization

ruchi393:
Disabled the large TA part of 1013 and large TA test 1034
when virtualization is enabled since it uses too much secure heap.
For QEMU environment, enough secure memory is not available
for each guest, so these testcases fail when CFG_VIRTUALIZATION=y


jenswi-linaro (commented):

ruchi393:
Thanks @jenswi-linaro . Tag applied.

PR #524, opened 2021-06-27
xtest: Add more tests for querying signature size for C_Sign/C_SignFinal

vesajaaskelainen:
It is possible that when querying memory that return buffer for signature
size does not have any meaningful value.

It has been observed that some PKCS#11 libraries does not initialize the
memory to zero and it is also completely fine based on specification.

Add test for the case so that it does not regress in future.

Specified in:
PKCS #11 Cryptographic Token Interface Base Specification
Version 2.40 Plus Errata 01
5.2 Conventions for functions returning output in a variable-length buffer

Related PRs:
- https://github.com/OP-TEE/optee_client/pull/273

### DIFF
--- a/host/xtest/pkcs11_1000.c
+++ b/host/xtest/pkcs11_1000.c
@@ -1676,6 +1676,15 @@ static void xtest_pkcs11_test_1008(ADBG_Case_t *c)
 			if (!ADBG_EXPECT_CK_OK(c, rv))
 				goto err_destr_obj;
 
+			/*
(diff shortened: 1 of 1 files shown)


ruchi393:
`Reviewed-by: Ruchika Gupta <ruchika.gupta@linaro.org>`

vesajaaskelainen:
Thanks for the review. Applied tags.

etienne-lms:

vesajaaskelainen:
Thanks for the review. Applied tags.

PR #526, opened 2021-06-27
xtest: Verify that sensitive information is protected

vesajaaskelainen:
When secret object's properties are not protected make sure that they can
be read normally.

When they are protected make sure that proper error values are reported:

- C_GetAttributeValue return value is CKR_ATTRIBUTE_SENSITIVE
- For field that is protected size is CK_UNAVAILABLE_INFORMATION

Specified in:
PKCS #11 Cryptographic Token Interface Base Specification
Version 2.40 Plus Errata 01
C_GetAttributeValue


vesajaaskelainen:
Thanks. Applied tags.

PR #527, opened 2021-06-29
xtest 1013: ignore out of memory errors on large TA

jforissier:
Regression test 1013 (concurrency) is split in two: first, a small TA
is used in sub-case 1013.1, then a larger one is loaded in 1013.2. The
first sub-case normally passes on all platforms, but the second one may
fail on memory constrained platforms or configurations. To address this
issue, a couple of compile-time checks have been introduced (on
CFG_PAGED_USER_TA and CFG_VIRTUALIZATION). There are two problems with
this approach:
1. It is testing for features rather than actual memory constraints,
which is fragile.
2. CFG_VIRTUALIZATION isn't even exported into the TA dev kit by the
OP-TEE OS build environment, so it may happen that its value is different
in optee_os and in optee_test. Moreover, virtualization should be
transparent to TAs.

This commit removes the compile-time tests and modifies 1013.2 to ignore
any out-of-memory error when loading the TA. Other errors are still
detected.

### DIFF
--- a/host/xtest/regression_1000.c
+++ b/host/xtest/regression_1000.c
@@ -1099,6 +1099,7 @@ static void xtest_tee_test_1013_single(ADBG_Case_t *c, double *mean_concurrency,
 	};
 	uint8_t out[32] = { };
 	pthread_t thr[NUM_THREADS] = { };
+	bool skip = false;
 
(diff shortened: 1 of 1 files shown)


jenswi-linaro (commented):

PR #528, opened 2021-06-29
xtest 1034: ignore out of memory errors on large TA

jforissier:
Same fix than commit fbdc175bb9f3 ("xtest 1013: ignore out of memory
errors on large TA") for regression test 1034.


jforissier:
Missed that one :-/ otherwise I would have put it in the same PR...

jenswi-linaro:

PR #531, opened 2021-07-06
regression 4006: Fix xtest no CRT fail

lubing521:
if enable mbedtls with CRT in optee-os,
these cases will cause op-tee OS check them fail
which no provide CRT.add macro controls to turn
off the test when libmbedtls with CRT enable in optee-os.

Signed-off-by: Lu Bing <lubing@eswin.com>

### DIFF
--- a/host/xtest/regression_4000.c
+++ b/host/xtest/regression_4000.c
@@ -24,6 +24,8 @@
 
 #include <assert.h>
 
+#include <mbedtls_config_kernel.h>
+
 static TEEC_Result ta_crypt_cmd_reset_operation(ADBG_Case_t *c, TEEC_Session *s,
 						TEE_OperationHandle oph)
 {
@@ -2724,6 +2726,8 @@ struct xtest_ac_case {
 	XTEST_AC_CASE(level, algo, mode, vect, XTEST_AC_ECDSA_UNION(vect))
 
 static const struct xtest_ac_case xtest_ac_cases[] = {
+
+#if !defined(CFG_CRYPTOLIB_NAME_mbedtls) ||  defined(MBEDTLS_RSA_NO_CRT)
 	/* RSA test without crt parameters */
 	XTEST_AC_RSA_CASE(0, TEE_ALG_RSA_NOPAD, TEE_MODE_ENCRYPT,
 			  ac_rsassa_vect1, NULL_ARRAY, WITHOUT_SALT),
@@ -2858,7 +2862,7 @@ static const struct xtest_ac_case xtest_ac_cases[] = {
 	XTEST_AC_RSA_CASE(1, TEE_ALG_RSAES_PKCS1_OAEP_MGF1_SHA1,
 			  TEE_MODE_ENCRYPT,
 			  ac_rsaes_oaep_vect10, NULL_ARRAY, WITHOUT_SALT),
-
+#endif /* !CFG_CRYPTOLIB_NAME_mbedtls || !MBEDTLS_RSA_NO_CRT */
 	/* RSA test with crt parameters */
 	XTEST_AC_RSA_CASE(0, TEE_ALG_RSA_NOPAD, TEE_MODE_ENCRYPT,
 			  ac_rsassa_vect1, ARRAY, WITHOUT_SALT),


host/xtest/regression_4000.c
```diff
@@ -2724,6 +2727,7 @@ struct xtest_ac_case {
 	XTEST_AC_CASE(level, algo, mode, vect, XTEST_AC_ECDSA_UNION(vect))
 
 static const struct xtest_ac_case xtest_ac_cases[] = {
+#if defined(CFG_CRYPTOLIB_NAME_mbedtls) &&  defined(MBEDTLS_RSA_NO_CRT)
```
jforissier, line 2730: Shouldn't this be `#if !defined(CFG_CRYPTOLIB_NAME_mbedtls) || defined(MBEDTLS_RSA_NO_CRT)`?

host/xtest/regression_4000.c
```diff
@@ -2859,6 +2863,7 @@ static const struct xtest_ac_case xtest_ac_cases[] = {
 			  TEE_MODE_ENCRYPT,
 			  ac_rsaes_oaep_vect10, NULL_ARRAY, WITHOUT_SALT),
 
+#endif
```
etienne-lms, line 2866: move before the empty line and recall the condition in an inline comment, i.e.:
`#endif /* !CFG_CRYPTOLIB_NAME_mbedtls || !MBEDTLS_RSA_NO_CRT */`

host/xtest/regression_4000.c
etienne-lms, line 2730: prefer:
`#if !defined(CFG_CRYPTOLIB_NAME_mbedtls) || defined(MBEDTLS_RSA_NO_CRT)`
so that these test case are still embedded when using libtomcryp as crypto lib.

host/xtest/regression_4000.c
```diff
@@ -24,6 +24,9 @@
 
 #include <assert.h>
 
+#if defined(CFG_CRYPTOLIB_NAME_mbedtls)
```
jforissier, line 27: This variable is not exported by https://github.com/OP-TEE/optee_os/blob/3.13.0/ta/ta.mk#L28-L42 so how does this work?

etienne-lms, line 27: i think you can remove the conditioning.

etienne-lms: host/xtest do include _.../export-ta_arm[32|64]/host_include/conf.h_

jenswi-linaro:
I'm not sure this is such a good idea. It's better to focus on the root of the problem in my opinion.

host/xtest/regression_4000.c
jforissier: @etienne-lms ah, yes, everything is exported under host_include (as `.mk` and `.h`). Thanks for the reminder :)

host/xtest/regression_4000.c
lubing521: libtomcryp support RSA with CRT and RSA without CRT ，libtomcryp can handle both situations well，
but mbedtls Alternative support through macro MBEDTLS_RSA_NO_CRT control. xtest 4006 some case no provide CRT will run fail when enable mbedtls with CRT in optee-os.

> I'm not sure this is such a good idea. It's better to focus on the root of the problem in my opinion.

regression_4000.c:3904: ta_crypt_cmd_asymmetric_decrypt(c, &session, op, ((void *)0), 0, out_enc, out_enc_size, out, &out_size) has an unexpected value: 0xffff0006 = TEEC_ERROR_BAD_PARAMETERS, expected 0x0 = TEEC_SUCCESS
  regression_4006.1 FAILED
  regression_4006 FAILED

if enable mbedtls with CRT, it will check CRT as below code and return MBEDTLS_ERR_RSA_BAD_INPUT_DATA

in function 
rsa_check_context

```
 188 #if !defined(MBEDTLS_RSA_NO_CRT)
 189     /* Modular exponentiation for P and Q is only
 190      * used for private key operations and if CRT
 191      * is used. */
 192     if( is_priv &&
 193         ( mbedtls_mpi_cmp_int( &ctx->P, 0 ) <= 0 ||
 194           mbedtls_mpi_get_bit( &ctx->P, 0 ) == 0 ||
 195           mbedtls_mpi_cmp_int( &ctx->Q, 0 ) <= 0 ||
 196           mbedtls_mpi_get_bit( &ctx->Q, 0 ) == 0  ) )
 197     {
 198         return( MBEDTLS_ERR_RSA_BAD_INPUT_DATA );
 199     }
 200 #endif /* !MBEDTLS_RSA_NO_CRT */
```

lubing521: Such modifications are more understandable,thanks.

host/xtest/regression_4000.c
lubing521: ok

host/xtest/regression_4000.c
jenswi-linaro: Yes, we've discussed that problem earlier. The problem is that OP-TEE cannot provide a consistent ABI with `MBEDTLS_RSA_NO_CRT` disabled. The job for xtest is to among other things highlight such problems.

By the way, how is the pull request to upstream mbedtls progressing?

lubing521: reply as below
“As I mentioned above, we need to schedule some resources to move this PR forward. This usually takes several months.“

detail 
https://github.com/ARMmbed/mbedtls/pull/4742

jenswi-linaro: Given that undefining `MBEDTLS_RSA_NO_CRT` is likely to break the GP tests it seems unlikely that we can accept such a change upstream.

lubing521: is there anything holding this PR back? 
Do I need to close this PR？
thanks.
@jenswi-linaro @etienne-lms @jforissier

jenswi-linaro: > is there anything holding this PR back?

I suppose you missed:
>> Given that undefining `MBEDTLS_RSA_NO_CRT` is likely to break the GP tests it seems unlikely that we can accept such a change upstream.

lubing521: I don't think this patch will break GP testing,it only reduce xtest-4006 without CRT test  when mbedtls with CRT enable in  optee-os. Other scenes were not affected.
I think that It is not possible to support this scenario by supporting dynamic checking of CRT parameters in the short term.
Because there's too much code involved.

jenswi-linaro: This repository is about testing upstream OP-TEE OS. If we undefine `MBEDTLS_RSA_NO_CRT` when compiling OP-TEE OS it will not pass the GP tests so we're not likely to merge such a patch, hence no need for this either.

lubing521: ok ，thanks.

lubing521:
This patch is not necessary, only for its effect mbedtls with CRT in optee-os .

PR #534, opened 2021-07-08
xtest: Add test cases for RSA support

vesajaaskelainen:
Adds test cases for:
- PKCS#1 v1.5 signing & verification support with several hash modes
- PSS signing & verification support
- OAEP encryption/decryption support

Relates to:
- https://github.com/OP-TEE/optee_os/issues/4283

Related PRs:
-

### DIFF
--- a/host/xtest/pkcs11_1000.c
+++ b/host/xtest/pkcs11_1000.c
@@ -6814,3 +6814,372 @@ close_lib:
 }
 ADBG_CASE_DEFINE(pkcs11, 1022, xtest_pkcs11_test_1022,
 		 "PKCS11: RSA PSS key generation and signing");
+
+static const char rsa_oaep_message[] = "Hello World";
+static char rsa_oaep_label[] = "TestLabel";
+
+#define RSA_OAEP_CRYPT_TEST(_test_name, _min_rsa_bits, _hash_algo, _mgf_algo, \
+			    _source_data, _source_data_len) \
+	{ \
+		.test_name = _test_name, \
+		.min_rsa_bits = _min_rsa_bits, \
+		.hash_algo = _hash_algo, \
+		.mgf_algo = _mgf_algo, \
+		.source_data = _source_data, \
+		.source_data_len = _source_data_len, \
+	}
+
+/* List of RSA OAEP crypto params to test out */
+static struct {
+	const char *test_name;
+	uint32_t min_rsa_bits;
+	CK_MECHANISM_TYPE hash_algo;
+	CK_RSA_PKCS_MGF_TYPE mgf_algo;
+	void *source_data;
+	size_t source_data_len;
+} rsa_oaep_crypt_tests[] = {
+	RSA_OAEP_CRYPT_TEST("RSA-OAEP/SHA1", 1024, CKM_SHA_1, CKG_MGF1_SHA1,
+			    NULL, 0),
+	RSA_OAEP_CRYPT_TEST("RSA-OAEP/SHA1/label", 1024, CKM_SHA_1,
+			    CKG_MGF1_SHA1, rsa_oaep_label,
+			    sizeof(rsa_oaep_label)),
+	RSA_OAEP_CRYPT_TEST("RSA-OAEP/SHA224", 1024, CKM_SHA224,
+			    CKG_MGF1_SHA224, NULL, 0),
+	RSA_OAEP_CRYPT_TEST("RSA-OAEP/SHA224/label", 1024, CKM_SHA224,
+			    CKG_MGF1_SHA224, rsa_oaep_label,
+			    sizeof(rsa_oaep_label)),
+	RSA_OAEP_CRYPT_TEST("RSA-OAEP/SHA256", 1024, CKM_SHA256,
+			    CKG_MGF1_SHA256, NULL, 0),
+	RSA_OAEP_CRYPT_TEST("RSA-OAEP/SHA256/label", 1024, CKM_SHA256,
+			    CKG_MGF1_SHA256, rsa_oaep_label,
+			    sizeof(rsa_oaep_label)),
+	RSA_OAEP_CRYPT_TEST("RSA-OAEP/SHA384", 1024, CKM_SHA384,
+			    CKG_MGF1_SHA384, NULL, 0),
+	RSA_OAEP_CRYPT_TEST("RSA-OAEP/SHA384/label", 1024, CKM_SHA384,
+			    CKG_MGF1_SHA384, rsa_oaep_label,
+			    sizeof(rsa_oaep_label)),
+	RSA_OAEP_CRYPT_TEST("RSA-OAEP/SHA512", 2048, CKM_SHA512,
+			    CKG_MGF1_SHA512, NULL, 0),
+	RSA_OAEP_CRYPT_TEST("RSA-OAEP/SHA512/label", 2048, CKM_SHA512,
+			    CKG_MGF1_SHA512, rsa_oaep_label,
+			    sizeof(rsa_oaep_label)),
+};
+
+static int test_rsa_oaep_operations(ADBG_Case_t *c,
+				    CK_SESSION_HANDLE session,
+				    const char *rsa_name, uint32_t rsa_bits)
+{
+	CK_RV rv = CKR_GENERAL_ERROR;
+	CK_OBJECT_HANDLE public_key = CK_INVALID_HANDLE;
+	CK_OBJECT_HANDLE private_key = CK_INVALID_HANDLE;
+
+	CK_MECHANISM mechanism = {
+		CKM_RSA_PKCS_KEY_PAIR_GEN, NULL, 0
+	};
+	CK_MECHANISM crypt_mechanism = {
+		CKM_RSA_PKCS_OAEP, NULL, 0
+	};
+	CK_RSA_PKCS_OAEP_PARAMS oaep_params = {
+		CKM_SHA256, CKG_MGF1_SHA256, CKZ_DATA_SPECIFIED, NULL, 0
+	};
+	CK_BYTE public_exponent[] = { 1, 0, 1 };
+	CK_BYTE id[] = { 123 };
+	CK_ULONG modulus_bits = 0;
+	CK_ATTRIBUTE public_key_template[] = {
+		{ CKA_ENCRYPT, &(CK_BBOOL){ CK_TRUE }, sizeof(CK_BBOOL) },
+		{ CKA_VERIFY, &(CK_BBOOL){ CK_TRUE }, sizeof(CK_BBOOL) },
+		{ CKA_WRAP, &(CK_BBOOL){ CK_FALSE }, sizeof(CK_BBOOL) },
+		{ CKA_MODULUS_BITS, &modulus_bits, sizeof(CK_ULONG) },
+		{ CKA_PUBLIC_EXPONENT, public_exponent,
+		  sizeof(public_exponent) }
+	};
+	CK_ATTRIBUTE private_key_template[] = {
+		{ CKA_TOKEN, &(CK_BBOOL){ CK_FALSE }, sizeof(CK_BBOOL) },
+		{ CKA_PRIVATE, &(CK_BBOOL){ CK_TRUE }, sizeof(CK_BBOOL) },
+		{ CKA_SUBJECT, subject_common_name,
+		  sizeof(subject_common_name) },
+		{ CKA_ID, id, sizeof(id) },
+		{ CKA_SENSITIVE, &(CK_BBOOL){ CK_TRUE }, sizeof(CK_BBOOL) },
+		{ CKA_DECRYPT, &(CK_BBOOL){ CK_TRUE }, sizeof(CK_BBOOL) },
+		{ CKA_SIGN, &(CK_BBOOL){ CK_TRUE }, sizeof(CK_BBOOL) },
+		{ CKA_UNWRAP, &(CK_BBOOL){ CK_FALSE }, sizeof(CK_BBOOL) }
+	};
+
+	CK_OBJECT_CLASS g_class = 0;
+	CK_KEY_TYPE g_key_type = 0;
+	CK_BYTE g_id[32] = { 0 };
+	CK_DATE g_start_date = { 0 };
+	CK_DATE g_end_date = { 0 };
+	CK_BBOOL g_derive = CK_FALSE;
+	CK_BBOOL g_local = CK_FALSE;
+	CK_MECHANISM_TYPE g_keygen_mecha = 0;
+	CK_BYTE g_subject[64] = { 0 };
+	CK_BBOOL g_encrypt = CK_FALSE;
+	CK_BBOOL g_verify = CK_FALSE;
+	CK_BBOOL g_verify_recover = CK_FALSE;
+	CK_BBOOL g_wrap = CK_FALSE;
+	CK_BBOOL g_trusted = CK_FALSE;
+	CK_BYTE g_public_key_info[1024] = { 0 };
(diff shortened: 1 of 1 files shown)


vesajaaskelainen:
Can add more tests here -- just tell what kind of test would be beneficial.

More complex tests start to need more openssl or so support.

host/xtest/pkcs11_1000.c
```diff
@@ -6099,3 +6099,348 @@ static void xtest_pkcs11_test_1020(ADBG_Case_t *c)
...
+		{ CKA_ALWAYS_AUTHENTICATE, &g_always_authenticate, sizeof(CK_BBOOL) },
+		{ CKA_PUBLIC_KEY_INFO, g_public_key_info, sizeof(g_public_key_info) },
+	};
+
+	uint8_t signature[512] = { 0 };
+	CK_ULONG signature_len = 0;
+
+	size_t i = 0;
+
+	Do_ADBG_BeginSubCase(c, "%s: Generate key pair", rsa_name);
```
etienne-lms, line 6244: maybe add `rsa_bits` info to the log trace.

host/xtest/pkcs11_1000.c
```diff
@@ -6099,3 +6099,348 @@ static void xtest_pkcs11_test_1020(ADBG_Case_t *c)
...
+			       ARRAY_SIZE(private_key_template),
+			       &public_key, &private_key);
+	if (!ADBG_EXPECT_CK_OK(c, rv))
+		goto err;
+
+	/* reset get public key template */
+	memset(g_id, 0, sizeof(g_id));
+	memset(g_subject, 0, sizeof(g_subject));
+	memset(g_public_key_info, 0, sizeof(g_public_key_info));
+	get_public_template[2].ulValueLen = sizeof(g_id);
```
etienne-lms, line 6260: Better ensure the indices are the right ones.
I would suggest:
```c
	memset(g_id, 0, sizeof(g_id));
	assert(get_public_template[2].pValue == g_id);
	get_public_template[2].ulValueLen = sizeof(g_id);

	memset(g_subject, 0, sizeof(g_subject));
	assert(get_public_template[2].pValue == g_subject);
	get_public_template[8].ulValueLen = sizeof(g_subject);

	memset(g_public_key_info, 0, sizeof(g_public_key_info));
	assert(get_public_template[2].pValue == g_public_key_info);
	get_public_template[14].ulValueLen = sizeof(g_public_key_info);

File: host/xtest/pkcs11_1000.c [hunk 3]
```diff
@@ -6444,3 +6444,373 @@ static void xtest_pkcs11_test_1021(ADBG_Case_t *c)
...
+		signature_len = sizeof(signature);
+
+		rv = C_SignInit(session, &sign_mechanism, private_key);
+		if (rsa_bits >= rsa_pss_sign_tests[i].min_rsa_bits) {
+			if (!ADBG_EXPECT_CK_OK(c, rv))
+				goto err_destr_obj;
+		} else {
+			if (!ADBG_EXPECT_CK_RESULT(c, CKR_KEY_SIZE_RANGE, rv))
+				goto err_destr_obj;
+			continue;
```
etienne-lms, line 6721: `Do_ADBG_EndSubCase()` missing before `continue`

host/xtest/pkcs11_1000.c
```diff
@@ -6444,3 +6444,373 @@ static void xtest_pkcs11_test_1021(ADBG_Case_t *c)
...
+		rv = C_Verify(session, (void *)rsa_pss_sign_tests[i].data,
+			      rsa_pss_sign_tests[i].data_size,
+			      (void *)signature, signature_len);
+		if (!ADBG_EXPECT_CK_OK(c, rv))
+			goto err_destr_obj;
+
+		Do_ADBG_EndSubCase(c, NULL);
+	}
+
+	Do_ADBG_BeginSubCase(c, "%s: Destroy keys", rsa_name);
```
etienne-lms, line 6743: could remove the subcase for object destruction.

etienne-lms (commented):
Minor comments only.
As for commits header line, prefer it mentions `pkcs11`, i.e. `xtest: pkcs11: add pkcs#1 RSA ... tests`.

vesajaaskelainen:
> As for commits header line, prefer it mentions `pkcs11`, i.e. `xtest: pkcs11: add pkcs#1 RSA ... tests`.

OK

host/xtest/pkcs11_1000.c
vesajaaskelainen: It is part of `rsa_name` eg. `RSA-2048` so left it out to remove duplication.

host/xtest/pkcs11_1000.c
vesajaaskelainen: OK.

host/xtest/pkcs11_1000.c
vesajaaskelainen: OK

host/xtest/pkcs11_1000.c
vesajaaskelainen: Right. Fixing.

host/xtest/pkcs11_1000.c
vesajaaskelainen: Updated to all commits with similar logic.

host/xtest/pkcs11_1000.c
vesajaaskelainen: Fixed to all commits with similar flow.

host/xtest/pkcs11_1000.c
vesajaaskelainen: Fixed to all commits.

vesajaaskelainen:
Updated the PR based on comments.

vesajaaskelainen:
Keep alive.

vesajaaskelainen:
Just rebased the PR.

etienne-lms (commented):
Comments for commit "xtest: pkcs11: Add PKCS #1 v1.5 RSA signing tests".
Comments apply to the 2 other commits.
LGTM once addressed.

host/xtest/pkcs11_1000.c
```diff
@@ -6106,3 +6107,354 @@ static void xtest_pkcs11_test_1020(ADBG_Case_t *c)
...
+
+#define RSA_SIGN_TEST(_test_name, _mecha, _data) \
+	{ \
+		.test_name = _test_name, \
+		.mecha = _mecha, \
+		.data = _data, \
+		.data_size = sizeof(_data) - 1, \
+	}
+
+/* List of RSA PKCS signing multi stage digest mechas */
```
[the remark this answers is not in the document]
vesajaaskelainen: OK

host/xtest/pkcs11_1000.c
```diff
@@ -6106,3 +6107,354 @@ static void xtest_pkcs11_test_1020(ADBG_Case_t *c)
...
+	CK_BBOOL g_sign = CK_FALSE;
+	CK_BBOOL g_sign_recover = CK_FALSE;
+	CK_BBOOL g_unwrap = CK_FALSE;
+	CK_BBOOL g_extract = CK_FALSE;
+	CK_BBOOL g_asensitive = CK_FALSE;
+	CK_BBOOL g_nextract = CK_FALSE;
+	CK_BBOOL g_wrap_with_trusted = CK_FALSE;
+	CK_BBOOL g_always_authenticate = CK_FALSE;
+
+	CK_ATTRIBUTE get_public_template[] = {
```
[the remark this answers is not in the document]
vesajaaskelainen: I'll add something similar.

host/xtest/pkcs11_1000.c
```diff
@@ -6106,3 +6107,354 @@ static void xtest_pkcs11_test_1020(ADBG_Case_t *c)
...
+			      (void *)signature, signature_len);
+		if (!ADBG_EXPECT_CK_OK(c, rv))
+			goto err_destr_obj;
+
+		Do_ADBG_EndSubCase(c, NULL);
+	}
+
+	rv = C_DestroyObject(session, private_key);
+	if (!ADBG_EXPECT_CK_OK(c, rv))
+		goto err_destr_obj;
```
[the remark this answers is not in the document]
vesajaaskelainen: OK

vesajaaskelainen:
Updated the PR with all known comments addressed.

host/xtest/pkcs11_1000.c
```diff
@@ -6106,3 +6107,357 @@ static void xtest_pkcs11_test_1020(ADBG_Case_t *c)
...
+		    &signature_len);
+	if (!ADBG_EXPECT_CK_OK(c, rv))
+		goto err_destr_obj;
+
+	rv = C_VerifyInit(session, &sign_mechanism, public_key);
+	if (!ADBG_EXPECT_CK_OK(c, rv))
+		goto err_destr_obj;
+
+	rv = C_Verify(session, (void *)digest_test_pattern_sha256,
+		    sizeof(digest_test_pattern_sha256), (void *)signature,
```
etienne-lms, line 6354: indentation (add 2 space chars)

host/xtest/pkcs11_1000.c
```diff
@@ -6106,3 +6107,357 @@ static void xtest_pkcs11_test_1020(ADBG_Case_t *c)
...
+			      (void *)signature, signature_len);
+		if (!ADBG_EXPECT_CK_OK(c, rv))
+			goto err_destr_obj;
+
+		Do_ADBG_EndSubCase(c, NULL);
+	}
+
+	rv = C_DestroyObject(session, private_key);
+	if (!ADBG_EXPECT_CK_OK(c, rv))
+		goto err_destr_pub_obj;
```
etienne-lms, line 6395: This (at next `goto` error case) will call `Do_ADBG_EndSubCase(c, NULL);` whereas it should not.

I would rather suggest to call `Do_ADBG_EndSubCase(c, NULL);` at loop beginning (instead of lines 6359 and 6390) and also call it right before `return 1;` at line 6400.

etienne-lms (commented):
Still a comments on error case sequence, applies to the 3 commits.

host/xtest/pkcs11_1000.c
vesajaaskelainen: right.

host/xtest/pkcs11_1000.c
vesajaaskelainen: OK

vesajaaskelainen: Added comment there also for reasoning so that idea is not lost why it looks a bit weird in order.

vesajaaskelainen:
Updated the PR with comments addressed.

etienne-lms:

jforissier:
`Do_ADBG_EndSubCase()` complains a bit in the IBART log, this can probably be fixed
As for the IBART failure, are we reaching a timeout perhaps? How long do the new test add to the overall execution time?

vesajaaskelainen:
> `Do_ADBG_EndSubCase()` complains a bit in the IBART log, this can probably be fixed
> As for the IBART failure, are we reaching a timeout perhaps? How long do the new test add to the overall execution time?

Missed this.

Need to check what new IBART run gives.

Speed is all about how fast it gets random numbers to generate keys. I filtered out over 2k keys to have somewhat fast time.

vesajaaskelainen:
> `Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>`

Thanks for the review -- Applied tags. Will check what the IBART warning was.

host/xtest/pkcs11_1000.c
```diff
@@ -6106,3 +6107,1129 @@ static void xtest_pkcs11_test_1020(ADBG_Case_t *c)
...
+		signature_len = sizeof(signature);
+
+		rv = C_SignInit(session, &sign_mechanism, private_key);
+		if (rsa_bits >= rsa_pss_sign_tests[i].min_rsa_bits) {
+			if (!ADBG_EXPECT_CK_OK(c, rv))
+				goto err_destr_obj;
+		} else {
+			if (!ADBG_EXPECT_CK_RESULT(c, CKR_KEY_SIZE_RANGE, rv))
+				goto err_destr_obj;
+			Do_ADBG_EndSubCase(c, NULL);
```
etienne-lms, line 6758: IBART complaint: this `Do_ADBG_EndSubCase()` must be removed since recent changes in this P-R.

vesajaaskelainen: I suppose it is two? -- anyway pushed the fixes for that.

I was wondering why I didn't see it thou -- could be that my PKCS11 TA debug level is rather verbose.

jbech-linaro:
Are we doing anything in here that makes the QEMU test run longer? I see a timeout at IBART/QEMU.

vchong:
> Are we doing anything in here that makes the QEMU test run longer? I see a timeout at IBART/QEMU.

Friendly reminder (without looking at the patches) that RSA key operations have always taken a long time.

vesajaaskelainen:
> Are we doing anything in here that makes the QEMU test run longer? I see a timeout at IBART/QEMU.

Yes we are -- we are generating bunch of RSA keys and then sing & verify operations also takes some extra time. Especially in QEMU it can be slow, in hardware with necessary RNG support it is better.

Some larger keys are filtered out in normal test invocation (like 3k & 4k RSA). Similar filteration was made for ECDSA keys. Most of the time is spent on OP-TEE's crypto library side.

Could we split testing to two parallel test jobs so the timeout for whole would not be so big issue?

jforissier:
> Could we split testing to two parallel test jobs so the timeout for whole would not be so big issue?

Or run the PKCS#11 tests on the HiKey board instead of QEMU ?

vchong:
> > Could we split testing to two parallel test jobs so the timeout for whole would not be so big issue?
> 
> Or run the PKCS#11 tests on the HiKey board instead of QEMU ?

Afaict none of these will help as in both cases we still only have a single instance into the core (software crypto library) where most of the time is spent.

vchong:
> > > Could we split testing to two parallel test jobs so the timeout for whole would not be so big issue?
> > 
> > 
> > Or run the PKCS#11 tests on the HiKey board instead of QEMU ?
> 
> Afaict none of these will help as in both cases we still only have a single instance into the core (software crypto library) where most of the time is spent.

Oh, if the HiKey jobs on the CI loop have no timeout then perhaps that might be ok.

jbech-linaro:
> Oh, if the HiKey jobs on the CI loop have no timeout then perhaps that might be ok.

Hikey has a timeout, but it was set differently compared to QEMU and allowed longer runs for xtest before timing out. I've just changed QEMU to the same and re-queued the build.

Two years ago we did a full cycle for a target in 6-7 minutes, now we're approaching 20-25 minutes per target. Not a problem as such with current activity. But sometimes I do see things 3-4 builds being queued up. The timeout detection in IBART is a bit crude as well. If it doesn't get back to the shell prompt in X seconds, then it times out. Nowadays X is a high number for building and testing. It would be better if it timeout after X seconds with no activity on the UART. But, I haven't been able to figure out how to implement that with pyexpect. Anyhow, now I'm derailing the discussion :)

vesajaaskelainen:
> > Could we split testing to two parallel test jobs so the timeout for whole would not be so big issue?
> 
> Or run the PKCS#11 tests on the HiKey board instead of QEMU ?

My gut feeling is that it is better to run same test suite in as many hardware as possible so you can catch different kind of problems (like 32bit vs 64bit issues, different crypto hardware, different memory configurations etc.). But perhaps some kind of fast smoke gate could be first and then trigger more heavy weight jobs? Or then just scale the hardware setup but that of course requires investment.

I have noticed that QEMU speed varies quite dramatically based on what kind of hardware it gets executed -- like on my home desktop in native qemu it is much faster than on my work laptop in vmware.

PR #545, opened 2021-09-21
regression 1010: test zero-size TEE_Malloc()

jforissier:
Check that TEE_Malloc() returns a non-NULL pointer when requested size
is zero and that the returned pointer is inaccessible for read or write.
Testing feature introduced in optee_os commit a83ee50a8047 ("libutee:
Handle zero sized buffer allocations") [1].

Link: https://github.com/OP-TEE/optee_os/commit/a83ee50a8047

### DIFF
--- a/ta/os_test/os_test.c
+++ b/ta/os_test/os_test.c
@@ -988,6 +988,8 @@ TEE_Result ta_entry_bad_mem_access(uint32_t param_types, TEE_Param params[4])
 	long int stack = 0;
 	long int stack_addr = (long int)&stack;
 	void (*volatile null_fn_ptr)(void) = NULL;
+	char *zero_size_malloc = NULL;
+	volatile char c = 0;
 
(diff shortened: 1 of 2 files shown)


ta/os_test/os_test.c
```diff
@@ -988,6 +988,8 @@ TEE_Result ta_entry_bad_mem_access(uint32_t param_types, TEE_Param params[4])
 	long int stack = 0;
 	long int stack_addr = (long int)&stack;
 	void (*volatile null_fn_ptr)(void) = NULL;
+	char *zero_size_malloc = NULL;
+	volatile char c = 0;
```
etienne-lms, line 992: why `volatile`?

jforissier: Because otherwise the read `c = *zero_size_malloc;` is optimized out by the compiler... (the value is not used)

etienne-lms:

PR #554, opened 2021-10-20
CMake Python fix

jforissier:

### DIFF
--- a/host/xtest/CMakeLists.txt
+++ b/host/xtest/CMakeLists.txt
@@ -16,10 +16,7 @@ if(OPENSSL_FOUND)
 	set (OPENSSL_PRIVATE_LINK OpenSSL::Crypto)
 endif()
 
-find_package(Python COMPONENTS Interpreter Development)
-if(NOT Python_Interpreter_FOUND)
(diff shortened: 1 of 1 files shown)


jforissier:
> https://dev.azure.com/OPTEE/optee_os/_build/results?buildId=1324&view=logs&j=705748f3-7146-5e86-79af-1c0266d20a8c&t=d3458a7d-1a81-5639-8e63-8927d807e53a&l=3646
> 
> I am a bit confused as the above test wasn't part of the regular checks before my PR was merged.

You're right, the Azure CI runs for PRs in `optee_os`. This repository (`optee_test`) runs IBART only so more limited testing.

Sadiq1811:
find_package() for threads used the REQUIRED keyword and it also has the custom error message.
I was just wondering if we need to maintain consistency here.

jforissier:
> find_package() for threads used the REQUIRED keyword and it also has the custom error message. I was just wondering if we need to maintain consistency here.

REQUIRED stops processing if the package is not found so I do not see how the error message could ever be printed. It should be removed IMO.

PR #560, opened 2021-11-09
[RFC 2] xtest: add regression_1035 to test user TA attestation

jforissier:
Add test cases for user TA attestation. At this point, we check
that:
- Two consecutive hashing of the same TA yield the same hash,
- This hash value is also returned after reloading the TA,
- A different hash is obtained if an additional shared library is mapped
into the TA memory space via dlopen()/dlsym().

The two hashing methods defined by the attestation PTA are tested.

TODO: add authentication code (nonce, signature verification).


jforissier:
Tests for https://github.com/OP-TEE/optee_os/pull/4972.

PR #570, opened 2021-12-01
[RFC] regression: Add test case 1035 for testing bti

ruchi393:
Check that not having appropriate bti launchpads when using
br or blr results in exception getting generated.

This PR helps in testing the bti support. 
For the cores supporting bti mechanism (ARMv8.5 and above), if the page is marked Guarded (GP bit set), not having a launchpad bti instruction for target addressed with br , blr results in an exception. For cores not supporting bti mechanism, bti instructions are treated as nops.

So we can have following scenarios in testing
With CFG_TA_BTI enabled
1. For cores not supporting BTI mechanism -> having/not having bti doesn't matter as bti's are treated as nops and GP bit is ignored and exception is not generated
2. For cores supporting BTI mechanism -> for pages marked with GP bit i.e CFG_TA_BTI enabled, not having bti would result in exception

So, effectively apart from CFG_TA_BTI, we also need to  know whether core supports the feature BTI or not. This can be determined by reading the register ID_AA64PFR1_EL1. 

In this PR, I am using an added command in the system PTA to get the cpu feature and determine if the core supports bti or not. I am not sure if that is the most appropriate way to do it, hence the RFC. 

For optee_os, the corresponding changes are availabel at
https://github.com/ruchi393/optee_os/tree/bti_pta
[ I have not pushed these changes in the existing PR for BTI for now. Would do so once we figure if this is the right way to do this]

### DIFF
--- a/host/xtest/regression_1000.c
+++ b/host/xtest/regression_1000.c
@@ -30,6 +30,7 @@
 #include <ta_sims_keepalive_test.h>
 #include <ta_concurrent.h>
 #include <ta_tpm_log_test.h>
+#include <ta_bti.h>
 #include <ta_supp_plugin.h>
 #include <sdp_basic.h>
 #include <pta_secstor_ta_mgmt.h>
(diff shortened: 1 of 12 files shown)


jforissier:
Hello @ruchi393,

IMO this test is not very relevant because what you are effectively testing is the BTI feature itself, with manually crafted assembly. But I think we can trust the hardware, if it reports that is supports BTI then it probably does it properly ;-)
I think what we want instead is pass the whole optee_test suite with `CFG_TA_BTI=y` on BTI-capable hardware while making  sure that BTI is actually enabled, i.e., the  GP bit is set on the TA code pages. Not sure how we could test that though.

ta/bti_test/ta_entry.c
```diff
@@ -0,0 +1,151 @@
...
+}
+
+/* Called each time a session is closed */
+void TA_CloseSessionEntryPoint(void *pSessionContext __unused)
+{
+}
+
+static TEE_Result get_cpu_features(uint32_t param_types, TEE_Param params[4])
+{
+	static const TEE_UUID system_uuid = PTA_SYSTEM_UUID;
```
jenswi-linaro, line 43: I'd prefer looking up this using:
``` C
TEE_GetPropertyAsBool(TEE_PROPSET_CURRENT_TA, "optee.ta.bti_enabled", &enabled)
```
and then add another entry to `tee_propset_ta` in `core/tee/tee_svc.c` to handle this property.

jenswi-linaro:
> IMO this test is not very relevant because what you are effectively testing is the BTI feature itself, with manually crafted assembly. But I think we can trust the hardware, if it reports that is supports BTI then it probably does it properly ;-)

One purpose of these tests is to see that we're able to set it up properly too.

> I think what we want instead is pass the whole optee_test suite with `CFG_TA_BTI=y` on BTI-capable hardware while making sure that BTI is actually enabled, i.e., the GP bit is set on the TA code pages. Not sure how we could test that though.

I suppose that's what we're testing in this PR. :-)

ruchi393:
> Hello @ruchi393,
> 
> IMO this test is not very relevant because what you are effectively testing is the BTI feature itself, with manually crafted assembly. But I think we can trust the hardware, if it reports that is supports BTI then it probably does it properly ;-) I think what we want instead is pass the whole optee_test suite with `CFG_TA_BTI=y` on BTI-capable hardware while making sure that BTI is actually enabled, i.e., the GP bit is set on the TA code pages. Not sure how we could test that though.

I agree with you partly here. We may not need all the variety of bti exceptions that I have added with different flavours of bti j and c and calling functions. Perhaps generating an exception for one of the cases should be enough. I think I got carried away there. However I think we should have 1 testcase atleast which checks if exception is getting generated  when CFG_TA_BTI is set and bti launchpad is missing. This would ensure that GP bit is getting set properly for the TA code page because this is what the TEE core does. This is what was the intention of the test case to test. 
The problem which I have mentioned is that ideally CFG_TA_BTI should be enabled for only cores which support this feature i.e  > ARMv8.5. However, how do we make sure of  that. To avoid failures in case of mis-configuration I added the call to system PTA to query if BTI is enabled or not. The easiest thing would be to test this case only if CFG_TA_BTI is set and assume that one who is setting this bit for his architecture knows that they are enabling it only if their hardware supports it.

For the complete xtest, for any xtest if there is any library linked which is not BTI enabled, the BTI bit actually doesn't get enabled. For eg for the tests involving os_test ta, if we compile with WITH_CXX_TESTS=y, BTI doesn't get enabled on this TA pages as the libstdc++.a and libgcc_eh.a linked with the TA are not BTI enabled. So I tested with WITH_CXX_TESTS=n in this case.
What I have tried to do is to manually compile with -force-bti as ld flags for TA and then check the warning messages manually during compilation for 
"warning: BTI turned on by -z force-bti when all inputs do not have BTI in NOTE section"

But I am not sure of a good way to do this automatically.

jforissier:
@ruchi393
> However I think we should have 1 testcase atleast which checks if exception is getting generated when CFG_TA_BTI is set and bti launchpad is missing. This would ensure that GP bit is getting set properly for the TA code page because this is what the TEE core does. This is what was the intention of the test case to test.

Right. I agree. I have no strong opinion on having extensive tests vs. only one case.

@jenswi-linaro 

> > I think what we want instead is pass the whole optee_test suite with `CFG_TA_BTI=y` on BTI-capable hardware while making sure that BTI is actually enabled, i.e., the GP bit is set on the TA code pages. Not sure how we could test that though.
> 
> I suppose that's what we're testing in this PR. :-)

Hmmmm indeed :wink:

host/xtest/regression_1000.c
```diff
@@ -2504,3 +2505,115 @@ static void xtest_tee_test_1034(ADBG_Case_t *c)
 }
 ADBG_CASE_DEFINE(regression, 1034, xtest_tee_test_1034,
 		 "Test loading a large TA");
+
+struct bti_test {
```
jforissier, line 2509: The whole test is not applicable to ARM32 or any other architecture that may later be supported. Should this be guarded by `#ifdef ARM64`? Or even `#ifdef CFG_TA_BTI`?

host/xtest/regression_1000.c
```diff
@@ -30,6 +30,7 @@
 #include <ta_sims_keepalive_test.h>
 #include <ta_concurrent.h>
 #include <ta_tpm_log_test.h>
+#include <ta_bti.h>
```
jforissier, line 33: Minor: `bti` out of context sounds mysterious. `<ta_arm_bti.h>` perhaps? Other occurrences in the test files are probably OK though.

ta/bti_test/sub.mk
```diff
@@ -0,0 +1,4 @@
+global-incdirs-y += include
+
+srcs-y += ta_entry.c
+srcs-y += bti_test.S
```
jforissier, line 4: Only for ARM64

host/xtest/regression_1000.c
ruchi393: Thanks, I will guard it by ARM64

PR #571, opened 2021-12-01
regression 2004: fix invalid assumption

jenswi-linaro:
Fixes the invalid assumption that a UDP packet is delivered to its
destination instantly. Also error out early in case received buffer
sizes doesn't match what's expected.


jenswi-linaro:
https://github.com/OP-TEE/optee_client/pull/292 is also needed to fix the error reported by IBART.

etienne-lms (commented):
in commit log: "(...) received buffer sizes **~~doesn't~~ don't** match what's expected."

jenswi-linaro:
Commit message updated, tag applied.

PR #574, opened 2021-12-17
Disable C++ file compilation WITH_CXX_TESTS=n

ruchi393:
Disable the compilation of C++ files from os_test_lib
and is_test_lib_dl if C++ tests are disabled i.e xtests
are built with option WITH_CXX_TESTS=n.

### DIFF
--- a/ta/os_test_lib/sub.mk
+++ b/ta/os_test_lib/sub.mk
@@ -1,6 +1,8 @@
 cppflags-$(WITH_TLS_TESTS) += -DWITH_TLS_TESTS=1
 global-incdirs-y += include
 srcs-y += os_test_lib.c
+ifeq ($(WITH_CXX_TESTS),y)
 ifneq ($(COMPILER),clang)
 srcs-y += os_test_lib_cxx.cpp
+endif
 endif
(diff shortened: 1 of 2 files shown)


jenswi-linaro:

etienne-lms (commented):
but I think the test on Clang could be removed. Handled in [`common.mk`](https://github.com/OP-TEE/build/blob/c0660ccd81ca4ba56cf47f9522028cfc56a5da88/common.mk#L48-L51)

jforissier:
s/is_test_lib_dl/os_test_lib_dl/

Disable C++ file compilation **when** WITH_CXX_TESTS=n

@etienne-lms yes but what happens if one uses Clang and adds `WITH_CXX_TESTS=y` to the command line? On the other hand even if it fails we could still do what you propose to simplify a bit, and logically speaking it makes sense.

So with or without Etienne's suggestion:

ruchi393:
Thanks for the review. For clarity I would prefer to keep both WITH_CXX_TESTS and clang.

ruchi393:
Commit message modifies and tags applied.

PR #572, opened 2021-12-02
[RFC 2] regression: Add test case 1035 for testing bti

ruchi393:
Check that not having appropriate bti launchpads when using
br or blr results in exception getting generated.

Signed-off-by: Ruchika Gupta <ruchika.gupta@linaro.org>

I have used the TA property to get the CPU architecture property as suggested by @jenswi-linaro in this version of RFC.

optee-os patches are available for adding the property are available at:
https://github.com/ruchi393/optee_os/pull/new/bti_ta_prop

I will push these to optee-os once the PR - https://github.com/OP-TEE/optee_os/pull/4967 gets finalised.

PR #575, opened 2021-12-17
regression: Add test case 1035 for testing bti

ruchi393:
Check that not having appropriate bti launchpads when using
br or blr results in exception getting generated.

Reviewed-by: Jerome Forissier <jerome@forissier.org>


ruchi393:
This PR is a replica of earlier PR with tag applied
https://github.com/OP-TEE/optee_test/pull/572

Wrong repository was used in the earlier PR. Cleaned that up

PR #584, opened 2022-01-20
ta: regression: Add test 1036 to test PAuth

ruchi393:
Test 1036 is added to test the ARM Security extension
featuure PAuth. It is compiled only if CFG_TA_PAUTH is
enabled when compiling optee.


ta/pauth_test/ta_arm_pauth.c
```diff
@@ -0,0 +1,43 @@
...
+			TEE_PROPSET_TEE_IMPLEMENTATION,
+			"org.trustedfirmware.optee.cpu.feat_pauth_implemented",
+			&implemented);
+	if (res != TEE_SUCCESS)
+		return res;
+
+	if (!implemented)
+		return TEE_ERROR_NOT_IMPLEMENTED;
+
+	for (int i = 0; i < 10; i++) {
```
etienne-lms, line 33: Could you add in short inline description to state what this test is expected to do?

ta/pauth_test/ta_entry.c
```diff
@@ -0,0 +1,68 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
```
jenswi-linaro, line 1: Any reason we need a new TA instead of adding this to the os_test TA

ruchi393: I followed the same thing which I did for BTI. Can move it to os_test if you think that is the right place.

ruchi393: Another idea could be to have a single TA for all ARM security extensions i.e combine the BTI and PAUTH together and later add MTE tests also to it.

ruchi393:
PR updated.
One comments is pending on which TA this code should belong to. Will do that after we conclude on it.

ta/pauth_test/ta_entry.c
jenswi-linaro: That's another option. I think there is a value in trying to keep the number of TAs down. So if it's not too inconvenient I'd rather have it all in the os_test TA.

ruchi393:
I have shifted the PAuth test to os-test which seems much more cleaner. Thanks @jenswi-linaro .

For ease, I have created another PR for same. All comments received in the PR have been addressed there
https://github.com/OP-TEE/optee_test/pull/586

jforissier:
Thanks @ruchi393, let's close this PR then.

PR #585, opened 2022-01-25
CFG_SE05X: consider the secure element hardware limitations when running the tests

ldts:
xtest: add CFG_CRYPTO_SE05X
--
The following cflag allows to disable  features not supported by
the SE05X hardware.

xtest: disable SE05X unsupported tests
--
The  SE050/1 secure elements do not support all the cryptographic
operations tested in the regression nor pkcs#11 tests.

Since the test framework wont handle a 'feature not supported' return
code from the secure world, this commit disables those tests that are
not supported by the hardware.

```
pkcs11:
	RSA sign:
		CKM_MD5_RSA_PKCS
	RSA OAEP crypto:
	    	RSA-OAEP/SHA224
		RSA-OAEP/SHA224/label
		RSA-OAEP/SHA256
		RSA-OAEP/SHA256/label
		RSA-OAEP/SHA384
		RSA-OAEP/SHA384/label
		RSA-OAEP/SHA512
		RSA-OAEP/SHA512/label
regression:
	RSA keypair generation:
	    	RSA-256
		RSA-384
		RSA-640
		RSA-768
		RSA-896
```

### DIFF
--- a/host/xtest/CMakeLists.txt
+++ b/host/xtest/CMakeLists.txt
@@ -109,6 +109,10 @@ if (CFG_PKCS11_TA)
 	list (APPEND SRC pkcs11_1000.c)
 endif()
 
+if (CFG_CRYPTO_SE05X)
+	add_compile_options(-DCFG_CRYPTO_SE05X)
+endif()
+
 ################################################################################
 # Built binary
 ################################################################################
--- a/host/xtest/Makefile
+++ b/host/xtest/Makefile
@@ -100,6 +100,11 @@ objs 	:= $(patsubst %.c,$(out-dir)/xtest/%.o, $(srcs))
 ifeq ($(CFG_PKCS11_TA),y)
 CFLAGS += -DCFG_PKCS11_TA
 endif
+
+ifeq ($(CFG_CRYPTO_SE05X),y)
+CFLAGS += -DCFG_CRYPTO_SE05X
+endif
+
 CFLAGS += -I./
 CFLAGS += -I./adbg/include
 CFLAGS += -I../supp_plugin/include
--- a/host/xtest/pkcs11_1000.c
+++ b/host/xtest/pkcs11_1000.c
@@ -6297,8 +6297,10 @@ static struct {
 	const void *data;
 	CK_ULONG data_size;
 } rsa_pkcs_sign_tests[] = {
+#ifndef CFG_CRYPTO_SE05X
 	RSA_SIGN_TEST("CKM_MD5_RSA_PKCS", CKM_MD5_RSA_PKCS,
 		      digest_test_pattern),
+#endif
 	RSA_SIGN_TEST("CKM_SHA1_RSA_PKCS", CKM_SHA1_RSA_PKCS,
 		      digest_test_pattern),
 	RSA_SIGN_TEST("CKM_SHA224_RSA_PKCS", CKM_SHA224_RSA_PKCS,
@@ -7051,6 +7053,7 @@ static struct {
 	RSA_OAEP_CRYPT_TEST("RSA-OAEP/SHA1/label", 1024, CKM_SHA_1,
 			    CKG_MGF1_SHA1, rsa_oaep_label,
 			    sizeof(rsa_oaep_label)),
+#ifndef CFG_CRYPTO_SE05X
 	RSA_OAEP_CRYPT_TEST("RSA-OAEP/SHA224", 1024, CKM_SHA224,
 			    CKG_MGF1_SHA224, NULL, 0),
 	RSA_OAEP_CRYPT_TEST("RSA-OAEP/SHA224/label", 1024, CKM_SHA224,
@@ -7071,6 +7074,7 @@ static struct {
 	RSA_OAEP_CRYPT_TEST("RSA-OAEP/SHA512/label", 2048, CKM_SHA512,
 			    CKG_MGF1_SHA512, rsa_oaep_label,
 			    sizeof(rsa_oaep_label)),
+#endif
 };
 
 static int test_rsa_oaep_operations(ADBG_Case_t *c,
--- a/host/xtest/regression_4000.c
+++ b/host/xtest/regression_4000.c
@@ -4401,12 +4401,16 @@ static void xtest_tee_test_4007_rsa(ADBG_Case_t *c)
 	TEEC_Session session = { };
 	uint32_t ret_orig = 0;
 	static const struct key_types_noparam key_types[] = {
+#ifndef CFG_CRYPTO_SE05X
 		{ 0, "RSA-256", TEE_TYPE_RSA_KEYPAIR, 1, 256, 256 },
 		{ 1, "RSA-384", TEE_TYPE_RSA_KEYPAIR, 1, 384, 384 },
+#endif
 		{ 1, "RSA-512", TEE_TYPE_RSA_KEYPAIR, 1, 512, 512 },
+#ifndef CFG_CRYPTO_SE05X
 		{ 1, "RSA-640", TEE_TYPE_RSA_KEYPAIR, 1, 640, 640 },
 		{ 1, "RSA-768", TEE_TYPE_RSA_KEYPAIR, 1, 768, 768 },
 		{ 1, "RSA-896", TEE_TYPE_RSA_KEYPAIR, 1, 896, 896 },
+#endif
 		{ 1, "RSA-1024", TEE_TYPE_RSA_KEYPAIR, 1, 1024, 1024 },
 		{ 1, "RSA-2048", TEE_TYPE_RSA_KEYPAIR, 1, 2048, 2048 },
 		{ 1, "RSA-3072", TEE_TYPE_RSA_KEYPAIR, 1, 3072, 3072 },


ldts:
Something to note:

In secure elements, private keys are always generated in internal memory and the keypair data reaching the xtests only store the handle instead of the actual data.

Incidentally, deleting the keys via libutee never reaches the crypto layer - which is fortunate in a way: keys created by the xtests are deleted using some funky abstraction (below snippet) that just releases the _bignums_ associated to the keys. 

==> This allows the keypairs to be copied and deleted while the copy remains valid - a pattern used in some tests.

```
static const struct attr_ops attr_ops[] = {
	[ATTR_OPS_INDEX_SECRET] = {
		.from_user = op_attr_secret_value_from_user,
		.to_user = op_attr_secret_value_to_user,
		.to_binary = op_attr_secret_value_to_binary,
		.from_binary = op_attr_secret_value_from_binary,
		.from_obj = op_attr_secret_value_from_obj,
		.free = op_attr_secret_value_clear, /* not a typo */
		.clear = op_attr_secret_value_clear,
	},
	[ATTR_OPS_INDEX_BIGNUM] = {
		.from_user = op_attr_bignum_from_user,
		.to_user = op_attr_bignum_to_user,
		.to_binary = op_attr_bignum_to_binary,
		.from_binary = op_attr_bignum_from_binary,
		.from_obj = op_attr_bignum_from_obj,
		.free = op_attr_bignum_free,
		.clear = op_attr_bignum_clear,
	},
	[ATTR_OPS_INDEX_VALUE] = {
		.from_user = op_attr_value_from_user,
		.to_user = op_attr_value_to_user,
		.to_binary = op_attr_value_to_binary,
		.from_binary = op_attr_value_from_binary,
		.from_obj = op_attr_value_from_obj,
		.free = op_attr_value_clear, /* not a typo */
		.clear = op_attr_value_clear,
	},
};

```

The drawback thought, is that keypairs generated by xtests will never be deleted from the SE050 NVM - so some manual deletion process must be implemented to release the space after the tests run.

Also notice that if something like below is ever implemented in OP-TEE, it will break a number of tests on the SE050

```
diff --git a/core/tee/tee_svc_cryp.c b/core/tee/tee_svc_cryp.c
index 3871e1975..b9f05f173 100644
--- a/core/tee/tee_svc_cryp.c
+++ b/core/tee/tee_svc_cryp.c
@@ -1010,10 +1010,41 @@ TEE_Result syscall_cryp_obj_get_attr(unsigned long obj, unsigned long attr_id,
 void tee_obj_attr_free(struct tee_obj *o)
 {
        const struct tee_cryp_obj_type_props *tp;
+       bool free_bignum = true;
        size_t n;
 
        if (!o->attr)
                return;
+
+       switch (o->info.objectType) {
+       case TEE_TYPE_RSA_PUBLIC_KEY:
+               free_bignum = false;
+               crypto_acipher_free_rsa_public_key(o->attr);
+               break;
+       case TEE_TYPE_RSA_KEYPAIR:
+               free_bignum = false;
+               crypto_acipher_free_rsa_keypair(o->attr);
+               break;
+       case TEE_TYPE_ECDSA_PUBLIC_KEY:
+       case TEE_TYPE_ECDH_PUBLIC_KEY:
+       case TEE_TYPE_SM2_DSA_PUBLIC_KEY:
+       case TEE_TYPE_SM2_PKE_PUBLIC_KEY:
+       case TEE_TYPE_SM2_KEP_PUBLIC_KEY:
+               free_bignum = false;
+               crypto_acipher_free_ecc_public_key(o->attr);
+               break;
+       case TEE_TYPE_ECDSA_KEYPAIR:
+       case TEE_TYPE_ECDH_KEYPAIR:
+       case TEE_TYPE_SM2_DSA_KEYPAIR:
+       case TEE_TYPE_SM2_PKE_KEYPAIR:
+       case TEE_TYPE_SM2_KEP_KEYPAIR:
+               free_bignum = false;
+               crypto_acipher_free_ecc_keypair(o->attr);
+               break;
+       default:
+               break;
+       }
+
        tp = tee_svc_find_type_props(o->info.objectType);
        if (!tp)
                return;
@@ -1021,6 +1052,9 @@ void tee_obj_attr_free(struct tee_obj *o)
        for (n = 0; n < tp->num_type_attrs; n++) {
                const struct tee_cryp_obj_type_attrs *ta = tp->type_attrs + n;
 
+               if (!free_bignum &&
+                   attr_ops[ta->ops_index].free == op_attr_bignum_free)
+                               continue;
                attr_ops[ta->ops_index].free((uint8_t *)o->attr + ta->raw_offs);
        }
 }
```

ldts:
any comments?

jenswi-linaro:
It would be nice to have these limitations with SE05 documented somewhere at https://optee.readthedocs.io/en/latest/

host/xtest/regression_1000.c
```diff
@@ -841,7 +841,9 @@ static void xtest_tee_test_1009(ADBG_Case_t *c)
 {
 	xtest_tee_test_1009_subcase(c, "TEE Wait 0.1s", 100, false);
 	xtest_tee_test_1009_subcase(c, "TEE Wait 0.5s", 500, false);
+#ifndef CFG_CRYPTO_SE05X
```
jforissier, line 844: How is this related to SE05X? You mention "SE05x runs single threaded", do you mean `xtest` doesn't use a compliant pthread library on the platform you're testing?

jforissier (commented):
Almost all the algorithms and key sizes that are disabled here are supposed to be mandatory per the GP spec. Shouldn't we fallback to a software implementation for those?

host/xtest/regression_1000.c
ldts: sorry no I think I didnt document this anywhere: to run the SE05X, we configure CFG_NUM_THREADS=1 which will cause this particular test to fail.

With that we avoid having to synchronize access to the SE05x: ie, when OP-TEE has sent RPC to Linux with an APDU frame and is waiting for a response, we cant allow for any other request to go through...plug and trust will have to get a response first. So at the moment we have this limitation. I dont know, maybe it doesn't  quite justify the conditional compile on this regression_1000? or perhaps the test could check whether OP-TEE has been compiled with just one thread?

jenswi-linaro: So instead of using a mutex you configure the system with a single thread?

ldts: Not the system, OP-TEE. 
And yes, but of course, since the option to do so is presented to the user. 
If that is an issue, then you just remove the option - which incidentally also breaks this test-  and I will prioritize the scheduling among my other deliverables...time is a limited resource as I am sure you know and functionality came before performance in this case.

ldts: btw there is a bit of debugging and a lot of concurrency testing involved in implementing this change to the Plug-and-Trust repository (I am hoping is not a OP-TEE change)..wont be nothing major but yes, I have been postponing it

jenswi-linaro: We don't. Being GP compliant is a high priority in OP-TEE.

jenswi-linaro: That said, I guess we can disable this test when CFG_NUM_THREADS=1.

ldts:
> It would be nice to have these limitations with SE05 documented somewhere at https://optee.readthedocs.io/en/latest/

@jenswi-linaro I am not sure where information about this secure element fits in the structure (no other crypto element is documented (ie, CAAM). could you suggest an area? maybe it needs some new entry - ie, crypto elements?

jenswi-linaro:
> > It would be nice to have these limitations with SE05 documented somewhere at https://optee.readthedocs.io/en/latest/
> 
> @jenswi-linaro I am not sure where information about this secure element fits in the structure (no other crypto element is documented (ie, CAAM). could you suggest an area? maybe it needs some new entry - ie, crypto elements?

Would just before https://optee.readthedocs.io/en/latest/architecture/crypto.html#add-a-new-crypto-implementation make sense?

ldts:
> > > It would be nice to have these limitations with SE05 documented somewhere at https://optee.readthedocs.io/en/latest/
> > 
> > 
> > @jenswi-linaro I am not sure where information about this secure element fits in the structure (no other crypto element is documented (ie, CAAM). could you suggest an area? maybe it needs some new entry - ie, crypto elements?
> 
> Would just before https://optee.readthedocs.io/en/latest/architecture/crypto.html#add-a-new-crypto-implementation make sense?

I am not totally sure because that is what we started doing to integrate this SE (extend **core/lib/** ) as per the documentation; but then we realized that it made more sense to have it as **core/drivers/crypto**  - 

But this is not documented either: how to add software crypto extensions vs hardware. The actual documentation only refers to the software cryptographic support. Maybe we need a section for hardware cryptography and its limitations?

jenswi-linaro:
> Maybe we need a section for hardware cryptography and its limitations?

Sounds good.

ldts:
ok @jenswi-linaro @cneveux @clementfaure . I think at the core should be the crypto driver interface and its limitations (as an example it is not easy to have two crypto drivers: the only way for the SE050 to coexist with the CAAM (so we could benefit from having a HUK) was to make sure that CFG_CRYPTO_DRIVER was **not set** at the time CAAM was being compiled but only later, at the time SE05 was being built)

host/xtest/regression_1000.c
ldts: @jenswi-linaro finally got around to addressing the thread issue in the plug and trust. So removing the need for the SE050.

ldts:
> An entry in the documentation sounds good.

@jenswi-linaro  https://github.com/OP-TEE/optee_docs/pull/155 

I will then create a section for the SE05x to expand as I go.

jenswi-linaro:

ldts:
@jforissier @jenswi-linaro added tags

PR #586, opened 2022-01-28
[v2] ta: regression: Add test 1036 to test PAuth

ruchi393:
Added the PAuth test in os_test TA as suggested.

Test 1036 is added to test the ARM Security extension
feature PAuth (Pointer Authentication).

### DIFF
--- a/host/xtest/regression_1000.c
+++ b/host/xtest/regression_1000.c
@@ -2641,3 +2641,49 @@ out:
 }
 ADBG_CASE_DEFINE(regression, 1035, xtest_tee_test_1035, "Test BTI");
 #endif
+
+static void xtest_tee_test_1036(ADBG_Case_t *c)
+{
+	TEEC_Session session = { };
+	TEEC_Operation op = TEEC_OPERATION_INITIALIZER;
+	uint32_t ret_orig = 0;
+	TEEC_Result res = TEEC_SUCCESS;
+
+	op.paramTypes = TEEC_PARAM_TYPES(TEEC_NONE, TEEC_NONE, TEEC_NONE,
+					 TEEC_NONE);
+
+	if (!ADBG_EXPECT_TEEC_SUCCESS(c,
+			xtest_teec_open_session(&session, &os_test_ta_uuid,
+						NULL, &ret_orig)))
+		return;
+
+	Do_ADBG_BeginSubCase(c, "PAuth NOP test");
+
+	res = TEEC_InvokeCommand(&session, TA_OS_TEST_CMD_PAUTH_NOP, &op,
+				 &ret_orig);
+	if (res == TEEC_ERROR_NOT_SUPPORTED) {
+		Do_ADBG_Log("Binary doesn't support PAuth - skip tests");
+		goto out;
+	}
+
+	if (!ADBG_EXPECT_TEEC_SUCCESS(c, res))
+		goto out;
+
+	Do_ADBG_EndSubCase(c, "PAuth NOP test");
+
+	Do_ADBG_BeginSubCase(c, "PAuth PAC corruption");
+
+	ADBG_EXPECT_TEEC_RESULT(c, TEEC_ERROR_TARGET_DEAD,
+			   TEEC_InvokeCommand(&session,
+					      TA_OS_TEST_CMD_PAUTH_CORRUPT_PAC,
+					      &op, &ret_orig));
+
+	ADBG_EXPECT_TEEC_ERROR_ORIGIN(c, TEEC_ORIGIN_TEE, ret_orig);
+
+	Do_ADBG_EndSubCase(c, "PAuth PAC corruption");
+
+out:
+	TEEC_CloseSession(&session);
+}
+ADBG_CASE_DEFINE(regression, 1036, xtest_tee_test_1036,
+		 "Test PAuth (Pointer Authentication)");
--- a/ta/os_test/pauth_a64.S
+++ b/ta/os_test/pauth_a64.S
@@ -1,0 +1,23 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2022, Linaro Limited
+ */
+
+#include <asm.S>
+
+/*
+ * void corrupt_pac(void)
+ *
+ * The function corrupts a single bit of the PAC to ensure that the
+ * authentication will fail. paciasp calculates and adds the PAC to the
+ * link register and autiasp is used to authenticate the PAC.
+ */
+FUNC corrupt_pac , :
+	paciasp
+	/* Flip a random bit in PAC field of the return address */
+	eor lr, lr , 1 << 58
+	autiasp
+	ret
+END_FUNC corrupt_pac
+
+emit_aarch64_feature_1_and     GNU_PROPERTY_AARCH64_FEATURE_1_BTI
--- a/ta/os_test/sub.mk
+++ b/ta/os_test/sub.mk
@@ -20,3 +20,11 @@ cxxflags-remove-cxx_tests.cpp-y += -pg
 srcs-y += cxx_tests_c.c
 cflags-remove-cxx_tests_c.c-y += -pg
 endif
+
+ifeq ($(sm),ta_arm64)
+srcs-$(CFG_TA_PAUTH) += pauth_a64.S
+srcs-$(CFG_TA_PAUTH) += ta_arm_pauth.c
+# -march=armv8.3-a enables the non-nops instructions for PAC. We are using one
+#  such instruction in the PAC test.
+cflags-$(CFG_TA_PAUTH) += -march=armv8.3-a
+endif
--- a/ta/os_test/ta_arm_pauth.c
+++ b/ta/os_test/ta_arm_pauth.c
@@ -1,0 +1,54 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2022, Linaro Limited
+ */
+
+#include <inttypes.h>
+#include <os_test.h>
+#include <string.h>
+#include <ta_os_test.h>
+#include <tee_internal_api.h>
+
+void corrupt_pac(void);
+
+/* Assuming VA_MASK considering maximum 48 bit of VA */
+#define VA_MASK		0xfffff0000000000
+
+static uint64_t sign_using_keyia(uint64_t ptr)
+{
+	asm volatile("paciza %0 " : "+r" (ptr));
+	return ptr;
+}
+
+TEE_Result ta_entry_pauth_test_nop(void)
+{
+	uint64_t pac = 0;
+	bool implemented = false;
+	TEE_Result res = TEE_ERROR_GENERIC;
+
+	res = TEE_GetPropertyAsBool(
+			TEE_PROPSET_TEE_IMPLEMENTATION,
+			"org.trustedfirmware.optee.cpu.feat_pauth_implemented",
+			&implemented);
+	if (res != TEE_SUCCESS)
+		return res;
+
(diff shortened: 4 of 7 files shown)


host/xtest/regression_1000.c
```diff
@@ -2641,3 +2641,50 @@ static void xtest_tee_test_1035(ADBG_Case_t *c)
...
+						NULL, &ret_orig)))
+		return;
+
+	Do_ADBG_BeginSubCase(c, "PAuth NOP test");
+
+	res = TEEC_InvokeCommand(&session, TA_OS_TEST_CMD_PAUTH_NOP, &op,
+				 &ret_orig);
+	if (res == TEEC_ERROR_NOT_SUPPORTED) {
+		Do_ADBG_Log("Binary doesn't support PAuth - skip tests");
+		Do_ADBG_EndSubCase(c, "PAuth NOP test");
```
jenswi-linaro, line 2666: Note that this isn't strictly needed, but it doesn't harm either.

ta/os_test/pauth_a64.S
```diff
@@ -0,0 +1,16 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2022, Linaro Limited
+ */
+
+#include <asm.S>
+
+FUNC corrupt_pac , :
```
jenswi-linaro, line 8: It could be nice to have the prototype of the C function in a comment above as a reminder of what the caller can expect. Perhaps even a short description of what the function does as I guess most people are not very familiar with the instructions `paciasp` and `autiasp`.

ta/os_test/pauth_a64.S
```diff
@@ -0,0 +1,16 @@
...
+	paciasp
+	/* Flip a random bit in PAC */
```
jenswi-linaro, line 10: I guess this means "Flip a random bit in PAC field of the return address".

ta/os_test/ta_arm_pauth.c
```diff
@@ -0,0 +1,52 @@
...
+			&implemented);
+	if (res != TEE_SUCCESS)
+		return res;
+
+	if (!implemented)
+		return TEE_ERROR_NOT_SUPPORTED;
+
+	/* Check if PAC instruction generates the authentication code */
+	for (int i = 0; i < 10; i++) {
+		asm volatile("paciza %0 " : "+r" (ptr));
```
jenswi-linaro, line 36: Please avoid inline assembly mixed with the code. I guess there's not much generic use of this instruction so putting at some common place might not be needed, but perhaps in a separate function above.
By the way, what is this trying to do? Just reading the description of the instruction in the reference manual still leaves some questions.

ta/os_test/pauth_a64.S
ruchi393: Sure, will add

ta/os_test/pauth_a64.S
ruchi393: Yes.

ta/os_test/ta_arm_pauth.c
ruchi393: Pointer authentication requires an address, modifier and key.
paczia Rd
With paczia the address is in the register Rd, modifier is 0 and key is APIAKEY
The instruction will generate and insert pointer authentication code on address specified in Rd with  modifier 0 and key A. The PAC is inserted in Rd itself.

host/xtest/regression_1000.c
```diff
@@ -2641,3 +2641,50 @@ static void xtest_tee_test_1035(ADBG_Case_t *c)
...
+		goto out;
+	}
+
+	if (!ADBG_EXPECT_TEEC_SUCCESS(c, res))
+		goto out;
```
etienne-lms, line 2671: missing call to `Do_ADBG_EndSubCase()`

host/xtest/regression_1000.c
```diff
@@ -2641,3 +2641,50 @@ static void xtest_tee_test_1035(ADBG_Case_t *c)
...
+
+	Do_ADBG_EndSubCase(c, "PAuth NOP test");
+
+	Do_ADBG_BeginSubCase(c, "PAuth PAC corruption");
+
+	res = TEEC_ERROR_TARGET_DEAD;
+	(void)ADBG_EXPECT_TEEC_RESULT(c, res,
```
etienne-lms, line 2678: minor suggestion:
```c
	ADBG_EXPECT_TEEC_RESULT(c, TEEC_ERROR_TARGET_DEAD,
			TEEC_InvokeCommand(...));

	ADBG_EXPECT_TEEC_ERROR_ORIGIN(c, TEEC_ORIGIN_TEE, ret_orig);

File: host/xtest/regression_1000.c [hunk 7]
```diff
@@ -2641,3 +2641,50 @@ static void xtest_tee_test_1035(ADBG_Case_t *c)
...
+					   &op, &ret_orig));
+
+	(void)ADBG_EXPECT_TEEC_ERROR_ORIGIN(c, TEEC_ORIGIN_TEE, ret_orig);
+
+	Do_ADBG_EndSubCase(c, "PAuth PAC corruption");
+
+out:
+	TEEC_CloseSession(&session);
+}
+ADBG_CASE_DEFINE(regression, 1036, xtest_tee_test_1036, "Test PAuth (Pointer Authentication)");
```
etienne-lms, line 2690: add a line break:
```c
ADBG_CASE_DEFINE(regression, 1036, xtest_tee_test_1036,
		 "Test PAuth (Pointer Authentication)");

File: ta/os_test/ta_arm_pauth.c [hunk 8]
```diff
@@ -0,0 +1,52 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2022, Linaro Limited
+ * All rights reserved.
```
etienne-lms, line 4: remove this line or move at end of previous line

host/xtest/regression_1000.c
jenswi-linaro: As I said above, this isn't strictly needed.

The test framework automatically ends all started subcases when the main test case is done.

host/xtest/regression_1000.c
ruchi393: Thanks. To maintain consistency with the other error for this testcase, will remove Do_ADBG_EndSubCase from here.

ta/os_test/pauth_a64.S
```diff
@@ -5,9 +5,17 @@
 
 #include <asm.S>
 
+
```
etienne-lms, line 8: remove extra empty line

ta/os_test/ta_arm_pauth.c
```diff
@@ -32,11 +37,8 @@ TEE_Result ta_entry_pauth_test_nop(void)
...
 	/* Check if PAC instruction generates the authentication code */
-	for (int i = 0; i < 10; i++) {
-		asm volatile("paciza %0 " : "+r" (ptr));
-		pac |= ptr & VA_MASK;
-		ptr = i;
-	}
+	for (int i = 0; i < 10; i++)
+		pac |= sign_using_keyia(i) & VA_MASK;
 
 	if (implemented && pac)
```
etienne-lms, line 43: Should `pac != 0 ` be checked for each of the _paciza_ instructions in the loop above?

ruchi393: I intentionally didn't add this check. Here we are assuming VA_MASK with 48 bits of VA. Since the VA MASK used over here is not exact, there are minimal chances that authentication code generated may be 0 for the mask used. Hence I repeated it over a range of numbers and ored the result. 

The main purpose of this test case is basically to ensure that pac is enabled and these instructions are not treated as nops.

ta/os_test/sub.mk
```diff
@@ -20,3 +20,11 @@ cxxflags-remove-cxx_tests.cpp-y += -pg
 srcs-y += cxx_tests_c.c
 cflags-remove-cxx_tests_c.c-y += -pg
 endif
+
+ifeq ($(sm),ta_arm64)
```
jenswi-linaro, line 24: Depending on `CFG_TA_PAUTH` being `y` should be enough.

ta/os_test/sub.mk
```diff
@@ -20,3 +20,11 @@ cxxflags-remove-cxx_tests.cpp-y += -pg
...
+srcs-$(CFG_TA_PAUTH) += pauth_a64.S
+srcs-$(CFG_TA_PAUTH) += ta_arm_pauth.c
+# -march=armv8.3-a enables the non-nops instructions for PAC. We are using one
+#  such instruction in the PAC test.
+cflags-$(CFG_TA_PAUTH) += -march=armv8.3-a
```
jenswi-linaro, line 29: Why isn't this always added to `ta_arm64-platform-cflags` when `CFG_TA_PAUTH` is `y`?

ruchi393: The instructions paciasp and aut which are used by compiler when CFG_TA_PAUTH is y are nop instructions. Using them without -march=armv8.3-a doesn't return a compilation error. While the instruction PACIZA used in the testcase here is a non-nop instruction for which the flag armv8.3-a needs to be enabled. If in future we start using non-nop PAuth instruction in TEE, we can introduce the cflag.

jenswi-linaro: I don't understand why the TA itself is supposed to keep track of these flags. I'd expect the TA dev-kit to do that.

ruchi393: I can add it to the ta dev-kit if you think that would be the right place. This is not required usually for TA's using PAuth. It is only required if TA's explicitly use a non-nop PAuth instruction using assembly/inline assembly.

ruchi393: With -mbranch-protection , paciasp and autiasp instructions are added as needed. These instructions are treated as nops when they are executed on architectures < v8.3. So compiler doesn't complain when you use these instructions. However instruction paciza is a non-nop instruction, so if it is being used it won't work on architectures < v8.3 i.e won't be treated as nops. So compiler complains for these instructions without explicit -march flag.

Ideally it would be good to have -march flag specified, however this should be driven by the platform port.

ta/os_test/sub.mk
ruchi393: Ok.

How's the sm set to ta_arm64 ? I tried using CFG_USER_TA_TARGETS=ta_arm32, but that doesn't seem to set this. i remember this worked when i had tested the same with BTI.

jenswi-linaro: In `export-ta_arm64/mk/conf.mk`. You select the devkit correspondingly.

ruchi393: Only CFG_TA_PAUTH is not enough.  PAC would work only with arm64 TA's. If we compile with COMPILE_S_USER=32, issues will be observed in compilation.

jenswi-linaro: It seems a bit strange to have a feature enabled which isn't supported even at build time.
But there isn't much we can do about that here. Let's keep it as it is for now.

ruchi393: Yes, we did discuss this when BTI was being added. This seemed the most feasible solution.

ruchi393:
Should I squash the patches ?

jenswi-linaro:
> Should I squash the patches ?

Please do.

ruchi393:
All review commits have been squashed together, @jenswi-linaro , @etienne-lms please have a ook

jenswi-linaro (commented):

ruchi393:
Thanks, PR updated with A-by tag

jforissier:
@etienne-lms any feedback?

PR #587, opened 2022-02-03
ecdh1 derive

ldts:
Given two set of keypairs, test deriving a shared secret over an
unsecure medium.

Tested using the SE051 and libsofthsm.2

Signed-off-by: Jorge Ramirez-Ortiz <jorge@foundries.io>

### DIFF
--- a/host/xtest/pkcs11_1000.c
+++ b/host/xtest/pkcs11_1000.c
@@ -7793,3 +7793,204 @@ close_lib:
 }
 ADBG_CASE_DEFINE(pkcs11, 1024, xtest_pkcs11_test_1024,
 		 "PKCS11: X509 Certificate operations");
+
+static void test_ecdh1_derive(ADBG_Case_t *c, CK_SESSION_HANDLE session,
+			      const char *curve_name, uint8_t *curve,
+			      size_t curve_size, uint32_t algo,
+			      size_t key_bytes)
+{
+	CK_RV rv = CKR_GENERAL_ERROR;
+	CK_MECHANISM keypairgen_mechanism = {
+		CKM_EC_KEY_PAIR_GEN, NULL, 0
+	};
+	CK_MECHANISM derive_mechanism = {
+		CKM_ECDH1_DERIVE, NULL, 0
+	};
+	CK_BYTE id[] = { 123 };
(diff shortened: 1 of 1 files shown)


ldts:
seems that it needs https://github.com/OP-TEE/optee_client/pull/298

etienne-lms:
Some minor comments.

Regarding the tests, it checks a same key is derived from keys. Better than nothing but is it enough for validating ECDH?
If you're patient, i can plan to look at porting [`run_xtest_tee_test_4218()`](https://github.com/etienne-lms/optee_test/blob/d7deefe021610b4113eaf8ec50d48058fbcae2d7/host/xtest/regression_4000.c#L6890-L6900) to a pkcs11_xxxx dedicated test. But i won't be able to look at that in the short term.

etienne-lms (commented):
(here are the minor comments)

ldts:
> Some minor comments.
> 
> Regarding the tests, it checks a same key is derived from keys. Better than nothing but is it enough for validating ECDH? If you're patient, i can plan to look at porting [`run_xtest_tee_test_4218()`](https://github.com/etienne-lms/optee_test/blob/d7deefe021610b4113eaf8ec50d48058fbcae2d7/host/xtest/regression_4000.c#L6890-L6900) to a pkcs11_xxxx dedicated test. But i won't be able to look at that in the short term.

um I thought this sort of test - closer to the real use case - would be a better fit hence why I went this way: it tests the feature while also shows an example of usage. 

But sure, I can drop it if you want - I only did it to help unblock the ECDH PR really....what do you suggest we do? I am ok either way

ldts:
My point has always been that the chances of a shared secret of this size being derived from different **generated** keypairs being equal are pretty slim unless the algorithm is correct. 
Maybe slightly inferior to the chances of an _erroneous_ ECDH1 derive implementation matching the shared secret against a predefined result.
So really we are just working with probabilities.
But still, if you could let me know if you want this PR closed let me know - I don't have any personal attachment to it :)

ldts:
have any tests for ECDH been merged yet? if not, when?

jenswi-linaro:
None. It seems you abandoned this PR. What do you think we should do? Merge stuff when there's no more activity even if there's no consensus or decision?

PR #590, opened 2022-03-09
regression 4006, 4014: do not set TEE_ATTR_ECC_CURVE for SM2

jforissier:
Attribute TEE_ATTR_ECC_CURVE does not apply to SM2 keys. Do not set it
when generating or populating such key objects.

### DIFF
--- a/host/xtest/regression_4000.c
+++ b/host/xtest/regression_4000.c
@@ -3600,6 +3600,8 @@ static bool create_key(ADBG_Case_t *c, TEEC_Session *s,
 	return true;
 }
 
+#define XTEST_NO_CURVE 0xFFFFFFFF /* implementation-defined as per GP spec */
+
 static void xtest_tee_test_4006(ADBG_Case_t *c)
 {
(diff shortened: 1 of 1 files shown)


jforissier:
Depends on https://github.com/OP-TEE/optee_os/pull/5212.

host/xtest/regression_4000.c
```diff
@@ -3600,6 +3600,8 @@ static bool create_key(ADBG_Case_t *c, TEEC_Session *s,
 	return true;
 }
 
+#define XTEST_NO_CURVE 0xFFFFFFFF /* implementation-defined as per GP spec */
```
etienne-lms, line 3603: this is a bit rude, but does the job

etienne-lms (commented):

host/xtest/regression_4000.c
jforissier: Yes ;-) this is local to this file so I considered not worth it to introduce a more complex mechanism.

PR #596, opened 2022-04-22
regression_6000.c: detect available storage IDs at runtime

jforissier:
The current code relies on compile-time defines (CFG_REE_FS and
CFG_RPMB_FS) to select the storage IDs that should be used when running
the secure storage tests (TEE_STORAGE_PRIVATE, TEE_STORAGE_PRIVATE_REE,
TEE_STORAGE_PRIVATE_RPMB). This doesn't work well when xtest is run with
a TEE core that was not built with the same defines. Although that is
not a generally supported thing, the storage differences can be handled
by adding a runtime check.

Some platforms don't have access to an RPMB while others may use only
RPMB (and disable the REE storage). By adding runtime detection, and
assuming other features are the same on all platforms, we can use the
same xtest binary in a common root FS image. One exmaple is the
Linaro LEDGE Reference Platform: the root FS [1] is shared while the
firmware [2] is per-platform.

Link: [1] https://github.com/Linaro/ledge-oe-manifest
Link: [2] https://git.codelinaro.org/linaro/dependable-boot/meta-ts

### DIFF
--- a/host/xtest/regression_6000.c
+++ b/host/xtest/regression_6000.c
@@ -18,27 +18,25 @@
 #include <tee_api_types.h>
 #include <util.h>
 
-#define DEFINE_TEST_MULTIPLE_STORAGE_IDS(test_name)			     \
-static void test_name(ADBG_Case_t *c)					     \
-{									     \
-	size_t i;							     \
-									     \
-	for (i = 0; i < ARRAY_SIZE(storage_ids); i++) {			     \
-		Do_ADBG_BeginSubCase(c, "Storage id: %08x", storage_ids[i]); \
-		test_name##_single(c, storage_ids[i]);			     \
-		Do_ADBG_EndSubCase(c, "Storage id: %08x", storage_ids[i]);   \
-	}								     \
+#define DEFINE_TEST_MULTIPLE_STORAGE_IDS(test_name)		 \
+static void test_name(ADBG_Case_t *c)				 \
+{								 \
+	size_t i = 0;						 \
+								 \
+	if (init_storage_info()) {				 \
+		Do_ADBG_Log("init_storage_info() failed");	 \
+		return;						 \
+	}							 \
+	for (i = 0; i < ARRAY_SIZE(storage_info); i++) {	 \
+		uint32_t id = storage_info[i].id;		 \
+								 \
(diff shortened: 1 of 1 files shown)


jenswi-linaro:

host/xtest/regression_6000.c
```diff
@@ -445,6 +422,118 @@ static TEEC_Result fs_get_obj_info(TEEC_Session *sess, uint32_t obj,
 	return TEEC_InvokeCommand(sess, TA_STORAGE_CMD_GET_OBJ_INFO, &op, &org);
 }
 
+/* Record availability of all  secure storage types at runtime */
```
etienne-lms, line 425: nit: 2 space char

host/xtest/regression_6000.c
```diff
@@ -445,6 +422,118 @@ static TEEC_Result fs_get_obj_info(TEEC_Session *sess, uint32_t obj,
...
+struct storage_info {
+	uint32_t id;
+	bool available;
+};
+
+static struct storage_info storage_info[] = {
+	{ .id = TEE_STORAGE_PRIVATE,  },
```
etienne-lms, line 432: ditto (and can remove the inner coma)

host/xtest/regression_6000.c
```diff
@@ -445,6 +422,118 @@ static TEEC_Result fs_get_obj_info(TEEC_Session *sess, uint32_t obj,
...
+	case TEEC_SUCCESS:
+		*avail = true;
+		break;
+	case TEE_ERROR_ITEM_NOT_FOUND:
+	case TEE_ERROR_STORAGE_NOT_AVAILABLE:
+	case TEE_ERROR_STORAGE_NOT_AVAILABLE_2:
+		*avail = false;
+		break;
+	default:
+		break;
```
etienne-lms, line 462: is this case expected? shouldn't this returns an error at line 470?

jforissier:
@etienne-lms all comments addressed, thanks.

etienne-lms (commented):

jforissier:
IBART seems to have run OK but the status was still shown as "Running" on GitHub CI so I had to manually cancel the job.

PR #600, opened 2022-05-04
GP: Pass upper 32 bits of TEE_PropSetHandle in value.b

jenswi-linaro:
With MTE enabled 64-bit pointer usually use the upper 32 bits too while
the GP tests assumes that 32 bits are enough. Fix this by passing the
upper 32 bits in value.b


jforissier (approved):
`Tested-by: Jerome Forissier <jerome.forissier@linaro.org> (vexpress-qemu_armv8a gp_200{11,36})`

jenswi-linaro:
Tags applied

PR #603, opened 2022-05-24
Makefile: fix ownership contamination when installing output files

clementfaure:
When installing output files with the `install` Makefile target, the
output file ownership is preserved because the file copy to the
destination directory is done with `cp -a/--archive`.

When using the `install` Makefile target in a Yocto recipe, it triggers
the following Bitbake host contamination warning:

WARNING: optee-test-3.17.0.imx-r0 do_package_qa: QA Issue: optee-test: /lib/optee_armtz/b689f2a7-8adf-477a-9f99-32e90c0ad0a2.ta is owned by uid 1001, which is the same as the user running bitbake. This may be due to host contamination
optee-test: /lib/optee_armtz/5ce0c432-0ab0-40e5-a056-782ca0e6aba2.ta is owned by uid 1001, which is the same as the user running bitbake.  This may be due to host contamination

`cp` instead of `cp -a` solves the issue.

Fixes: 2a1ef2c76f ("Add install target")
Signed-off-by: Clement Faure <clement.faure@nxp.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,10 +75,10 @@ checkpatch-staging:
 install:
 	$(echo) '  INSTALL ${DESTDIR}/lib/optee_armtz'
 	$(q)mkdir -p ${DESTDIR}/lib/optee_armtz
-	$(q)find $(out-dir) -name \*.ta -exec cp -a {} ${DESTDIR}/lib/optee_armtz \;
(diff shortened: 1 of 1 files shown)


jforissier (approved):

clementfaure:
@jforissier Do you what was the original reason for having `cp -a` instead of `cp` ?

clementfaure:
Ok, I see :D
Thanks for the quick review. Tags applied!

PR #604, opened 2022-06-04
xtest: exclude test cases when CFG_MEMTAG not set

hoiho-amzn:

### DIFF
--- a/host/xtest/regression_1000.c
+++ b/host/xtest/regression_1000.c
@@ -3091,6 +3091,7 @@ static void xtest_tee_test_1037(ADBG_Case_t *c)
 ADBG_CASE_DEFINE(regression, 1037, xtest_tee_test_1037,
 		 "Remote attestation");
 
+#if defined(CFG_MEMTAG)
 static void xtest_tee_test_1038(ADBG_Case_t *c)
 {
 	TEEC_Session session = { };
@@ -3193,3 +3194,4 @@ out:
 }
 ADBG_CASE_DEFINE(regression, 1038, xtest_tee_test_1038,
 		 "Test MTE (Memory Tag Extension)");
+#endif
--- a/ta/crypt/cryp_taf.c
+++ b/ta/crypt/cryp_taf.c
@@ -7,7 +7,9 @@
 
 #include <assert.h>
 #include <cryp_taf.h>
+#if defined(CFG_MEMTAG)
 #include <memtag.h>
+#endif
 #include <tee_internal_api.h>
 #include <trace.h>
 #include <user_ta_header.h>
--- a/ta/os_test/os_test.c
+++ b/ta/os_test/os_test.c
@@ -7,7 +7,9 @@
 #include <compiler.h>
 #include <dlfcn.h>
 #include <link.h>
+#if defined(CFG_MEMTAG)
 #include <memtag.h>
+#endif
 #include <setjmp.h>
 #include <stdint.h>
 #include <string.h>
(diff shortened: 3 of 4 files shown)


ta/crypt/cryp_taf.c
```diff
@@ -7,7 +7,9 @@
 
 #include <assert.h>
 #include <cryp_taf.h>
+#if defined(CFG_MEMTAG)
 #include <memtag.h>
```
hoiho-amzn, line 11: I am not sure why this needs to be included - none of the code in this file calls any `memtag_*` calls.

host/xtest/regression_1000.c
```diff
@@ -3091,6 +3091,7 @@ static void xtest_tee_test_1037(ADBG_Case_t *c)
 ADBG_CASE_DEFINE(regression, 1037, xtest_tee_test_1037,
 		 "Remote attestation");
 
+#if defined(CFG_MEMTAG)
```
jforissier, line 3094: Please do a runtime check instead, same as the attestation test (see line 3067) so that xtest doesn't depend on the OP-TEE configuration too much. Historically we have used ifdefs but there are use cases to avoid that (common root FS with the test suite, used for different platforms that may or may not enable the same optee_os options).

ta/crypt/cryp_taf.c
jforissier: Then just remove it.

host/xtest/regression_1000.c
hoiho-amzn: Ack - still have to use ifdefs for the TA side though.

jforissier: Why?

jforissier: Sorry my reference to line 3067 was incorrect, in fact I don't understand what you are trying to accomplish. The code is already supposed to skip the MTE tests if MTE is not supported.

hoiho-amzn: I got compilation like this in `os_test.c`:

```
| os_test.c: In function 'ta_entry_memtag_use_after_free':
| os_test.c:1473:14: error: implicit declaration of function 'memtag_is_enabled' [-Werror=implicit-function-declaration]
|  1473 |         if (!memtag_is_enabled())
|       |              ^~~~~~~~~~~~~~~~~
| os_test.c:1473:14: warning: nested extern declaration of 'memtag_is_enabled' [-Wnested-externs]
| os_test.c: In function 'ta_entry_memtag_invalid_tag':
| os_test.c:1493:14: error: implicit declaration of function 'memtag_get_tag' [-Werror=implicit-function-declaration]
|  1493 |         if (!memtag_get_tag(p)) {
|       |              ^~~~~~~~~~~~~~
| os_test.c:1493:14: warning: nested extern declaration of 'memtag_get_tag' [-Wnested-externs]
| os_test.c:1499:14: error: implicit declaration of function 'memtag_insert_tag' [-Werror=implicit-function-declaration]
|  1499 |         p2 = memtag_insert_tag(p, (memtag_get_tag(p) + 1) & 0xf);
|       |              ^~~~~~~~~~~~~~~~~
| os_test.c:1499:14: warning: nested extern declaration of 'memtag_insert_tag' [-Wnested-externs]
| os_test.c:1499:12: warning: assignment to 'uint32_t *' {aka 'unsigned int *'} from 'int' makes pointer from integer without a cast [-Wint-conversion]
|  1499 |         p2 = memtag_insert_tag(p, (memtag_get_tag(p) + 1) & 0xf);
|       |            ^
```

hoiho-amzn: Okay, I think I got it - I was using `3.17.0` of `optee-os`, but `optee-test` was on latest so there is a mismatch.

jforissier: No worries, these things happen! :wink:

PR #605, opened 2022-06-28
regression 4007_x25519: do not fail if X25519 is not supported

jforissier:
When OP-TEE is built with MBedTLS as the core crypto library, X25519 is
not supported. Do not fail "xest regression_4007_x25519" in that case.
Note that a similar check is done in regression_4015 already.

Fixes: 00b3f2cbf4ad ("Add x25519 test cases")


jenswi-linaro:

etienne-lms (commented):

PR #606, opened 2022-07-08
Xilinx/AMD Versal ACAD support

ldts:
With these changes and https://github.com/OP-TEE/optee_os/pull/5426  all xtests (regression and pkcs11) pass on the Versal ACAD platform

### DIFF
--- a/host/xtest/CMakeLists.txt
+++ b/host/xtest/CMakeLists.txt
@@ -113,6 +113,10 @@ if (CFG_CRYPTO_SE05X)
 	add_compile_options(-DCFG_CRYPTO_SE05X)
 endif()
 
+if (CFG_CRYPTO_VERSAL)
+	add_compile_options(-DCFG_CRYPTO_VERSAL)
+endif()
+
 ################################################################################
 # Built binary
 ################################################################################
--- a/host/xtest/Makefile
+++ b/host/xtest/Makefile
@@ -105,6 +105,10 @@ ifeq ($(CFG_CRYPTO_SE05X),y)
 CFLAGS += -DCFG_CRYPTO_SE05X
 endif
 
+ifeq ($(CFG_CRYPTO_VERSAL),y)
+CFLAGS += -DCFG_CRYPTO_VERSAL
+endif
+
 CFLAGS += -I./
 CFLAGS += -I./adbg/include
 CFLAGS += -I../supp_plugin/include
--- a/host/xtest/regression_4000.c
+++ b/host/xtest/regression_4000.c
@@ -2426,7 +2426,7 @@ static const struct xtest_ae_case ae_cases[] = {
 	XTEST_AE_CASE_AES_CCM(vect1, 3, 2),
 	XTEST_AE_CASE_AES_CCM(vect2, 7, 13),
 	XTEST_AE_CASE_AES_CCM(vect3, 5, 21),
-
+#ifndef CFG_CRYPTO_VERSAL
 	XTEST_AE_CASE_AES_GCM(vect1, 0, 0, NULL_ARRAY, NULL_ARRAY, NULL_ARRAY),
 	XTEST_AE_CASE_AES_GCM(vect2, 0, 9, NULL_ARRAY, ARRAY, ARRAY),
 	XTEST_AE_CASE_AES_GCM(vect3, 0, 9, NULL_ARRAY, ARRAY, ARRAY),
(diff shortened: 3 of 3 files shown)


jforissier:
I'd like to avoid adding compile-time switches and prefer run-time selection instead (rationale mentioned in https://github.com/OP-TEE/optee_test/commit/88fe494ce3289d1152930e878e483f4f6b634b82). In this case you could split 4005 into 4005_ccm and 4005_gcm, then to exclude the GCM tests you would run `xtest -x 4005_gcm`. Note we have similar code already: `4007_rsa`,`4007_dh` etc.

ldts:
@jforissier right I can split the test but how do I then implement the runtime detection? AFAIK there is not an interface for the platform to announce its cryptographic capabilities so it will still need to be resolved at build time, wont it?

jforissier:
@ldts `ta_crypt_cmd_is_algo_supported(..., TEE_ALG_AES_GCM, TEE_CRYPTO_ELEMENT_NONE)` perhaps?

ldts:
@jforissier no unfortunately not: it doesnt have enough granularity on the definition of support: AES_GCM is supported - in fact it is widely used by OP-TEE core and implemented and delivered by the versal/authenc.c driver. So the driver does work as expected.

However what is not supported are the special tests implemented by 4005 (unaligned sizes and such). Hence why we need to disable 4005..or at least until either the PLM supports it. r we add the necessary maths to the versal OP-TEE driver.

ldts:
@jforissier we had this same discussion for the SE05x btw: algorithms being supported but only some use cases (ie key sizes) run by the xtests not supported.
hence why I proposed a way to have a pta disable certain xtests.

jforissier:
Then use the `xtest -x` option to exclude the tests you expect to fail on your platform.

jforissier:
> sure. the thing is, how do we communicate that to users without them having to read documents which they might even not know they have to read? because a user will run xtest and think there is something broken in the platform...

There **is** something broken in the platform. The GP spec makes no difference between aligned and unaligned buffers.

ldts:
um, I dont know, I guess it depends...one wanting to build on it and add functionality while keeping the users informed could argue that is not _broken_, but a use case that is not implemented.  It could also be argued that perhaps it would be helpful if the xtest - since it is not a certification tool - could pass the information to the user instead of just failing.
But sure, a user that hits the problem and receives no feedback will see it as broken.

The way I see it, a pta that calls a platform interface to query the list of unsupported xtests would be a good development tool trivial to implement....or we could do nothing.

jenswi-linaro:
Wouldn't it be easier to just fix the problem in the implementation instead using temporary buffers?

ldts:
not sure what we are talking about anymore :) ; I mean it is pretty obvious: not easier - or I would have done it - but preferable, absolutely. But it is always a combination of time and responsibilities. Not everyone can afford the luxury of spending all of its time time on a single code base. you know that.
This is really about how xtests should work not about how a particular case: if my understanding is right you are saying that xtests does not care about controlled failures and its main responsibility is to report red/green (which is acceptable I guess). IMO it could do more and be more flexible and help the user a bit more: but it does is necessary but is it sufficient?
just need clarification on this really

jenswi-linaro:
> I mean it is pretty obvious: not easier - or I would have done it - but preferable, absolutely.

I think you may be underestimating how much time and effort is spent on brushing the problem under the carpet. It's not just your own time your spending here.

ldts:
Right, and by reading the way you are qualifying the problem I think you obviously do not understand what I am talking about. Lets leave it at that.

jforissier:
> you are saying that xtests does not care about controlled failures and its main responsibility is to report red/green

`xtest` does have options to select and filter out tests (you haven't replied to my suggestion about using `-x` btw). "`make check`" (see qemu_v8.mk) on the other hand is a pass/fail command suitable for CI and regression tests; it can contain all the platform-specific options if needed.

ldts:
> `xtest` does have options to select and filter out tests (you haven't replied to my suggestion about using `-x` btw). "`make check`" (see qemu_v8.mk) on the other hand is a pass/fail command suitable for CI and regression tests; it can contain all the platform-specific options if needed.

I did. I posted a follow up question that you didnt reply to (perhaps you jumped in to your conclusion without reading my answer). It was right after that brilliant _brushing under the carpet_ comment. But lets move on. I dont sense an interest on  making xtests a bit better so no need to go in circles. Just replying since you took the time.

PR #608, opened 2022-08-09
host: supp_plugin: fix cross compilation

davidson-k:
In host/supp_plugin/Makefile, CC is assigned the cross compiler set in
the Makefile only if the CC is not assigned any value previously. But
according to the GNU make implicit variables, CC will have a default
value of cc and hence the cross compiler set in the Makefile will not be
used while compiling the supp_plugin using makefiles.
"--no-builtin-variables" will cancel all variables used by implicit
rules.

Signed-off-by: Davidson K <davidson.kumaresan@arm.com>


jforissier (commented):

PR #611, opened 2022-08-22
xtest: fix compilation issue

pmoos:
This fixes the following compilation issue:

external/optee_test/host/xtest/regression_1000.c:2817:56: error: unused parameter 'nonce' [-Werror,-Wunused-parameter]
static void check_measurement(ADBG_Case_t *c, uint8_t *nonce, size_t nonce_size,
                                                       ^
external/optee_test/host/xtest/regression_1000.c:2817:70: error: unused parameter 'nonce_size' [-Werror,-Wunused-parameter]
static void check_measurement(ADBG_Case_t *c, uint8_t *nonce, size_t nonce_size,
                                                                     ^

Signed-off-by: Pierre Moos <pmoos@baylibre.com>

### DIFF
--- a/host/xtest/regression_1000.c
+++ b/host/xtest/regression_1000.c
@@ -2817,6 +2817,9 @@ err:
 static void check_measurement(ADBG_Case_t *c, uint8_t *nonce, size_t nonce_size,
 			      uint8_t *hash, uint8_t *buf)
 {
+	(void)nonce;
+	(void)nonce_size;
+
 	if (hash)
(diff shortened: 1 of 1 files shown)


jforissier:
Hi @pmoos,

Thanks for the patch. A slightly cleaner fix would be to change the definition of `check_signature()` to a static inline function and apply the `(void)nonce` etc. trick there, but it would be more complex (especially if we consider changing the two other macros to static inline functions too for consistency...). So:

pmoos:
Thanks for the review @jforissier , I've added your reviewed-by

PR #613, opened 2022-08-24
xtest: fix compilation issue

pmoos:
This fixes the following compilation error:

external/optee_test/host/xtest/sdp_basic.c:687:91: error: unused parameter 'ion_heap' [-Werror,-Wunused-parameter]
static void usage(const char *progname, size_t size, int loop, const char *heap_name, int ion_heap)

Signed-off-by: Pierre Moos <pmoos@baylibre.com>


pmoos:
Thanks :)

PR #615, opened 2022-09-05
[RFC] Signing TAs using subkeys

jenswi-linaro:
Depends on https://github.com/OP-TEE/optee_os/pull/5523


ta/subkey_notes.txt
```diff
@@ -0,0 +1,71 @@
...
+	--key ../../optee_os/keys/default_ta.pem --in top_level_subkey.pem \
+	--out top_level_subkey.bin --max-depth 4 --name-size 64 \
+	--subkey-version 1
+
+The mid level subkey has to have a UUID in the namespace of the top level
+level subkey. The UUID is determined as
+UUIDv5(f04fa996-148a-453c-b037-1dcfbad120a6, "mid_level_subkey") =
+d28416d1-fc99-5762-ac9d-d2311e0b0472
+
+The sign script can do with this with:
```
jforissier, line 39: extra `with`

ta/subkey_notes.txt
```diff
@@ -0,0 +1,71 @@
...
+openssl genrsa -out identity_subkey2.pem
+
+The top level subkey is signed with the root key using (where the UUID is
+generated):
+
+../../optee_os/scripts/sign_encrypt.py sign-subkey \
+	--uuid f04fa996-148a-453c-b037-1dcfbad120a6 \
+	--key ../../optee_os/keys/default_ta.pem --in top_level_subkey.pem \
+	--out top_level_subkey.bin --max-depth 4 --name-size 64 \
+	--subkey-version 1
```
jforissier, line 32: Could `--subkey-version` be omitted and be 1 by default?

jforissier (commented):
A few comments/questions as I'm trying to familiarize with the UUID derivation.

ta/subkey_notes.txt
```diff
@@ -0,0 +1,71 @@
...
+openssl genrsa -out mid_level_subkey.pem
+openssl genrsa -out identity_subkey2.pem
+
+The top level subkey is signed with the root key using (where the UUID is
+generated):
+
+../../optee_os/scripts/sign_encrypt.py sign-subkey \
+	--uuid f04fa996-148a-453c-b037-1dcfbad120a6 \
+	--key ../../optee_os/keys/default_ta.pem --in top_level_subkey.pem \
+	--out top_level_subkey.bin --max-depth 4 --name-size 64 \
```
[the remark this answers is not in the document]
jenswi-linaro: TL;DL; version first. `name-size` is the size of a the string following this subkey.bin, it just have to be large enough. The string is the name of the next subkey or TA and not part of this subkey.bin.

Here's a tricky part. To derive the UUID we need a string. Later when checking that the UUID was indeed derived in the namespace of the previous UUID we need that string again. But to allow one subkey.bin to be used for more than one TA we can't have that string a signed part of this subkey.bin. So `name-size` means that after this subkey.bin there are this many bytes following which holds the string used to derive the UUID of the next element (subkey or TA).

ta/subkey_notes.txt
jenswi-linaro: Yes, it currently defaults to 0.

ta/subkey_notes.txt
```diff
@@ -0,0 +1,71 @@
...
+	--name mid_level_subkey
+Subkey UUID: f04fa996-148a-453c-b037-1dcfbad120a6
+Next subkey UUID: d28416d1-fc99-5762-ac9d-d2311e0b0472
+
+The mid level subkey is signed using:
+
+../../optee_os/scripts/sign_encrypt.py sign-subkey \
+	--uuid d28416d1-fc99-5762-ac9d-d2311e0b0472 \
+	--key top_level_subkey.pem --subkey top_level_subkey.bin \
+	--max-depth 3 --name-size 64 --subkey-version 1 \
```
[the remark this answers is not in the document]
jenswi-linaro: Makes sense.

ta/subkey_notes.txt
jenswi-linaro: To clarify, I'd rather keep the default at 0. I'm using a `1` here instead for testing purposes.

jenswi-linaro:
Replaced by https://github.com/OP-TEE/optee_test/pull/621

PR #625, opened 2022-11-11
xtest: stats: Add loaded ta context dump

weizhaojiang:
OP-TEE supports a debug method to dump user TA's runtime context through stats PTA. The patch is to extend xtest to support it.

Tested-by: Weizhao Jiang <weizhaoj@amazon.com>
Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>
Signed-off-by: Weizhao Jiang <weizhaoj@amazon.com>

### DIFF
--- a/host/xtest/stats.c
+++ b/host/xtest/stats.c
@@ -27,6 +27,7 @@
 #define STATS_CMD_PAGER_STATS	0
 #define STATS_CMD_ALLOC_STATS	1
 #define STATS_CMD_MEMLEAK_STATS	2
+#define STATS_CMD_TA_STATS	3
 
 #define TEE_ALLOCATOR_DESC_LENGTH 32
 struct malloc_stats {
@@ -37,6 +38,13 @@ struct malloc_stats {
 	uint32_t num_alloc_fail;	  /* Number of failed alloc requests */
 	uint32_t biggest_alloc_fail;	  /* Size of biggest failed alloc */
 	uint32_t biggest_alloc_fail_used; /* Alloc bytes when above occurred */
+};
+
+struct ta_dump_stats {
+	TEEC_UUID uuid;
+	uint32_t panicked;	/* True if TA has panicked */
+	uint32_t sess_count;	/* Number of opened session */
+	struct malloc_stats heap;
 };
 
 static const char *stats_progname = "xtest --stats";
@@ -50,6 +58,7 @@ static int usage(void)
 	fprintf(stderr, " --pager        Print pager statistics\n");
 	fprintf(stderr, " --alloc        Print allocation statistics\n");
 	fprintf(stderr, " --memleak      Dump memory leak data on secure console\n");
+	fprintf(stderr, " --ta           Print loaded TAs context\n");
 
 	return EXIT_FAILURE;
 }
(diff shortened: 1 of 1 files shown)


host/xtest/stats.c
```diff
@@ -39,6 +40,13 @@ struct malloc_stats {
 	uint32_t biggest_alloc_fail_used; /* Alloc bytes when above occurred */
 };
 
+struct ta_dump_stats {
+	TEEC_UUID uuid;
+	uint32_t panicked;	/* True if TA has panicked, written from asm */
+	uint32_t sess_num;  /* Number of opened session */
```
jenswi-linaro, line 46: sess_count

host/xtest/stats.c
```diff
@@ -209,6 +218,98 @@ static int stat_memleak(int argc, char *argv[])
...
+		return usage();
+
+	open_sess(&ctx, &sess);
+
+	op.paramTypes = TEEC_PARAM_TYPES(TEEC_MEMREF_TEMP_OUTPUT,
+					 TEEC_NONE, TEEC_NONE, TEEC_NONE);
+	res = TEEC_InvokeCommand(&sess, STATS_CMD_TA_STATS, &op, &eo);
+	if (res != TEEC_ERROR_SHORT_BUFFER) {
+		if (res == TEEC_ERROR_NOT_SUPPORTED)
+			printf("Build OP-TEE OS without CFG_TA_STATS ?= y\n");
```
jenswi-linaro, line 243: I'm not sure this is needed.

host/xtest/stats.c
```diff
@@ -209,6 +218,98 @@ static int stat_memleak(int argc, char *argv[])
...
+
+	op.params[0].tmpref.buffer = stats;
+	op.params[0].tmpref.size = stats_size_bytes;
+	res = TEEC_InvokeCommand(&sess, STATS_CMD_TA_STATS, &op, &eo);
+	if (res)
+		errx(EXIT_FAILURE,
+		     "TEEC_InvokeCommand: res %#"PRIx32" err_orig %#"PRIx32,
+		     res, eo);
+
+	if (op.params[0].tmpref.size != stats_size_bytes)
```
jenswi-linaro, line 271: What if a TA was loaded or unloaded in between?  Perhaps it's better to allocate a bit extra and then accept that we might get a bit less. If it still wasn't enough allocate more and try again. It makes the code a bit more complicated but it will improve the user experience in the odd case.

host/xtest/stats.c
weizhaojiang: done

host/xtest/stats.c
weizhaojiang: This is to reminder user to have "CFG_TA_STATS ?= y" build if they want to run this cmd. I think it's nice to have it.

host/xtest/stats.c
weizhaojiang: Got your point. We can always allocate two more `stats` and if it's still not enough, try 10 times before return error.

host/xtest/stats.c
```diff
@@ -209,6 +218,109 @@ static int stat_memleak(int argc, char *argv[])
...
+	op.params[0].tmpref.size = stats_size_bytes;
+	res = TEEC_InvokeCommand(&sess, STATS_CMD_TA_STATS, &op, &eo);
+	if (res == TEEC_ERROR_SHORT_BUFFER && retry_count > 0) {
+		retry_count --;
+		goto retry;
+	}
+
+	if (res) {
+		printf("TEEC_InvokeCommand: res %#"PRIx32" err_orig %#"PRIx32,
+		       res, eo);
```
etienne-lms, line 284: `goto out`

host/xtest/stats.c
```diff
@@ -39,6 +40,13 @@ struct malloc_stats {
 	uint32_t biggest_alloc_fail_used; /* Alloc bytes when above occurred */
 };
 
+struct ta_dump_stats {
+	TEEC_UUID uuid;
+	uint32_t panicked;	/* True if TA has panicked, written from asm */
```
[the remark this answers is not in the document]
weizhaojiang: done

host/xtest/stats.c
```diff
@@ -50,6 +58,7 @@ static int usage(void)
 	fprintf(stderr, " --pager        Print pager statistics\n");
 	fprintf(stderr, " --alloc        Print allocation statistics\n");
 	fprintf(stderr, " --memleak      Dump memory leak data on secure console\n");
+	fprintf(stderr, " --ta           Print loaded ta context\n");
```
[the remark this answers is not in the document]
weizhaojiang: done

host/xtest/stats.c
```diff
@@ -209,6 +218,98 @@ static int stat_memleak(int argc, char *argv[])
...
+		if (res == TEEC_ERROR_NOT_SUPPORTED)
+			printf("Build OP-TEE OS without CFG_TA_STATS ?= y\n");
+		errx(EXIT_FAILURE,
+		     "TEEC_InvokeCommand: res %#"PRIx32" err_orig %#"PRIx32,
+		     res, eo);
+	}
+
+	stats_size_bytes = op.params[0].tmpref.size;
+	if (stats_size_bytes == 0) {
+		printf("not available loaded TA!\n");
```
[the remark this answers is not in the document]
weizhaojiang: done

host/xtest/stats.c
weizhaojiang: good catch...done.

host/xtest/stats.c
jenswi-linaro: It litters the code in my opinion.

weizhaojiang: okay, done with removing that line printf().

host/xtest/stats.c
```diff
@@ -209,6 +218,107 @@ static int stat_memleak(int argc, char *argv[])
...
+	if (argc != 1)
+		return usage();
+
+	open_sess(&ctx, &sess);
+retry:
+	op.paramTypes = TEEC_PARAM_TYPES(TEEC_MEMREF_TEMP_OUTPUT,
+					 TEEC_NONE, TEEC_NONE, TEEC_NONE);
+	res = TEEC_InvokeCommand(&sess, STATS_CMD_TA_STATS, &op, &eo);
+	if (res != TEEC_ERROR_SHORT_BUFFER) {
+		printf("TEEC_InvokeCommand: res %#"PRIx32" err_orig %#"PRIx32,
```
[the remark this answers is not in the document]
weizhaojiang: refined, thanks.

host/xtest/stats.c
```diff
@@ -209,6 +218,107 @@ static int stat_memleak(int argc, char *argv[])
...
+		printf("realloc(%zu) failed", stats_size_bytes);
+		goto out;
+	}
+	buff = stats;
+
+	op.params[0].tmpref.buffer = stats;
+	op.params[0].tmpref.size = stats_size_bytes;
+	res = TEEC_InvokeCommand(&sess, STATS_CMD_TA_STATS, &op, &eo);
+	if (res == TEEC_ERROR_SHORT_BUFFER && retry_count > 0) {
+		retry_count --;
```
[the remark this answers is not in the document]
weizhaojiang: removed....thanks.

host/xtest/stats.c
```diff
@@ -209,6 +218,107 @@ static int stat_memleak(int argc, char *argv[])
...
+
+	open_sess(&ctx, &sess);
+retry:
+	op.paramTypes = TEEC_PARAM_TYPES(TEEC_MEMREF_TEMP_OUTPUT,
+					 TEEC_NONE, TEEC_NONE, TEEC_NONE);
+	res = TEEC_InvokeCommand(&sess, STATS_CMD_TA_STATS, &op, &eo);
+	if (res != TEEC_ERROR_SHORT_BUFFER) {
+		printf("TEEC_InvokeCommand: res %#"PRIx32" err_orig %#"PRIx32"\n",
+		       res, eo);
+		goto out;
```
etienne-lms, line 246: For all the error case, I think you should report the error. Can use `errx()` as done in other `stat_xxx()` function in this source file. Here:
```
		errx(EXIT_FAILURE,
		     "STATS_CMD_TA_STATS: expected %#"PRIx32", got %#"PRIx32" orig",
		     stats_size_bytes, op.params[1].tmpref.size);

Comment (weizhaojiang) [reply host/xtest/stats.c:246]: `errx()` will call `exit (1);`. But here we might need to `goto out;` and release resource before exit xtest.
I realized that  `errx()` will output log to `stderr`. It seems to me there is no much difference we output to  `stderr` or `stdout`. Like to know which device you would suggest?

**Comment by weizhaojiang** [comment]
@jenswi-linaro @etienne-lms I have updated the commit. Can you revisit the code and let me know any comment? thanks.

File: host/xtest/stats.c (hunk 10 above)
Comment (jenswi-linaro) [reply host/xtest/stats.c:246]: `stderr` is normally unbuffered, while `stdout` might be buffered. Then it's also different file descriptors so if you redirect the output you'll get different results.

I'd prefer using `errx()` too, I believe all the resources are reclaimed when the process exits any way.

Comment (weizhaojiang) [reply host/xtest/stats.c]: thanks, have changed to errx().
like to consult one thing, how the session can be reclaimed when the process exits? When a client open a session, there is corresponding session context -- `struct tee_ta_session` in OP-TEE core nad the session will be put into a link -- `open_sessions.`  When the client exits without calling closeSession(), the session context in OP-TEE core will be reclaimed?

Comment (jenswi-linaro) [reply host/xtest/stats.c]: > When the client exits without calling closeSession(), the session context in OP-TEE core will be reclaimed?

Yes, the context file descriptor is closed `optee_release()` in the kernel driver will be called. This function tries to close all remaining open sessions belonging to this context.

Comment (weizhaojiang) [reply host/xtest/stats.c]: okay, great to know that.

File: host/xtest/stats.c [hunk 11]
```diff
@@ -209,6 +218,99 @@ static int stat_memleak(int argc, char *argv[])
...
+					 TEEC_NONE, TEEC_NONE, TEEC_NONE);
+	res = TEEC_InvokeCommand(&sess, STATS_CMD_TA_STATS, &op, &eo);
+	if (res != TEEC_ERROR_SHORT_BUFFER)
+		errx(EXIT_FAILURE,
+		     "TEEC_InvokeCommand: res %#"PRIx32" err_orig %#"PRIx32,
+		     res, eo);
+
+	stats_size_bytes = op.params[0].tmpref.size;
+	if (stats_size_bytes == 0)
+		errx(EXIT_FAILURE, "No loaded TA found");
```
etienne-lms, line 250: This is not an error case. `errx()` is dedicated to error case so prefer use here `return close_sess(&ctx, &sess);` as you initially implemented.

weizhaojiang: okay, done.

etienne-lms:

weizhaojiang:
> `Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>`

thanks for your reviewing.

etienne-lms:
@weizhaojiang, could you update your pull request by adding the review tag in the commit message, here:
```patch
 xtest: stats: Add loaded ta context dump
 OP-TEE supports a debug method to dump user TA's runtime context
 through stats PTA. The patch is to extend xtest to support it.
 
 Tested-by: Weizhao Jiang <weizhaoj@amazon.com>
 Signed-off-by: Weizhao Jiang <weizhaoj@amazon.com>
+Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>
```

weizhaojiang:
> 

sure, done. Thanks.

PR #629, opened 2022-12-06
Enable compatibility with TEE Internal Core API v1.1

jenswi-linaro:
While the TEE Internal Core API in OP-TEE OS is updated, enable compatibility with the last support version. This is done by compiling xtest with __OPTEE_CORE_API_COMPAT_1_1=1 and including <tee_api_compat.h> too. The TAs enable it with
CFG_TA_OPTEE_CORE_API_COMPAT_1_1==y in ta/ta_common.mk.

Backwards compatibility is enabled for the GP test suite TAs in the same way, using a new patch.

### DIFF
--- a/host/xtest/CMakeLists.txt
+++ b/host/xtest/CMakeLists.txt
@@ -16,6 +16,8 @@ if(OPENSSL_FOUND)
 	set (OPENSSL_PRIVATE_LINK OpenSSL::Crypto)
 endif()
 
+add_compile_options(-D__OPTEE_CORE_API_COMPAT_1_1=1)
+
 find_package(Python REQUIRED)
 
 include(GNUInstallDirs)
--- a/host/xtest/Makefile
+++ b/host/xtest/Makefile
@@ -33,6 +33,8 @@ define cc-bits
 $(if $(filter arm, $(1)),32,$(if $(filter aarch64, $(1)),64,unknown-arch))
 endef
 
+CFLAGS += -D__OPTEE_CORE_API_COMPAT_1_1=1
+
 # OpenSSL is used by:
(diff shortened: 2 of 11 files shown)


jenswi-linaro:
Depends on https://github.com/OP-TEE/optee_os/pull/5688.
Note that "regression 4016: fix TEE_ATTR_EDDSA_PREHASH parameter" is in https://github.com/OP-TEE/optee_test/pull/630 too.

jenswi-linaro:
This is likely to cause problems until "Update reported version to GP Core Internal API v1.3.1" from https://github.com/OP-TEE/optee_os/pull/5688 is merged.

jforissier:
> This is likely to cause problems until "Update reported version to GP Core Internal API v1.3.1" from [OP-TEE/optee_os#5688](https://github.com/OP-TEE/optee_os/pull/5688) is merged.

Oh, I totally missed your word of warning :cry: and the fact that there is no CI doesn't help.
Why does the TA need to `#include <tee_api_compat.h>` at all? Could we get rid of that requirement?

jenswi-linaro:
TAs don't need it, but xtest needs it since it doesn't include `<tee_internal_api.h>`.

jforissier:
Sorry I meant `xtest`. The app is telling the API files which version it expects, not the opposite, right?

jenswi-linaro:
Yes.

jenswi-linaro:
Should I create a new PR in optee_so with only "Update reported version to GP Core Internal API v1.3.1" which we can review quickly? It's not a large patch.

jenswi-linaro:
https://github.com/OP-TEE/optee_os/pull/5689

PR #630, opened 2022-12-06
A fix and some cleanup

jenswi-linaro:

### DIFF
--- a/host/xtest/regression_4000.c
+++ b/host/xtest/regression_4000.c
@@ -5982,7 +5982,7 @@ static void xtest_tee_test_4016_ed25519(ADBG_Case_t *c)
 out:
 	TEEC_CloseSession(&session);
 }
-ADBG_CASE_DEFINE(regression, 4016_ed25519, xtest_tee_test_4016_ed25519,
+ADBG_CASE_DEFINE(regression, 4016, xtest_tee_test_4016_ed25519,
 		 "Test TEE Internal API ED25519 sign/verify");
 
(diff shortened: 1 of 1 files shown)


host/xtest/regression_4000.c
```diff
@@ -5299,6 +5299,7 @@ static void xtest_tee_test_4012(ADBG_Case_t *c)
 }
 ADBG_CASE_DEFINE(regression, 4012, xtest_tee_test_4012,
 		"Test seeding RNG entropy");
+#endif /*CFG_SYSTEM_PTA*/
 
 static void xtest_tee_test_4013(ADBG_Case_t *c)
```
jforissier, line 5304: This one uses `TA_CRYPT_CMD_DERIVE_TA_UNIQUE_KEY{,SHM}` which depend on `CFG_SYSTEM_PTA`

jforissier (commented):
For commit "regression 4013-4016: remove CFG_SYSTEM_PTA dependency": one comment below.

But perhaps it would be better to remove the `CFG_SYSTEM_PTA` conditionals altogether? That is deal gracefully with a possible open session error in test 4012.

jforissier:
For "regression 4016: drop _ed25519 suffix" and "regression 4016: fix TEE_ATTR_EDDSA_PREHASH parameter":

jenswi-linaro:
I'm dropping "regression 4013-4016: remove CFG_SYSTEM_PTA dependency". I'll open a new PR for that one.

Tags applied to the other commits.

PR #635, opened 2023-01-06
riscv: ta: os_test: declare undefined instruction for RISC-V

maroueneboubakri:
RISC-V instruction size is 32-bits, 0x0 triggers an illegal instruction exception.

### DIFF
--- a/ta/os_test/os_test.c
+++ b/ta/os_test/os_test.c
@@ -980,6 +980,8 @@ static void undef_instr(void)
 	__asm__(".word 0x0");
 #elif defined(ARM32)
 	__asm__(".word 0xe7ffffff");
+#elif defined(RV64) || defined(RV32)
+	__asm__(".word 0x0");
 #else
 #error "Unsupported architecture"
 #endif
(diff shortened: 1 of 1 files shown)


jforissier (commented):

ta/os_test/os_test.c
```diff
@@ -980,6 +980,8 @@ static void undef_instr(void)
 	__asm__(".word 0x0");
 #elif defined(ARM32)
 	__asm__(".word 0xe7ffffff");
+#elif defined(RV64) || defined(RV32)
+	__asm__(".word 0x0");
```
etienne-lms, line 984: don't want to mix arm and riscv? :)
```patch
-#if defined(ARM64)
+#if defined(ARM64)|| defined(RV64) || defined(RV32)
 	__asm__(".word 0x0");
 #elif defined(ARM32)
```

etienne-lms (approved):
`Acked-by: Etienne Carriere <etienne.carriere@linaro.org>` (no need to address my comment)

ta/os_test/os_test.c
maroueneboubakri: For better code reading, I suggest to keep it as it.

maroueneboubakri:
Tags applied.

PR #639, opened 2023-01-20
Makefile: use GNU standards for directory variables

clementfaure:


jforissier (commented):

PR #643, opened 2023-01-31
pkcs11: disable 1026 for se05x

igoropaniuk:
Disable RSA AES wrap/unwrap test when CFG_CRYPTO_SE05X=y, as it's currently not supported when SE050/1 secure elements are being used. When CFG_CRYPTO_SE05X=y this test fails:

pkcs11_1026.1 Test RSA AES wrap/unwrap of 128 AES key with 2048 RSA E/TC:? 0
E/TC:? 0 TA panicked with code 0xffff0009
E/LD:  Status of TA 7f10a757-4139-4eae-90c9-f2b2eb118139
E/LD:   arch: arm
E/LD:  region  0: va 0x40002000 pa 0x00000000 size 0x002000 flags rw-s (ldelf)
E/LD:  region  1: va 0x40004000 pa 0x00000000 size 0x006000 flags r-xs (ldelf)
E/LD:  region  2: va 0x4000a000 pa 0x00000000 size 0x001000 flags rw-s (ldelf)
E/LD:  region  3: va 0x4000b000 pa 0x00000000 size 0x004000 flags rw-s (ldelf)
E/LD:  region  4: va 0x4000f000 pa 0x00000000 size 0x001000 flags r--s
E/LD:  region  5: va 0x40010000 pa 0x00000000 size 0x002000 flags rw-s (stack)
E/LD:  region  6: va 0x40012000 pa 0xdceea000 size 0x001000 flags rw-- (param)
E/LD:  region  7: va 0x40013000 pa 0xc980f54c size 0x002000 flags rw-- (param)
E/LD:  region  8: va 0x40058000 pa 0x00001000 size 0x031000 flags r-xs [0]
E/LD:  region  9: va 0x40089000 pa 0x00032000 size 0x011000 flags rw-s [0]
E/LD:   [0] 7f10a757-4139-4eae-90c9-f2b2eb118139 @ 0x40058000

Signed-off-by: Igor Opaniuk <igor.opaniuk@foundries.io>

### DIFF
--- a/host/xtest/pkcs11_1000.c
+++ b/host/xtest/pkcs11_1000.c
@@ -8343,6 +8343,7 @@ out:
 	return rv;
 }
 
+#ifndef CFG_CRYPTO_SE05X
 static void xtest_pkcs11_test_1026(ADBG_Case_t *c)
 {
 	CK_RV rv = CKR_GENERAL_ERROR;
@@ -8467,3 +8468,4 @@ close_lib:
 
 ADBG_CASE_DEFINE(pkcs11, 1026, xtest_pkcs11_test_1026,
 		 "PKCS11: RSA AES Key Wrap/Unwrap tests");
+#endif


host/xtest/pkcs11_1000.c
```diff
@@ -8343,6 +8343,7 @@ static CK_RV test_rsa_aes_wrap(ADBG_Case_t *c, CK_SESSION_HANDLE session,
 	return rv;
 }
 
+#ifndef CFG_CRYPTO_SE05X
```
jforissier, line 8346: I don't think this should appear like that in the test code, because SE05X should be an implementation detail. Either there should be a software fallback (preferred), or we accept that not all implementations support this feature in which case we just accept `TEE_ERROR_NOT_IMPLEMENTED` as a success case.
@ldts what do you think?

ldts: Right, the thing is that it doesn't make a lot of sense IMO to wrap an RSA private key when the secure element is enabled since what is being passed back to criptoki is not a valid RSA private key but a 64bit handle + the public components.

I haven't had the time to investigate why it fails yet; I guess it could be mbedtls trying to parse what it expects to be a private key and fails at it? But I agree with you, rather than just disabling the test, the test should accept a not implemented (or not supported even better) as a pass. not sure how hard doing it will be?

ldts: Also thinking about it, it would NOT be right to disable it in xtests as some Secure Elements will in fact not complain (ie the NXP SE050-E, which doesnt support RSA in the element and does it in software instead).

So you are right, it should not be disabled here.

ldts:
Just for completeness, the tests dont pass because currently the plug and trust does not support TEE_ALG_RSAES_PKCS1_OAEP_MGF1_XXX with XXX different than SHA1.

enabling CFG_NXP_SE05X_RSA_DRV_FALLBACK is ok for this use case.

PR #647, opened 2023-02-04
xtest: fix calling C_Finalize before lib is initialized

MatthewSukyoung:
C_Finalize() is called from some places through close_lib() even before pkcs11 library is initialized. These needs to be replaced by return.

Signed-off-by: Matthew(Sukyoung) Chae <matthewc@axis.com>


MatthewSukyoung:
All comments have been addressed.

etienne-lms (commented):
thanks

PR #648, opened 2023-02-07
xtest: sdp: remove ION support

jforissier:
The ION allocator is deprecated and not present in recent kernels. Remove the code that uses it.

Compile-tested only, for QEMUv8 (make CFG_SECURE_DATA_PATH=y), with a local hack to bring in <linux/dma-heap.h>.


jenswi-linaro:

jforissier:
Hmmmm... so it looks like ion might still be used with the current setup actually... and although we are supposed to be using a 6.2 kernel... but maybe Buildroot uses older header files? @etienne-lms any idea?

etienne-lms:
_linux/dma-heap.h_ was introduced in kernel v5.6.
It looks we didn't update Buildroot config switch when we upgrade the kernel version:
```
build/br-ext/configs/toolchain-br:BR2_KERNEL_HEADERS_4_19=y
```

Aside this build issue, I'm fine with this change: `Acked-by: Etienne Carriere <etienne.carriere@linaro.org>`.

jforissier:
Pushed a fix to hopefully address the non-SDP build (`CFG_SECURE_DATA_PATH != y`). The case with SDP enabled needs a fix in build.git, not sure which one yet -- probably upgrade the toolchains so that they come with a recent enough version of the kernel headers.

jforissier:
> _linux/dma-heap.h_ was introduced in kernel v5.6. It looks we didn't update Buildroot config switch when we upgrade the kernel version:
> 
> ```
> build/br-ext/configs/toolchain-br:BR2_KERNEL_HEADERS_4_19=y
> ```

Unfortunately the Arm toolchains appear to come with very outdated versions of the kernel headers (even the GCC 11.3 toolchain I just tried still comes with 4.19), which means no `<linux/dma-heap.h>`.
```
$ ls -l bin/aarch64-linux-gnu-gcc
lrwxrwxrwx 1 jerome jerome 26 Feb 10 12:23 bin/aarch64-linux-gnu-gcc -> aarch64-none-linux-gnu-gcc
$ ./bin/aarch64-linux-gnu-gcc --version | head -1
aarch64-linux-gnu-gcc (Arm GNU Toolchain 11.3.Rel1) 11.3.1 20220712
$ grep -r "#define LINUX_VERSION_CODE"
aarch64-none-linux-gnu/libc/usr/include/linux/version.h:#define LINUX_VERSION_CODE 267277
```

267277 == 0x04140D so version is 0x04.0x14.0x0D == 4.20.13

jforissier:
Ack tags added, they were given before I added the `#ifdef CFG_SECURE_DATA_PATH` guard on `#include <sdp_basic.h>` but I guess that's fine.

etienne-lms (commented):
s/linuix/linux/ in commit message

PR #650, opened 2023-02-17
xtest: pkcs11: fix 64-bit xtest_pkcs11_1014 failed

408419237:
64-bit xtest unsigned long is 8 bytes,while the
g_len is 4 bytes, crossing the bounds and causing
the contamination variable to g_decrypt

FAILED:
pkcs11_1000.c:3496: Expression "g_decrypt == CK_TRUE" (0 == 1) is false pkcs11_1014.1 FAILED

### DIFF
--- a/host/xtest/pkcs11_1000.c
+++ b/host/xtest/pkcs11_1000.c
@@ -3528,7 +3528,7 @@ static void xtest_pkcs11_test_1014(ADBG_Case_t *c)
 	CK_BBOOL g_decrypt = CK_FALSE;
 	CK_BBOOL g_wrap = CK_FALSE;
 	CK_BBOOL g_unwrap = CK_FALSE;
-	uint32_t g_len = 0;
+	CK_ULONG g_len = 0;
 	CK_ATTRIBUTE get_template[] = {
 		{ CKA_LABEL, (CK_UTF8CHAR_PTR)g_label, sizeof(g_label) },
 		{ CKA_ID, (CK_BYTE_PTR)g_id, sizeof(g_id) },
@@ -3539,7 +3539,7 @@ static void xtest_pkcs11_test_1014(ADBG_Case_t *c)
 		{ CKA_DECRYPT, &g_decrypt, sizeof(CK_BBOOL) },
 		{ CKA_WRAP, &g_wrap, sizeof(CK_BBOOL) },
 		{ CKA_UNWRAP, &g_unwrap, sizeof(CK_BBOOL) },
-		{ CKA_VALUE_LEN, &g_len, sizeof(CK_ULONG) },
+		{ CKA_VALUE_LEN, &g_len, sizeof(g_len) },
 	};
 	CK_ATTRIBUTE set_template[] = {
 		{ CKA_LABEL, (CK_UTF8CHAR_PTR)new_label, strlen(new_label) },


408419237:
32bit xtest_pkcs11 is ok, 64-bit xtest_pkcs11_1014 failed

jforissier (commented):
LGTM.

@etienne-lms is there a reason why `get_template` is not initialized with:
```
        { CKA_VALUE_LEN, &g_len, sizeof(g_len) },
```
...rather than:
```
        { CKA_VALUE_LEN, &g_len, sizeof(CK_ULONG) },
```
...as we normally do? I suppose the implementation would have returned an error making the issue obvious.

jforissier:
First we need to know what type `g_len` needs to be (`uint32_t`, `CK_ULONG` or whatever). This I can't tell because I don't know PKCS#11 well enough. Then in this PR the `sizeof()` needs to be updated accordingly to fix the issue, that's all.

Changing `sizeof(<type>)` to `sizeof(<variable>)` as I suggested should be done for **all** occurrences as a separate commit if we agree it's a good thing to do.

408419237:
Thank you, I have modified it

etienne-lms:
I think the initial change is the right one:
```patch
-	uint32_t g_len = 0;
+	CK_ULONG g_len = 0;
```

The spec defines `CKA_VALUE_LEN` as an attribute for several key objects, always of type `CK_ULONG`, including generic `CKO_SECRET_KEY` objects. https://docs.oasis-open.org/pkcs11/pkcs11-curr/v2.40/pkcs11-curr-v2.40.html, e.g. sections "2.6.2 Generic secret key objects" and "2.8.2 AES secret key objects". Applies also to spec v3.0.

> @etienne-lms is there a reason why get_template is not initialized with:
> `{ CKA_VALUE_LEN, &g_len, sizeof(g_len) },`
> ...rather than:
> `{ CKA_VALUE_LEN, &g_len, sizeof(CK_ULONG) },`
> ...as we normally do? I suppose the implementation would have returned an error making the issue obvious.

The spec defines many types used in such attributes lists but the IDs are untype raw value (defined with `CKx_*` macros). To make sure we use the expected types, we explicitly used `CK_*` type cast and `sizeof()` value in the initialized attributes and equivalent Type-Length-Value structures, in _pkcs11_1000.c_.
```c
	CK_ATTRIBUTE some_object[] = {
		{ CKA_FOO, &(CK_TYPE_RELATED_CKA_FOO){CK_FOO_BAR}, sizeof(CK_TYPE_RELATED_CKA_FOO) },
		...
```

408419237:
cka_value_len
I see that the spec defines CKA_VALUE_LEN as a property of multiple key objects, always CK_ULONG，Thank you, I have modified it.

jforissier:
> I see that the spec defines CKA_VALUE_LEN as a property of multiple key objects, always CK_ULONG

Fine. 

@etienne-lms according to the table any other data type is invalid so does our implementation check that `size == sizeof(CK_ULONG)` when `attribute == CKA_VALUE_LEN`?

etienne-lms:
Yes, it is checked in libckteec, in [`serialize_ck_attribute()`](https://github.com/OP-TEE/optee_client/blob/3.20.0/libckteec/src/serialize_ck.c#L194-L198), based on [`ck_attr_is_ulong()`](https://github.com/OP-TEE/optee_client/blob/3.20.0/libckteec/src/serialize_ck.c#L145).

etienne-lms (commented):

I suggest to change the commit header line:
```
xtest: pkcs11_1014: fix CKA_VALUE_LEN type

Fixes type used to store a CKA_VALUE_LEN attribute value, that is
CK_ULONG as per PKCS#11 specification, no uint32_t.

The issue was found on 64-bit xtest where...
```

408419237:
Thank you, I have modified it,Please help check.

jforissier:
> Yes, it is checked in libckteec, in [`serialize_ck_attribute()`](https://github.com/OP-TEE/optee_client/blob/3.20.0/libckteec/src/serialize_ck.c#L194-L198), based on [`ck_attr_is_ulong()`](https://github.com/OP-TEE/optee_client/blob/3.20.0/libckteec/src/serialize_ck.c#L145).

Very good. Then I think my suggestion for using `sizeof(var)` instead of `sizeof(CK_ULONG)` would be safer and more consistent (no overflow possible and error if type is incorrect). But it's not for this PR anyways.

408419237:
I think sizeof(var) is safer,I looked at other examples，I have modified it，thank you，Please help check.
sizeof

jforissier:

etienne-lms:

PR #651, opened 2023-02-17
xtest: pkcs11: fix 64-bit xtest_pkcs11_1014 failed

408419237:
64-bit xtest unsigned long is 8 bytes,while the
g_len is 4 bytes, crossing the bounds and causing
the contamination variable to g_decrypt

FAILED:
pkcs11_1000.c:3496: Expression "g_decrypt == CK_TRUE" (0 == 1) is false pkcs11_1014.1 FAILED


408419237:
Thank you, I have modified it

jforissier:
Why did you close the previous PR instead of continuing the review there? Now we have lost context of the questions I asked.

PR #656, opened 2023-03-14
regression 1006/1013: skip tests when pager is constrainted

etienne-lms:
Uses PTA stats during test to see when pager is enabled and operates with a small page pool. When so, some tests can be simplified or skiped unless what they are very slow to execute.


jenswi-linaro:

etienne-lms:
Removed an inline comment (in 2nd commit, about a 100kB threshold whereas it's 128kB) and applied @jenswi-linaro tag.

Issue #657, opened 2023-03-24
SO PIN in PKCS#11 tests


jforissier:
Hello,

The PKCS#11 test suite uses a very unusual SO PIN:

https://github.com/OP-TEE/optee_test/blob/5db8ab4c733d5b2f4afac3e9aef0a26634c4b444/host/xtest/pkcs11_1000.c#L569

It is not a valid UTF-8 string as specified by the standard, but most importantly it contains a null byte which means the command-line tool `pkcs11-tool` cannot read this PIN from the `--so-pin` argument. Even worse: the command like `pkcs11-tool --init-token $'\x00\x01\x02\x03\x04\x05\x06\x07\x08\x09\x0a'` --slot-index 2 --label tests` will actually provide an **empty** SO PIN (since 0x00 comes first) which seems to be enabling the "TEE identity based authentication" on the slot (see https://github.com/OP-TEE/optee_os/commit/1a27b197d8b16ef2d4dab8d205f359627b6f0232).

Questions:
- Don't we have a security issue with `CFG_PKCS11_TA_AUTH_TEE_IDENTITY=y`? It looks like anyone can just wipe out a token by re-initializing it with an empty SO PIN at any time using the above command. @vesajaaskelainen 
- How about the default SO/user PINs used in `xtest` are changed to more manageable values such as "12345678", "12345" etc.?

For a bit of background: I'm writing a couple of tests based on `pkcs11-tool` and `openssl` and I'd like to re-use the slot used by xtest, but it is currently impossible due to the above.

etienne-lms:
I think you're right on the pin codes used in xtest, they should be char values `0` `1` ..., not decimal values, so that we can connect a pkcs11 token with standard tools after the tocke has been initialized with xtest.

Regarding the validity of the UTF8 chars used in a pin, the spec says:
>  This standard allows PIN values to contain any valid UTF8 character, but the token may impose subset restrictions.

Value 0 is a valid UTF8 char, no? It represents NUL and it not printable in ASCII, but still a valid UTF8 character.

Regarding `CFG_PKCS11_TA_AUTH_TEE_IDENTITY=y`, it relies on a null PIN, not a PIN starting with a NUL byte. UTF8 does not enforce UTF8 strings to end on NUL byte and not include NUL bytes before UTF8 string end, AFAICT.
[...]

PR #658, opened 2023-03-24
pkcs11: remove NUL byte from SO PIN and use ASCII characters

jforissier:
[Edited]

PKCS#11 PINs are UTF-8 character strings. The default SO PIN used in the
test suite contains a NUL, which causes issues with for example
pkcs11-tool which interprets the NUL byte as the end of the string, thus
making the slot impossible to use from the command line.

Therefore, and for better compatibility with command-line tools after
xtest has been run, use simpler (ASCII) values for both the SO and user
PINs.

### DIFF
--- a/host/xtest/pkcs11_1000.c
+++ b/host/xtest/pkcs11_1000.c
@@ -583,10 +583,9 @@ ADBG_CASE_DEFINE(pkcs11, 1002, xtest_pkcs11_test_1002,
  * Helpers for tests where we must log into the token.
  * These define the genuine PINs and label to be used with the test token.
  */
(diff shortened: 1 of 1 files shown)


jforissier:
Force-pushed to reword the commit message after @etienne-lms [comments](https://github.com/OP-TEE/optee_test/issues/657#issuecomment-1483090559). No code change.

etienne-lms (commented):

vesajaaskelainen:
`Acked-by: Vesa Jääskeläinen <vesa.jaaskelainen@vaisala.com>`

PR #660, opened 2023-03-29
Fix Storage TA ABI TEE_ObjectInfo

jenswi-linaro:
With the updated types for GP 1.3 TEE_ObjectInfo contains elements with the type size_t. If xtest is built for AArch64 while the storage TA is built for AArch32 the ABI will become incompatible. Fix this by adding struct ta_storage_obj_info with fixed width integers only and use that in the TA ABI used by xtest.


jforissier:
I guess that explains the failure I'm seeing on QEMUv8 with `COMPILE_S_USER=32` (namely, xtest 8017.1) :wink: 

jenswi-linaro:
Tag applied

PR #662, opened 2023-04-03
xtest: regression_6016: use 3 threads instead of 4

jforissier:
Since OP-TEE commit cda03b63cf30 ("Enable SHA-3 support by default") [1] regression_6016 fails when OP-TEE OS is compiled for QEMUv8 with "make CFG_ULIBS_SHARED=y CFG_WITH_PAGER=y":

 * regression_6016 Storage concurency o regression_6016.1 Storage id: 00000001 threads: 4, loops: 8 regression_6000.c:1784: xtest_teec_open_session(&arg[m].session, &storage_ta_uuid, ((void *)0), &orig) has an unexpected value: 0xffff000c = TEEC_ERROR_OUT_OF_MEMORY, expected 0x0 = TEEC_SUCCESS

Use 3 threads instead of 4 to reduce the memory footprint.

Link: [1] https://github.com/OP-TEE/optee_os/commit/cda03b63cf304472adce7688a07168f4c66af5c7

### DIFF
--- a/host/xtest/regression_6000.c
+++ b/host/xtest/regression_6000.c
@@ -1770,7 +1770,7 @@ exit:
 }
 
 
-#define NUM_THREADS 4
+#define NUM_THREADS 3
 static void xtest_tee_test_6016_loop(ADBG_Case_t *c, uint32_t storage_id)
 {
 	struct test_6016_thread_arg arg[NUM_THREADS] = { };


etienne-lms (commented):

jenswi-linaro:

jforissier:
> I mean, if `CFG_NUM_THREADS` is 2, which is smaller than `NUM_THREADS` which is 3.
> Will xtest 6016 be broken?

No, it should still work. This test is kind of a stress test to make sure there is no error or crash with concurrent access, nothing more.

gagachang:
> No, it should still work. This test is kind of a stress test to make sure there is no error or crash with concurrent access, nothing more.

I see. Thanks!

PR #665, opened 2023-05-14
ta: pkcs11: test invalid handles on destroyed token object

etienne-lms:
This change reproduces issue https://github.com/OP-TEE/optee_os/issues/6005 where existing object handles on a destroyed token object are not updated and can lead to TA crash.

The issue is fixed with https://github.com/OP-TEE/optee_os/pull/6018 which makes test pkcs11_1028 added by the P-R to succeed.


host/xtest/xtest_main.c
```diff
@@ -166,6 +166,10 @@ int main(int argc, char *argv[])
 #ifdef CFG_SECURE_DATA_PATH
 	else if (argc > 1 && !strcmp(argv[1], "--sdp-basic"))
 		return sdp_basic_runner_cmd_parser(argc-1, &argv[1]);
+#endif
+#ifdef CFG_PKCS11_TA
+	else if (argc > 1 && !strcmp(argv[1], "--pkcs11_1028_destroy_token_object"))
```
jenswi-linaro, line 171: argc == 2?

jforissier:
> The test relies on running another PKCS11 client from another process.
> Using pthread_create() does not suit our needs as it is seen as the
> same client from the PKCS11 library.

fork()/exec()?

etienne-lms:
Do you mean I should rephrase to something like the below?
```
The test relies on running another PKCS11 client from another process
which is realized by executing shell command
"xtest --pkcs11_1028_destroy_token_object"
to created a specific process the PKCS11 library identifies as a different
client that finds the token object and destroys it.
```

jforissier:
@etienne-lms no I mean that the user should just need to run `xtest pkcs11_1028` to run this scenario, as is done for any other test. It doesn't matter if several processes are needed that is an implementation detail ;-)

etienne-lms:
User only needs to run `xtest pkcs11_1028`. The test executes `xtest --pkcs1_...` (and checks return value) on purpose.

host/xtest/pkcs11_1000.c
```diff
@@ -8736,3 +8737,127 @@ static void xtest_pkcs11_test_1027(ADBG_Case_t *c)
...
+	if (!ADBG_EXPECT_CK_OK(c, rv)) {
+		rv = C_DestroyObject(session, obj_handle);
+		ADBG_EXPECT_CK_OK(c, rv);
+		goto out_session;
+	}
+
+	ret = system("xtest --pkcs11_1028_destroy_token_object");
+	ADBG_EXPECT(c, ret, 0);
+
+	/* Session B has deleted the object: handle shll no more be valid */
```
etienne-lms, line 8848: s/shll/shall/

jforissier:
> User only needs to run `xtest pkcs11_1028`. The test executes `xtest --pkcs1_...` (and checks return value) on purpose.

Indeed, sorry I should have looked more closely at the patch before commenting. That part looks good.

host/xtest/xtest_main.c
jforissier: +1 for `pkcs11-1028-destroy-token-object`

etienne-lms:
> Indeed, sorry I should have looked more closely at the patch before commenting.

That's fine. I prefer a comment to be wrong rather than a valid comment to not be posted :)

etienne-lms:
Comments addressed + few changes in source file inline description comments.

etienne-lms:
I've added a fixup commit for a minor simplification in `xtest_pkcs11_1028_destroy_token_object()`.

host/xtest/pkcs11_1000.c
```diff
@@ -8744,6 +8735,126 @@ static void xtest_pkcs11_test_1027(ADBG_Case_t *c)
...
+		goto out_session;
+
+	rv = C_GetObjectSize(session, obj_handle, &obj_size);
+	if (!ADBG_EXPECT_CK_OK(c, rv)) {
+		rv = C_DestroyObject(session, obj_handle);
+		ADBG_EXPECT_CK_OK(c, rv);
+		goto out_session;
+	}
+
+	ret = system("xtest --pkcs11-1028-destroy-token-object");
```
jenswi-linaro, line 8842: I wonder if it wouldn't be better to use what's in `argv[0]` than assuming that `xtest` is OK.

etienne-lms: Nice catch. Maybe export `argv[0]` from _xtest_main.c_ with some global variable (`char *xtest_argv_0`)?

jenswi-linaro: Sure

etienne-lms:
review comment addressed

host/xtest/pkcs11_1000.c
```diff
@@ -8744,6 +8735,139 @@ static void xtest_pkcs11_test_1027(ADBG_Case_t *c)
...
+out_session:
+	C_CloseSession(session);
+out_lib:
+	close_lib();
+
+	return ret;
+}
+
+/*
+ * This test invloves 2 client sessions towards the PKCS11 token.
```
jforissier, line 8801: involves

host/xtest/xtest_main.c
```diff
@@ -107,6 +108,9 @@ void usage(char *program)
 #endif
 #ifdef CFG_SECURE_DATA_PATH
 	printf("\t--sdp-basic [opts] Basic Secure Data Path test setup ('-h' for usage)\n");
+#endif
+#ifdef CFG_PKCS11_TA
+	printf("\t--pkcs11-1028-destroy-token-object  Execute test for pkcs11_1028 case\n");
```
jforissier, line 113: "Used internally by pkcs11_208" maybe? (to make it a bit more clear that this option is not useful for the end user)

jforissier (commented):
For commit "xtest: pcsk11: test 1028 check handles of destroyed object": minor comments only.

jforissier (commented):
For commit "xtest: export xtest command string":

...but it must come before "xtest: pcsk11: test 1028 check handles of destroyed object"

host/xtest/stats.c
```diff
@@ -47,11 +47,11 @@ struct ta_dump_stats {
 	struct malloc_stats heap;
 };
 
-static const char *stats_progname = "xtest --stats";
+static const char *stats_option = "--stats";
 
 static int usage(void)
 {
-	fprintf(stderr, "Usage: %s [OPTION]\n", stats_progname);
+	fprintf(stderr, "Usage: %s %s [OPTION]\n", xtest_argv_0, stats_option);
```
jforissier, line 54: I'd rather have this function invoked as `usage(argv[0]);` by the various `stat_*()` functions because all other "applets" work like that.

jforissier (commented):
Reviewing "xtest: stats: use exported xtest command name".

host/xtest/xtest_main.c
etienne-lms: +1. agree, I'll add this info.

host/xtest/stats.c
etienne-lms: OK, i missed that. Thanks.

host/xtest/xtest_main.c
```diff
@@ -153,6 +157,9 @@ int main(int argc, char *argv[])
 
 	init_ossl();
 
+	if (argc >= 1)
```
etienne-lms, line 160: testing `argc >= 0` is useless I guess :)

host/xtest/stats.c
etienne-lms: see https://github.com/OP-TEE/optee_test/pull/672

host/xtest/xtest_main.c
```diff
@@ -108,6 +108,9 @@ void usage(char *program)
 #endif
 #ifdef CFG_SECURE_DATA_PATH
 	printf("\t--sdp-basic [opts] Basic Secure Data Path test setup ('-h' for usage)\n");
+#endif
+#ifdef CFG_PKCS11_TA
+	printf("\t--pkcs11-1028-destroy-token-object  Used internally by pkcs11_208\n");
```
jforissier, line 113: 1028 :wink:

etienne-lms:
typo fixed, thanks.

etienne-lms:
I appended a fixup commit to address https://github.com/OP-TEE/optee_test/pull/665#pullrequestreview-1461915893.

host/xtest/xtest_main.c
```diff
@@ -41,6 +41,7 @@ ADBG_SUITE_DEFINE(ffa_spmc);
 #endif
 ADBG_SUITE_DEFINE(regression);
 
+char *xtest_argv_0;
```
jenswi-linaro, line 44: I'd prefer `xtest_progname` since that's the expected use.

etienne-lms:
comment addressed.

etienne-lms:
@jforissier, could you have a quick look and confirm your review tags?

jforissier:
@etienne-lms review tags confirmed, thanks!

etienne-lms:
Fixup commits squashed. 
@jforissier review tags are applied.

etienne-lms:
thanks

PR #669, opened 2023-05-29
xtest: regression_1037: Close attestation session

cck-wcs:
Reviewed-by: liushiwei <liushiwei@eswincomputing.com>

### DIFF
--- a/host/xtest/regression_1000.c
+++ b/host/xtest/regression_1000.c
@@ -3124,6 +3124,8 @@ static void xtest_tee_test_1037(ADBG_Case_t *c)
 		return;
 	}
 
+	TEEC_CloseSession(&session);
+
 	Do_ADBG_BeginSubCase(c, "Get public key");
 	get_att_public_key(c);
 	Do_ADBG_EndSubCase(c, "Get public key");


host/xtest/regression_1000.c
```diff
@@ -3141,6 +3141,8 @@ static void xtest_tee_test_1037(ADBG_Case_t *c)
 	Do_ADBG_EndSubCase(c, "TA attestation (memory)");
 
 	free_att_key();
+
+	TEEC_CloseSession(&session);
```
jenswi-linaro, line 3145: Good catch. I wonder if it wouldn't make sense to put this just after the `if (...` above since it's not by anything past that point.

cck-wcs: > Good catch. I wonder if it wouldn't make sense to put this just after the `if (...` above since it's not by anything past that point.
Hi @jenswi-linaro, do you mean to put TEEC_CloseSession directly on line 3126?

jenswi-linaro: > do you mean to put TEEC_CloseSession directly on line 3126?

Yes

cck-wcs: > > do you mean to put TEEC_CloseSession directly on line 3126?
> 
> Yes

OK

jenswi-linaro (commented):

cck-wcs:
> Reviewed-by: Jens Wiklander <jens.wiklander@linaro.org>

OK

PR #672, opened 2023-06-05
xtest: fix xtest tool name in usage trace messages

etienne-lms:
Fixes the xtest program name printed in applets usage and error trace messages. Stats applet assumes it is named xtest instead of using caller argv[0]. Other applets simply print the applet option name instead of xtest tool name followed by applet option name.

### DIFF
--- a/host/xtest/aes_perf.c
+++ b/host/xtest/aes_perf.c
@@ -23,6 +23,7 @@
 
 #include "crypto_common.h"
 #include "xtest_helpers.h"
+#include "xtest_test.h"
 
 #ifdef CFG_SECURE_DATA_PATH
 #include "sdp_basic.h"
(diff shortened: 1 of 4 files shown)


host/xtest/xtest_main.c
```diff
@@ -156,21 +156,21 @@ int main(int argc, char *argv[])
 	init_ossl();
 
 	if (argc > 1 && !strcmp(argv[1], "--sha-perf"))
-		return hash_perf_runner_cmd_parser(argc-1, &argv[1]);
+		return hash_perf_runner_cmd_parser(argc, argv);
```
jenswi-linaro, line 159: How about as first thing save `argv[0]` in a global variable? It should save us from having to do the same thing again and again.

etienne-lms: ok

etienne-lms:
I've force pushed  the branch instead of appending a fixup commit that would have almost reverted the initially proposed commit.
This change includes a commit already under review in https://github.com/OP-TEE/optee_test/pull/665.

jforissier (commented):

host/xtest/xtest_main.c
```diff
@@ -41,6 +41,7 @@ ADBG_SUITE_DEFINE(ffa_spmc);
 #endif
 ADBG_SUITE_DEFINE(regression);
 
+char *xtest_argv_0;
```
jenswi-linaro, line 44: As I said in the other PR, I'd prefer `xtest_progname` as that's the expected use.

etienne-lms:
Comment addressed, as for https://github.com/OP-TEE/optee_test/pull/665.
Let's wait https://github.com/OP-TEE/optee_test/pull/665 is merged and i'll rebase this P-R.

etienne-lms:
I've squashed the commits and rebased as https://github.com/OP-TEE/optee_test/pull/665 has been merged.

jforissier:
My R-b stands.

etienne-lms:
tag applied. thanks.

jenswi-linaro:

PR #688, opened 2023-07-24
xtest: regression_1000: remove unneeded stat.h include

jonmason:
Hack to work around musl compile error:
 In file included from optee-test/3.17.0-r0/recipe-sysroot/usr/include/sys/stat.h:23,
                  from optee-test/3.17.0-r0/git/host/xtest/regression_1000.c:25:
 optee-test/3.17.0-r0/recipe-sysroot/usr/include/bits/stat.h:17:26: error: expected identifier or '(' before '[' token
    17 |         unsigned __unused[2];
       |                          ^

stat.h is not needed, since it is not being used in this file.  So removing it.

### DIFF
--- a/host/xtest/regression_1000.c
+++ b/host/xtest/regression_1000.c
@@ -24,7 +24,6 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
-#include <sys/stat.h>
 #include <sys/types.h>
 #include <ta_arm_bti.h>
 #include <ta_concurrent.h>


jforissier:
This indicates a more fundamental problem which is that `optee_os` [exports](https://github.com/OP-TEE/optee_os/blob/3.22.0/ta/ta.mk#L115-L127) some header files to `$(out-dir)/export-$(sm)/host_include` which really belong in the TA build environment only (that is, when the C library is libutils which is the C library provided by `optee_os`).
In this case the conflict is caused by `compiler.h` which defines an `__unused` macro.

I believe the exports need reviewing and fixing, but in the meantime this patch looks fine to me.

jenswi-linaro:

jforissier:
@jonmason could you please add the Reviewed-by and Acked-by tags and force push? Then I will merge this. Thanks.

PR #689, opened 2023-07-25
xtest: add -DOPENSSL_API_COMPAT=10100

jforissier:
regression_1000.c uses OpenSSL functions that have been deprecated after version 1.1. In particular, building with version 3.0 causes the following error (and other similar ones not pasted here):

 optee_test/host/xtest/regression_1000.c: In function 'check_signature':
 optee_test/host/xtest/regression_1000.c:2735:9: error: 'SHA256_Init' is deprecated: Since OpenSSL 3.0 [-Werror=deprecated-declarations]
  2735 |         SHA256_Init(&ctx);
       |         ^~~~~~~~~~~

Add -DOPENSSL_API_COMPAT=10100 to the build flags to avoid this and be flexible with regards to which version of OpenSSL build environment has to provide.


jenswi-linaro:

PR #697, opened 2023-10-06
ta/crypt: update to mbedTLS 3.4.0 API

jenswi-linaro:
The mbedtls_pk_parse_key() has two new paramters with the new 3.x API in. Fix the call of mbedtls_pk_parse_key(). The X.509 self test is removed so remove the call to that test in the TA.

Acked-by: Jerome Forissier <jerome.forissier@linaro.org>


jenswi-linaro:
Depends on https://github.com/OP-TEE/optee_os/pull/6351

jforissier (commented):

jenswi-linaro:
Tag applied.

PR #699, opened 2023-10-16
Update help of xtest --stats

JJayat:
The command 'xtest --stats --time' is not documented in the usage
 description

### DIFF
--- a/host/xtest/stats.c
+++ b/host/xtest/stats.c
@@ -58,6 +58,7 @@ static int usage(void)
 	fprintf(stderr, " --alloc        Print allocation statistics\n");
 	fprintf(stderr, " --memleak      Dump memory leak data on secure console\n");
 	fprintf(stderr, " --ta           Print loaded TAs context\n");
+	fprintf(stderr, " --time         Print REE and TEE time\n");
 
 	return EXIT_FAILURE;
 }


etienne-lms (commented):
`Signed-off-by:` tag is missing in the commit message.
I suggest to rephrase the message in describing what's done:
> The command 'xtest --stats --time' is **~~not~~ now** documented in the usage
> description.

`Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>` once addressed.

host/xtest/stats.c
```diff
@@ -58,6 +58,7 @@ static int usage(void)
 	fprintf(stderr, " --alloc        Print allocation statistics\n");
 	fprintf(stderr, " --memleak      Dump memory leak data on secure console\n");
 	fprintf(stderr, " --ta           Print loaded TAs context\n");
+	fprintf(stderr, " --time         Print system time\n");
```
jforissier, line 61: `Print REE and TEE time`

jforissier (commented):
With my comment addressed:

etienne-lms (commented):
could you format the commit message header line into `xtest: ...`? maybe:
`xtest: add help for --stats --time`.
Please also add a period  in the message:
```patch
 The command 'xtest --stats --time' is now documented in the usage
- description
+description.

PR #703, opened 2023-10-31
ta: os_test: fetch system time after the one second delay for proper comparaison

clementfaure:
…time comparaison

In the regression_1006 test, the wrapped TA time is compared to the system time. The time delay for this test is one second. The system time is fetched at the very begining of the test function.

For this test to pass, it is assumed that the system time fetched is at least one second. That might not be the case in a system where the test is started right after a deep sleep state where the system counter gets reset.

To solve this issue, fetch the system counter after the one second delay.

### DIFF
--- a/ta/os_test/os_test.c
+++ b/ta/os_test/os_test.c
@@ -635,10 +635,9 @@ static TEE_Result test_time(void)
 	printf("TA time %u.%03u\n", (unsigned int)t.seconds,
 	       (unsigned int)t.millis);
 
-	if (t.seconds > sys_t.seconds) {
-		EMSG("Unexpected wrapped time %u.%03u (sys_t %u.%03u)\n",
-		     (unsigned int)t.seconds, (unsigned int)t.millis,
-		     (unsigned int)sys_t.seconds, (unsigned int)sys_t.millis);
+	if (t.seconds > 1) {
+		EMSG("Unexpected wrapped time %u.%03u\n",
+		     (unsigned int)t.seconds, (unsigned int)t.millis);
 		return TEE_ERROR_BAD_STATE;
 	}
 


ta/os_test/os_test.c
```diff
@@ -635,6 +631,10 @@ static TEE_Result test_time(void)
 	printf("TA time %u.%03u\n", (unsigned int)t.seconds,
 	       (unsigned int)t.millis);
 
+	TEE_GetSystemTime(&sys_t);
+	printf("system time %u.%03u\n", (unsigned int)sys_t.seconds,
+	       (unsigned int)sys_t.millis);
+
 	if (t.seconds > sys_t.seconds) {
```
jenswi-linaro, line 638: This test seems a bit strange. What are we testing against here, a time that has wrapped too much?

clementfaure: I guess this test checks if the TA time wrapped correctly?
If the TA time wraps correctly after the one second delay, the TA time should be inferior to one second. That TA time will always be inferior to the system time (with that patch).
If the TA time does not wrap and stays to its max value, the TA time will be superior to system time. There is still a chance where TA time and system time are at the maximum value at the very same time ... but that's unlikely.

etienne-lms: I guess the test checks that TA time has indeed wrapped. Maybe testing against 0 should be sufficient.

etienne-lms:
> For this test to pass, it is assumed that the system time fetched is at least one second. That might not be the case in a system where the test is started right after a deep sleep state where the system counter gets reset.

It seems the issue is rather that System Time got reset during a low power state transition whereas it should not.
That said, I think testing TA wrapped time against 0 is enough for this test.

clementfaure:
> > For this test to pass, it is assumed that the system time fetched is at least one second. That might not be the case in a system where the test is started right after a deep sleep state where the system counter gets reset.
> 
> It seems the issue is rather that System Time got reset during a low power state transition whereas it should not. That said, I think testing TA wrapped time against 0 is enough for this test.

I agree that a system counter reset should not be expected, we are investigating. However, I still think the logic of this test is wrong. Let's imagine you system boots super-fast and run xtest, this test will fail.

etienne-lms:
I agree that there is no need to test TA wrapped time against TEE system time. We just need to test it wrapped. As we expect only few seconds have passed, testing t_seconds < 2 should be enough, or maybe 5.
Testing against 0 as I previously suggested is a non-sense :astonished:.

ta/os_test/os_test.c
```diff
@@ -635,10 +635,9 @@ static TEE_Result test_time(void)
 	printf("TA time %u.%03u\n", (unsigned int)t.seconds,
 	       (unsigned int)t.millis);
 
-	if (t.seconds > sys_t.seconds) {
-		EMSG("Unexpected wrapped time %u.%03u (sys_t %u.%03u)\n",
-		     (unsigned int)t.seconds, (unsigned int)t.millis,
-		     (unsigned int)sys_t.seconds, (unsigned int)sys_t.millis);
+	if (t.seconds > 1) {
```
etienne-lms, line 638: In case of high CPU load, I think the test can legitimately reach a bit more than 1sec. So using 2 here is safer IMO.

clementfaure: The test will pass with TA time up to 1s999ms.
You suggest doing `if (t.seconds > 2) { error } ` for a TA time up to 2s999ms ?

etienne-lms: You're right. Fine as is.

etienne-lms (commented):

clementfaure:
Tag applied

jenswi-linaro:

clementfaure:
Tags applied

PR #706, opened 2023-11-13
xtest: combine AES and hash into crypto_perf

yuzexiyzx:
There is quite a bit of copy in thes files.So, I cosonlidate th code to make it easier to maintain.


jforissier:
The asymmetric cipher app is buggy in many ways. Please test it and fix it so that it behaves as documented by `xtest --asym-cipher-perf --help`. I tried some random combinations that are supposedly valid as per the help string, but unfortunately I got errors:

```
$ xtest --asym-cipher-perf -n 10 -t DH
TEEC_InvokeCommand: 0xffff0006 (orig=4)
$ xtest --asym-cipher-perf -n 10 -t ECDSA
ECC curve is not supported!
$ xtest --asym-cipher-perf -n 10 -t ECDH
--asym-cipher-perf, invalid main_algo
Usage: --asym-cipher-perf [-h]
[...]
$ xtest --asym-cipher-perf -h 2>&1 | grep '???'
  -a ALGO          Crypto algorithm tested when TYPE is RSA [???]
$ xtest --asym-cipher-perf -m ENC
The algo[4294967295] is not valid!
$ xtest --asym-cipher-perf -a RSA_NOPAD -m ENC
GENKEYPAIR, invalid mode
```

yuzexiyzx:
OK.I will test it and fix it.

yuzexiyzx:
For $ xtest --asym-cipher-perf -a RSA_NOPAD -m ENC
-a should be after -t and -m

jforissier:
> For $ xtest --asym-cipher-perf -a RSA_NOPAD -m ENC -a should be after -t and -m

Not acceptable. There is no reason for a specific order here.

yuzexiyzx:
-a is specific for RSA algorithm and it is different for enc or dec and sign or verify.So,it is valid without -m or -t parameters,but it is invalid to modify -m or -t parameter after -a parameter.

yuzexiyzx:
There are many connections between parameters, For example,-a is specific for RSA algorithm, and ECDSA only accepts sign and verify… Is it acceptable to show it in usage?Or I should add some check of these parameters.

yuzexiyzx:
The valid width_bits of dh，rsa，ecc are also different.I cannot use a default value to make all scenes valid.

jforissier:
Maybe drop `-t` and `-m` and keep only `-a` with all the supported algorithms? That's a big list but it doesn't matter, it should be simpler than to check consistency of the parameters, and even easier to use too.

yuzexiyzx:
> Maybe drop `-t` and `-m` and keep only `-a` with all the supported algorithms? That's a big list but it doesn't matter, it should be simpler than to check consistency of the parameters, and even easier to use too.

May I seperate this pull requerest into two? It would take some time to solve these problems.Is it acceptalbe to merge the refactoring of aes and hash first? I will add SM4 and asymmetric cipher algothms later.

etienne-lms:
> May I seperate this pull requerest into two? It would take some time to solve these problems.Is it acceptalbe to merge the refactoring of aes and hash first? I will add SM4 and asymmetric cipher algothims later.

Sure. I agree it would help having the 1st commit be merged. Please create a new P-R for this commit. You can keep the applied review tags.

jenswi-linaro:

etienne-lms:
IBART fails. @yuzexiyzx, could you rebase on master tip. I hope it will fix issues found by  IBART.

jforissier:
Not rebased

yuzexiyzx:
Does rebasing means updating?
image
I have done this.

etienne-lms:
> Does rebasing means updating?

No, this magic button (Update branch button of Github WebUI) will not do the job needed here.

One way is to cherry pick your change on top of latest OP-TEE master branch, and then force push the result to your branch. See https://optee.readthedocs.io/en/latest/general/contribute.html#finalizing-your-contribution

yuzexiyzx:
@etienne-lms I have done that.Is the status right now?

etienne-lms:
All good, can be merged IMO.

jforissier:
I updated the commit subject and fixed a white space error reported by checkpatch. The commit is now merged into master as e18381f54eb86326f64c0cbe472e417decb8be39.

Thanks @yuzexiyzx!

PR #707, opened 2023-11-17
Add NIST cipher and hash tests vectors

omasse-linaro:
Following the commit https://github.com/OP-TEE/optee_test/commit/f6efe24adcf5b559b5ab90fa92d96701232b7c1e , proposition is to add more cipher and hash tests vectors to optee-test.

This PR remove the dependency on NIST package by adding all the needed NIST vectors definitions in new include files.
As it was already done for ECDSA in commit c7654961daf246fe04c923a7a3ff7a9a7d660443


omasse-linaro:
> Why? That's quite a big amount of code duplication... what are the benefits of having the vectors here as opposed to downloading the `.rsp` files and generating the test cases at build time?

you can find this in the following PR https://github.com/OP-TEE/optee_test/pull/698
'.rsp' file are downloaded and parsed to generate the test cases.

jenswi-linaro:
Apart from adding quite a lot of what would normally be unused, there's no license or copyright on the new files.

omasse-linaro:
> Apart from adding quite a lot of what would normally be unused, there's no license or copyright on the new files.

indeed.

jenswi-linaro:
What is your reasoning for the added license and copyright?
Before we dig deeper into that perhaps we should decide if we want to do it like this at all. I'm not in favor of this massive increase in size.

jforissier:
> > Why? That's quite a big amount of code duplication... what are the benefits of having the vectors here as opposed to downloading the `.rsp` files and generating the test cases at build time?
> 
> you can find this in the following PR #698 '.rsp' file are downloaded and parsed to generate the test cases.

Indeed, sorry I haven't taken time to review it yet.

Frankly I do not like the idea of generating the C files once for all. This patch introduces 300K lines of code compares with 1K for 968, for no obvious benefit.

PR #711, opened 2023-11-21
xtest: pkcs11: fix false positive error report on CKR_HOST_MEMORY

etienne-lms:
Fix false positive error case in test pkcs11_1004 and pkcs11_1007.

In the pkcs11_1004 test, the client creates as many PKCS#11 objects as possible until the system fails on memory allocation error. In the pkcs11_1007 test, the client prepares as many cipher sessions as possible until the system fails on memory allocation error. Both test implementations consider only CKR_DEVICE_MEMORY return code as the condition for expected resource exhaustion while they should also consider the host resource exhaustion return code CKR_HOST_MEMORY.

Fixes: 2d6dc9312935 ("xtest: pkcs11: add symmetric cipher tests")

### DIFF
--- a/host/xtest/pkcs11_1000.c
+++ b/host/xtest/pkcs11_1000.c
@@ -1235,7 +1235,8 @@ static void test_create_destroy_session_objects(ADBG_Case_t *c)
 				    ARRAY_SIZE(cktest_session_object),
 				    obj_hdl + n);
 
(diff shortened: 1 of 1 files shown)


jbech-linaro (commented):

etienne-lms:
review tag applied

jenswi-linaro:

PR #718, opened 2023-12-06
xtest: pkcs11: Add check that CK_TOKEN_INFO outputs correct CK_ULONG's

vesajaaskelainen:
OP-TEE's PKCS#11 TA TEE<->REE interface treats `CK_ULONG`'s as 32 bit.

In 64 bit machines this creates a problem for `CK_ULONG` values that are architecture dependent.

One of those defines is `CK_UNAVAILABLE_INFORMATION`.

This adds test case to make sure that when OP-TEE's PKCS#11 TA should give `CK_UNAVAILABLE_INFORMATION` it actually does that.

Depends on:
- https://github.com/OP-TEE/optee_client/pull/367

### DIFF
--- a/host/xtest/pkcs11_1000.c
+++ b/host/xtest/pkcs11_1000.c
@@ -309,6 +309,36 @@ static void xtest_pkcs11_test_1001(ADBG_Case_t *c)
 		if (!ADBG_EXPECT_CK_OK(c, rv))
 			goto out;
 
+		/**
+		 * OP-TEE's PKCS#11 implementaion always responds with
+		 * CK_UNAVAILABLE_INFORMATION for fields:
+		 *
+		 * - ulMaxSessionCount
+		 * - ulMaxRwSessionCount
+		 * - ulTotalPublicMemory
+		 * - ulFreePublicMemory
+		 *
+		 * Verify that CK_UNAVAILABLE_INFORMATION is correctly
+		 * translated (32 bit vs 64 bit difference).
+		 */
+
+		if (!ADBG_EXPECT_COMPARE_UNSIGNED(c,
+				token_info.ulMaxSessionCount, ==,
+				CK_UNAVAILABLE_INFORMATION))
+			goto out;
+		if (!ADBG_EXPECT_COMPARE_UNSIGNED(c,
+				token_info.ulMaxRwSessionCount, ==,
+				CK_UNAVAILABLE_INFORMATION))
+			goto out;
+		if (!ADBG_EXPECT_COMPARE_UNSIGNED(c,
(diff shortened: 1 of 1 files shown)


vesajaaskelainen:
Just rebased to master.

vesajaaskelainen:
Just rebased to master.

vesajaaskelainen:
@jforissier now that the client part is merged -- could you trigger test run again for this PR?

jforissier:
@vesajaaskelainen sure. Done!

host/xtest/pkcs11_1000.c
```diff
@@ -309,6 +309,36 @@ static void xtest_pkcs11_test_1001(ADBG_Case_t *c)
 		if (!ADBG_EXPECT_CK_OK(c, rv))
 			goto out;
 
+		/**
+		 * Verify that fields with CK_UNAVAILABLE_INFORMATION gets
+		 * correct value
```
etienne-lms, line 314: I think it would be better to highlight here that the pkcs11 always return CK_UNAVAILABLE_INFORMATION for `ulMaxSessionCount`, `ulMaxRwSessionCount`, `ulTotalPublicMemory` and `ulFreePublicMemory` so we can test them safely.

As per `ulTotalPrivateMemory`  and `ulFreePrivateMemory` I think we can implement them one day, so I would not enforce there values in this test.

etienne-lms (commented):
I would suggest to mention pkcs11_1001 in the commit header line, as:
`xtest: pkcs11_1001: Test CK_UNAVAILABLE_INFORMATION output value`

host/xtest/pkcs11_1000.c
```diff
@@ -309,6 +309,36 @@ static void xtest_pkcs11_test_1001(ADBG_Case_t *c)
...
+		 */
+
+		if (!ADBG_EXPECT_COMPARE_UNSIGNED(c,
+			token_info.ulMaxSessionCount, ==,
+			CK_UNAVAILABLE_INFORMATION))
+			goto out;
```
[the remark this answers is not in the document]
vesajaaskelainen: Ok

host/xtest/pkcs11_1000.c
vesajaaskelainen: Ok

vesajaaskelainen:
Sorry for a bit of delay on updates for the PR.

But now all review comments are addressed and PR has been re-based and re-tested.

etienne-lms (commented):

vesajaaskelainen:
> `Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>`

Thanks for the review!

Applied tag.

PR #724, opened 2024-01-29
regression_4000: do not check DH private value key size

clementfaure:
Depending on the crypto engine used, the generated DH private value might be different from the expected key size.
This is true for the CAAM black key implementation where the private key is encapsulated in a blob format that contains additionnal data, resulting in a bigger buffer.

### DIFF
--- a/host/xtest/regression_4000.c
+++ b/host/xtest/regression_4000.c
@@ -4368,6 +4368,21 @@ struct key_attrs {
 	uint32_t keysize_check;
 };
 
+static bool is_caam_black_key(uint8_t *buf, size_t size)
+{
+	/*
+	 * This value is a magic number for the a CAAM Black key. This value
+	 * must match the value defined in optee-os
+	 * core/drivers/crypto/caam/caam_key.c
+	 */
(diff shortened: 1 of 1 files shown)


host/xtest/regression_4000.c
```diff
@@ -4457,7 +4457,7 @@ static bool test_dh_key_pair(ADBG_Case_t *c, TEEC_Session *s,
 		KEY_ATTR(TEE_ATTR_DH_PRIME, false),
 		KEY_ATTR(TEE_ATTR_DH_BASE, false),
 		KEY_ATTR(TEE_ATTR_DH_PUBLIC_VALUE, false),
-		KEY_ATTR(TEE_ATTR_DH_PRIVATE_VALUE, check_keysize),
+		KEY_ATTR(TEE_ATTR_DH_PRIVATE_VALUE, false),
```
jenswi-linaro, line 4460: Would it be possible to do something clever in `test_keygen_attributes()` to see if it's a black key and then allow for a larger size? Anything is better than nothing.

jenswi-linaro:
Nice with the reference to understand what's going on.

clementfaure:
Force-push: tag applied + rebase

PR #729, opened 2024-02-12
ci: avoid "No space left on device" error

jforissier:
Remove unneeded files from the root filesystem to avoid a "No space left on device" error. A similar trick was committed to optee_os for the same reason [1].

Link: https://github.com/OP-TEE/optee_os/commit/fa1950059f4128a9b58558d4970c153595463ff3 [1]


etienne-lms (commented):

Issue #731, opened 2024-02-13
xtest 1009 - Explanation about cancellationrequest


bvalliere:
Hello

According to the section 4.10.2 (TEE_UnmaskCancellation) of Global Platform specification:
`By default, tasks created to handle a TA entry point have cancellation masked, so that a TA does not have to
cope with the effects of cancellation requests.`
So my understanding of this paragraph is that the cancellation mask is related to each entry point.

Nevertheless, when we see your implementation and the test 1009 related to CancellationRequest:
The `TEE_UnmaskCancellation` function is just call in OpenSessionEntryPoint.
```
TEE_Result TA_OpenSessionEntryPoint(uint32_t nParamTypes,
				    TEE_Param pParams[4],
				    void **ppSessionContext)
{
	(void)nParamTypes;
	(void)pParams;
	(void)ppSessionContext;
	DMSG("TA_OpenSessionEntryPoint");
	TEE_UnmaskCancellation();
```
So I think that your interpretation is that there is only one cancellation mask for all entry points.

So if my understanding related to the specification is correct and if I understood correctly your interpretation then I think that there is a problem with your test.

Can you confirm my hypothesis please ?

Regards

jenswi-linaro:
You have a valid point. OP-TEE keeps the cancellation mask as part of the session state and is maintained as long as the session is open, but from the specification, the cancellation mask should be reset on each entry. Feel free to propose pull requests to fix this.

c-auger:
Hi, in fact we are working for a company producing its own secure operating system, and we have implemented from scratch our own GP compliant TEE implementation. We are reusing your test suite and your frontend to check the compliance of our own implementation, but we are not using your backend.

As a result, when your test-suite fails on our implementation, we try to understand the discrepancy and fix our own implementation. Thus we are not knowledgeable of your exact implementation. I took a quick glance on it, and inferred that this might be fixed by editing https://github.com/OP-TEE/optee_os/blob/master/core/kernel/tee_ta_manager.c`, around line 764 to add a `sess->cancel_mask=true` (only the `invoke_command` part seems necessary, the `open_session` already calls `tee_ta_init_session()` which seems to already do the job).
But I am not aware of the implications of such a modifications and whether or not this would be correct. Furthermore, as we develop our own implementation, we have no incentive to stress-test this kind of modification, nor to enter a process of pull requests if the modifications are more complex than that.

PR #732, opened 2024-02-15
cancellation: Unmask inside of invoke command

c-auger:
The specification seems to indicate (see https://github.com/OP-TEE/optee_test/issues/731)
that the cancellation mask is reset at the start of every entry point.
As a result, the TEE_UnmakCancellation() should not be called at the end of
TA_OpenSessionEntryPoint(), but at the start of TA_InvokeCommandEntryPoint().

Also, note that this patch should not break the current test.
Still, there is a test which could be written to test that unmasking cancellations
really does not affect subsequent entry points. There is no such test for now,
and if one were designed, then the current implementation would report a failure to this test,
until the implementation is fixed with respect to the Global Platform specification.
If you need such a test, I could consider writing one.

### DIFF
--- a/ta/os_test/ta_entry.c
+++ b/ta/os_test/ta_entry.c
@@ -36,7 +36,6 @@ TEE_Result TA_OpenSessionEntryPoint(uint32_t nParamTypes,
 	(void)pParams;
 	(void)ppSessionContext;
 	DMSG("TA_OpenSessionEntryPoint");
-	TEE_UnmaskCancellation();
 	return TEE_SUCCESS;
 }
 
@@ -63,6 +62,7 @@ TEE_Result TA_InvokeCommandEntryPoint(void *pSessionContext,
 				      TEE_Param pParams[4])
 {
 	(void)pSessionContext;
+	TEE_UnmaskCancellation();
 
 	switch (nCommandID) {
 	case TA_OS_TEST_CMD_INIT:


jforissier (commented):
> cancellation: Unmask inside of invoke command
>
> The specification seems to indicate (see https://github.com/OP-TEE/optee_test/issues/731)
> that the cancellation mask is reset at the start of every entry point.
> As a result, the TEE_UnmakCancellation() should not be called at the end of
> TA_CreateEntryPoint(), but at the start of TA_OpenSessionEntryPoint().

s/TA_CreateEntryPoint()/TA_OpenSessionEntryPoint()/
s/TA_OpenSessionEntryPoint()/TA_InvokeCommandEntryPoint()/

With that fixed:

jenswi-linaro:
> Also, note that this patch should not break the current test. Still, there is a test which could be written to test that unmasking cancellations really does not affect subsequent entry points. There is no such test for now, and if one were designed, then the current implementation would report a failure to this test, until the implementation is fixed with respect to the Global Platform specification. If you need such a test, I could consider writing one.

Sure, if you don't mind, please provide such a test in a separate PR. I can take care of the secure side of things.

jforissier:
Hi @c-auger, thanks for the patch!

> Also, note that this patch should not break the current test. Still, there is a test which could be written to test that unmasking cancellations really does not affect subsequent entry points. There is no such test for now, and if one were designed, then the current implementation would report a failure to this test, until the implementation is fixed with respect to the Global Platform specification. If you need such a test, I could consider writing one.

That would be nice, please feel free to open a new PR. Then we can consider how to make the implementation compliant.

jforissier:
@c-auger we need a `Signed-off-by: Your Name <your email>` tag, please.

etienne-lms (commented):
Commit header line should rather be "ta: os_test: Unmask cancellation from invoke command handler".

jforissier:
@c-auger thanks!

PR #733, opened 2024-02-19
xtest: add SM4 perf test

yuzexiyzx:
Add perf test for SM4 algorithm


host/xtest/aes_perf.c
```diff
@@ -203,7 +211,7 @@ static void usage(const char *applet_optname, int keysize, int mode,
 	fprintf(stderr, "  -i|--in-place Use same buffer for input and output (decrypt in place)\n");
 	fprintf(stderr, "  -k SIZE       Key size in bits: 128, 192 or 256 [%u]\n", keysize);
 	fprintf(stderr, "  -l LOOP       Inner loop iterations [%u]\n", l);
-	fprintf(stderr, "  -m MODE       AES mode: ECB, CBC, CTR, XTS, GCM [%s]\n", mode_str(mode));
+	fprintf(stderr, "  -m MODE       mode: AES-ECB, AES-CBC, AES-CTR, AES-XTS, AES-GCM, SM4-ECB, SM4-CBC, SM4-CTR, SM4-XTS [%s]\n", mode_str(mode));
```
jforissier, line 214: No. The modes should not be changed. ECB is a mode, AES-ECB is not: it is a combination of a cipher and a mode. Please add a `-c|--cipher` option with values `AES` and `SM4` (default `AES`).

(Another option would be to keep your suggestion but change "-m MODE" to "-a ALGO", but I prefer two separate options because no need to expand the list of all combinations).

Also please do not forget to change:
```C
	fprintf(stderr, "AES performance testing tool for OP-TEE\n");
```
-> "AES/SM4 performance testing..."

PR #737, opened 2024-04-09
openssl: add option to provide the openssl library path

davidson-k:
Add a new variable to have the directory that contains the openssl libraries needed for xtest. The build system will look for the needed libraries specified in this variable during the linking of xtest.


host/xtest/Makefile
```diff
@@ -40,7 +40,7 @@ endef
 WITH_OPENSSL ?= y
 ifeq ($(WITH_OPENSSL),y)
 CFLAGS += -DOPENSSL_FOUND=1 -DOPENSSL_API_COMPAT=10100
-LDFLAGS += -lcrypto
+LDFLAGS += -L$(OPENSSL_DIR)/lib -lcrypto
```
jenswi-linaro, line 43: Is this needed for the CMakefile too?

davidson-k: I haven't tried compiling with cmake to really understand how that works and if that's needed or not.  Do you want me to compile with cmake as well ?

jenswi-linaro: No, that's not needed, the CI loop should cover the building.
We try to keep the makefiles in sync featurewise.

davidson-k: This patch is adding one more path to be searched while linking libcrypto.so provided by openssl.  This patch is not needed in both traditional make and cmake if we use the build system recommended by optee.  It is needed only when we want to use a openssl library that is compiled outside of the optee recommended build system.  

Now, coming to your question.  To achieve the above mentioned functionality we need to make changes in CMakeLists.txt as well.  But the current CMakeLists.txt is written very specific to be working with optee build system.  For eg, using traditional Makefile, we use a build flag to define if we want to link openssl libraries or not.  But in CMakeLists.txt, it searches for the openssl package and it links the libraries only if it finds the package.  I will need to re-write the logic here if I want to add the same functionality to CMakeLists.txt.

I will not try to do the same for CMakeLists.txt as well as I don't have enough proficiency to write/refactor the cmake files in a proper way.

host/xtest/Makefile
```diff
@@ -107,6 +107,7 @@ CFLAGS += -I../supp_plugin/include
 CFLAGS += -I$(out-dir)/xtest
 
 CFLAGS += -I$(OPTEE_CLIENT_EXPORT)/include
+CFLAGS += -I$(OPENSSL_DIR)/include
```
jenswi-linaro, line 110: What if `$(OPENSSL_DIR)` is an empty string? The compiler might warn about a non-existing include directory.

davidson-k: yes, it is throwing errors, :(

davidson-k: I have removed including the directory.  I can pass the directory through `CPATH` variable while calling `make`

jenswi-linaro: I assume you mean `CFLAGS`, what's stopping you from doing the same thing with `LDFLAGS`?

host/xtest/Makefile
etienne-lms, line 43: Is `OPENSSL_DIR` a standard make or environment variable? I seems to me it's awaited only for Rust based builds. Won't this change break existing build environments that use this Makefile but (maybe) do not define `OPENSSL_DIR`?
Maybe use:
```patch
 ifeq ($(WITH_OPENSSL),y)
 CFLAGS += -DOPENSSL_FOUND=1 -DOPENSSL_API_COMPAT=10100
 LDFLAGS += -lcrypto
+ifneq (,$(OPENSSL_DIR))
+LDFLAGS += -L$(OPENSSL_DIR)/lib
+endif

davidson-k: I have now changed the variable to `OPENSSL_LIBRARY_DIR` to avoid any confusion with the standard variables.

host/xtest/Makefile
davidson-k: To be precise, I cannot use `CFLAGS` and `LDFLAGS` while calling make as that will override those variables used in the xtest makefile.  If I want to append the LDFLAGS in the xtest Makefile with the LDFLAGS passed during make, I have to `override` the LDFLAGS in xtest Makefile.

For info: https://www.gnu.org/software/make/manual/html_node/Override-Directive.html

I have made those changes as well for reference. https://github.com/davidson-k/optee_test/commit/56bb3d3bbc5773ff5d3a361cb9f909684900774b
 
I'm ok with either of these two solution.

davidson-k: gentle ping @jenswi-linaro  :)

etienne-lms:

davidson-k:
can this be merged ?

jforissier:
> can this be merged ?

Absolutely. Sorry for the delay.

davidson-k:
Thank you !!

PR #738, opened 2024-04-10
xtest: regression 4006: Test RSA-OAEP with a different MGF1 hash

samitolvanen:
Add tests for TEE_ALG_RSAES_PKCS1_OAEP_MGF1_SHA256 with SHA-1 as the MGF1 hash.

Depends on https://github.com/OP-TEE/optee_os/pull/6779. Note that this fails (as expected) with mbedtls.


jforissier:
> Note that this fails (as expected) with mbedtls.

We need to have this as a separate test case (for example, `regression_4006_1`) so that it can be excluded from the run when MbedTLS is used as the core crypto lib:
```
make -j$(nproc) check CFG_CRYPTOLIB_NAME=mbedtls CFG_CRYPTOLIB_DIR=lib/libmbedtls XTEST_ARGS="-t regression -x regression_4006_1"
```

jenswi-linaro:
> > Note that this fails (as expected) with mbedtls.
> 
> We need to have this as a separate test case (for example, `regression_4006_1`) so that it can be excluded from the run when MbedTLS is used as the core crypto lib:
> 
> ```
> make -j$(nproc) check CFG_CRYPTOLIB_NAME=mbedtls CFG_CRYPTOLIB_DIR=lib/libmbedtls XTEST_ARGS="-t regression -x regression_4006_1"
> ```

Another option is to accept the `TEE_ERROR_NOT_SUPPORTED` error as a valid response (and skip the subtest) when `TEE_ATTR_RSA_OAEP_MGF_HASH` doesn't match the internal hash algorithm.

jforissier:
@jenswi-linaro that would work with me

samitolvanen:
> Another option is to accept the `TEE_ERROR_NOT_SUPPORTED` error as a valid response (and skip the subtest) when `TEE_ATTR_RSA_OAEP_MGF_HASH` doesn't match the internal hash algorithm.

Alright, I decided to go with this option. PTAL.

jenswi-linaro:

host/xtest/regression_4000.c
```diff
@@ -4150,12 +4164,20 @@ static void xtest_tee_test_4006(ADBG_Case_t *c)
...
+			res = ta_crypt_cmd_asymmetric_encrypt(c, &session, op,
 					algo_params, num_algo_params, tv->ptx,
-					tv->ptx_len, out_enc, &out_enc_size)))
+					tv->ptx_len, out_enc, &out_enc_size);
+
+			if (tv->algo == TEE_ALG_RSAES_PKCS1_OAEP_MGF1_SHA256 &&
+			    res == TEEC_ERROR_NOT_SUPPORTED) {
+				Do_ADBG_Log("RSA-OAEP with a different MGF1 "
+					    "hash not supported: skip "
+					    "subcase");
```
etienne-lms, line 4176: Is it worth to split this trace message?
Ditto at line 4258.

host/xtest/regression_4000.c
```diff
@@ -4214,12 +4236,32 @@ static void xtest_tee_test_4006(ADBG_Case_t *c)
...
+			res = ta_crypt_cmd_asymmetric_decrypt(c, &session, op,
+					algo_params, num_algo_params, tv->ctx,
+					tv->ctx_len, out, &out_size);
+
+			if (tv->algo == TEE_ALG_RSAES_PKCS1_OAEP_MGF1_SHA256 &&
+			    res == TEEC_ERROR_NOT_SUPPORTED) {
+				Do_ADBG_Log("RSA-OAEP with a different MGF1 "
+					    "hash not supported: skip "
+					    "subcase");
+				goto skip;
```
etienne-lms, line 4261: `break` is enough IMHO.

etienne-lms (commented):
`Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>` with comments addressed or not.

host/xtest/regression_4000.c
samitolvanen: Probably not, I'm fine with not splitting this.

host/xtest/regression_4000.c
samitolvanen: Ack. I'll push an update with these addressed.

PR #740, opened 2024-04-17
TA makefile updates

jforissier:


ta/Makefile.gmake
```diff
@@ -48,7 +48,19 @@ all: ta
 
 .PHONY: ta
 ta:
-	$(q)$(foreach dir,$(TA_DIRS), $(MAKE) -C $(dir) O=$(out-dir)/ta/$(dir) &&) true
+
+define add-ta-target
+.PHONY: ta-$(1)
+ta-$(1):
+	$$(MAKE) -C $(1)
```
jenswi-linaro, line 55: `$(MAKE)` should be enough, since we don't expect that to change.
Isn't `out-dir` missing?

jforissier:
Update

jenswi-linaro (commented):

PR #741, opened 2024-04-23
xtest regression case 4017

jenswi-linaro:
Depends on https://github.com/OP-TEE/optee_os/pull/6802

### DIFF
--- a/host/xtest/regression_4000.c
+++ b/host/xtest/regression_4000.c
@@ -6230,3 +6230,478 @@ out:
 }
 ADBG_CASE_DEFINE(regression, 4016, xtest_tee_test_4016_ed25519,
 		 "Test TEE Internal API ED25519 sign/verify");
+
+struct buf_state {
+	bool buffer_two_blocks;
+	bool auth_enc;
+	bool decrypt;
+	uint8_t *plain_text;
+	uint8_t *ciph_text;
+	uint8_t *tmp_text;
+	size_t text_size;
+	size_t block_size;
+	uint8_t *iv;
+	size_t iv_len;
+	uint8_t tag[96 / 8];
+	uint8_t tmp_tag[96 / 8];
+	TEE_OperationHandle oph;
+	TEEC_Session *s;
+
+	size_t in_count;
+	size_t out_count;
+	size_t buf_count;
+};
+
+static bool process_bytes_4017(ADBG_Case_t *c, struct buf_state *bs,
+			       size_t count)
+{
+	TEE_Result res = TEEC_SUCCESS;
+	size_t new_buf_count = 0;
+	void *reference = NULL;
+	void *src = NULL;
+	size_t out = 0;
+	size_t dlen = 0;
+
+	if (bs->decrypt) {
+		src = bs->ciph_text + bs->in_count;
+		reference = bs->plain_text + bs->out_count;
+	} else {
+		src = bs->plain_text + bs->in_count;
+		reference = bs->ciph_text + bs->out_count;
+	}
+
+	bs->in_count += count;
+	if (bs->buffer_two_blocks) {
+		if (bs->in_count > bs->block_size)
+			new_buf_count = ((bs->in_count - 1) % bs->block_size) +
+					bs->block_size + 1;
+		else
+			new_buf_count = bs->in_count;
+	} else {
+		new_buf_count = bs->in_count % bs->block_size;
+	}
+
+	out = bs->buf_count + count - new_buf_count;
+	bs->out_count += out;
+	bs->buf_count = new_buf_count;
+
+	if (bs->auth_enc) {
+		res = ta_crypt_cmd_ae_update(c, bs->s, bs->oph, src, count,
+					     NULL, &dlen);
+		if (!res)
+			return ADBG_EXPECT_COMPARE_UNSIGNED(c, dlen, ==, 0) &&
+			       ADBG_EXPECT_COMPARE_UNSIGNED(c, dlen, ==, out);
+		if (!ADBG_EXPECT_TEEC_RESULT(c, TEEC_ERROR_SHORT_BUFFER, res) ||
+		    !ADBG_EXPECT_COMPARE_UNSIGNED(c, dlen, ==, out))
+			return false;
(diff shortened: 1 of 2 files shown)


etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>` for commits
"ta: crypt: fix authenc entries" and
"xtest: return out-buffer sizes unconditionally".

host/xtest/regression_4000.c
```diff
@@ -6192,3 +6192,482 @@ static void xtest_tee_test_4016_ed25519(ADBG_Case_t *c)
...
+	out = bs->buf_count + count - new_buf_count;
+	bs->out_count += out;
+	bs->buf_count = new_buf_count;
+
+	if (bs->auth_enc) {
+		res = ta_crypt_cmd_ae_update(c, bs->s, bs->oph, src, count,
+					     NULL, &dlen);
+		if (!res)
+			return ADBG_EXPECT_COMPARE_UNSIGNED(c, dlen, ==, 0) &&
+			       ADBG_EXPECT_COMPARE_UNSIGNED(c, dlen, ==, out);
```
etienne-lms, line 6255: Do you mean
```c
return ADBG_EXPECT_COMPARE_UNSIGNED(c, dlen, ==, out);
```
?

etienne-lms (commented):
For commit "xtest: regression: add case 4017":
1 comment otherwise test 4017 implementation LGTM, pretty well factorized. I'll have another look.

host/xtest/regression_4000.c
jenswi-linaro: That works, but with a NULL output buffer it should be `0` also.

jforissier (commented):
For "xtest: regression: add case 4017", with my comments addressed:

jenswi-linaro:
@jforissier's comment addressed and tags applied.

host/xtest/regression_4000.c
```diff
@@ -6192,3 +6192,482 @@ static void xtest_tee_test_4016_ed25519(ADBG_Case_t *c)
...
+			goto out;
+		res = ta_crypt_cmd_cipher_do_final(c, bs.s, bs.oph,
+						   bs.plain_text, bs.text_size,
+						   bs.ciph_text, &dlen);
+		if (!ADBG_EXPECT_TEEC_SUCCESS(c, res) ||
+		    !ADBG_EXPECT_COMPARE_UNSIGNED(c, bs.text_size, ==, dlen))
+			goto out;
+	}
+
+	for (n = 0; n < bs.text_size; n++) {
```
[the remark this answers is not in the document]
jenswi-linaro: OK, done.

jforissier:
@etienne-lms any comment on "xtest: regression: add case 4017"?

host/xtest/regression_4000.c
```diff
@@ -6192,3 +6192,498 @@ static void xtest_tee_test_4016_ed25519(ADBG_Case_t *c)
...
+		return false;
+
+	if (initial_count &&
+	    !ADBG_EXPECT_TRUE(c, process_bytes_4017(c, bs, initial_count)))
+		return false;
+
+	for (n = 0; n < middle_count && n + initial_count < bs->text_size;
+	     n += middle_increment)
+		if (!ADBG_EXPECT_TRUE(c, process_bytes_4017(c, bs, 1)))
+			return false;
```
etienne-lms, line 6397: this looks strange to me. I think it works because `middle_increment` is always 1, as stated in the inline comment at line 6561:
```
		 * If level < 12 test with middle_count of 32 else 0.
		 * (middle_count - n) bytes are processed one by one.
```

Unless you plan some day to change this `middle_increment` in the future, I think it would more better to remove `middle_increment` and use `1` straight. Otherwise, I suggest to change line 6396:
```patch
-		if (!ADBG_EXPECT_TRUE(c, process_bytes_4017(c, bs, 1)))
+		if (!ADBG_EXPECT_TRUE(c, process_bytes_4017(c, bs, middle_increment)))
```
and line 6562:
```patch
 		 * If level < 12 test with middle_count of 32 else 0.
-		 * (middle_count - n) bytes are processed one by one.
+		 * (middle_count - n) bytes are processed by middle_increment
+		 * bytes bursts.
```

If I understood that correctly of course :)

host/xtest/regression_4000.c
```diff
@@ -6192,3 +6192,498 @@ static void xtest_tee_test_4016_ed25519(ADBG_Case_t *c)
...
+	if (!ADBG_EXPECT_TRUE(c, do_algo_4017(c, &sess, TEE_ALG_AES_CCM, 0)))
+		goto out;
+	Do_ADBG_EndSubCase(c, "TEE_ALG_AES_CCM");
+
+	Do_ADBG_BeginSubCase(c, "TEE_ALG_AES_CCM 1 extra byte");
+	if (!ADBG_EXPECT_TRUE(c, do_algo_4017(c, &sess, TEE_ALG_AES_CCM, 1)))
+		goto out;
+	Do_ADBG_EndSubCase(c, "TEE_ALG_AES_CCM 1 extra byte");
+
+out:
```
etienne-lms, line 6684: misses an call to `Do_ADBG_EndSubCase()` when we reach this point before line 6682 is reached.

Actually, I wonder if the `goto out` are really useful. If removed, we'll get a status of all the cases, maybe flowed of error trace messages if many test cases fail but i think its worth it, unless the session crashed soemwhere during the test cases.

host/xtest/regression_4000.c
jenswi-linaro: I'll drop the `middle_increment`.

host/xtest/regression_4000.c
jenswi-linaro: I'll remove the gotos.

jenswi-linaro:
Updated

etienne-lms (commented):
for commit "xtest: regression: add case 4017"

jenswi-linaro:
Tag applied

jenswi-linaro:
Rebased to fix CI issue.

PR #749, opened 2024-06-26
license: Add text of GPL-2.0 and BSD-2 licenses to repository

jamesr:
Resolves #748. The BSD license text's year and copyright holder are copied from https://github.com/OP-TEE/optee_os/blob/master/LICENSE

PR #752, opened 2024-07-24
license: Add copyright notice line to BSD license text

jamesr:
The copyright notice is part of the license text (see https://opensource.org/license/bsd-2-clause). I mistakenly removed this when applying review feedback for #749 - sorry about that. This adds the line back.

### DIFF
--- a/LICENSE-BSD
+++ b/LICENSE-BSD
@@ -1,3 +1,5 @@
+Copyright (c) 2015, Linaro Limited
+
 Redistribution and use in source and binary forms, with or without
 modification, are permitted provided that the following conditions are met:
 


LICENSE-BSD
```diff
@@ -1,3 +1,5 @@
+Copyright (c) 2015, Linaro Limited
```
jforissier, line 1: This mention doesn't properly reflect all the copyright owners, so I wonder what can be derived from this notice. I suppose it should be considered the default when a file doesn't have one.

jforissier (commented):

LICENSE-BSD
jamesr: That's my understanding - the particular files and/or git history can be more definitive for a particular file or set of files.

PR #754, opened 2024-08-29
xtest: copy dma-heap.h from Linux kernel v6.9

jenswi-linaro:
Copy the needed dma-heap.h from Linux kernel v6.9. This is a stable ABI so it doesn't have to be synced with each new kernel.

dma-heap.h can be removed when the cross toolchains provide the file.


jforissier (commented):
I suppose `optee_test/Android.mk` should be updated too (`LOCAL_C_INCLUDES`).

jenswi-linaro:
Comment addressed and tags applied.

PR #755, opened 2024-09-05
FIX: benchmark: fix wrong case title

wangfuan:
the benchmark 2002 case have wrote wrong SHA algo name, now correct it.


jforissier (approved):

Thanks!

PR #756, opened 2024-09-06
xtest: pkcs11: fix test_find_objects() when wrong count

etienne-lms:
Fix test_find_objects() so that C_FindObjectsFinal() is properly called even if we don't find the expected number of objects.

Fixes: afe2b54fb68b ("xtest: pkcs11: add find object tests")


jforissier (commented):

etienne-lms:
Tag applied

PR #757, opened 2024-09-06
xtest: add pkcs11_1031 for CKM_RSA_X_509 sign/verify

etienne-lms:
Basic test on signing/verifying with raw RSA mechanism CKM_RSA_X_509.

Depends on https://github.com/OP-TEE/optee_os/pull/7030 and https://github.com/OP-TEE/optee_client/pull/386.

### DIFF
--- a/host/xtest/pkcs11_1000.c
+++ b/host/xtest/pkcs11_1000.c
@@ -10017,3 +10017,390 @@ out_close_lib:
 }
 ADBG_CASE_DEFINE(pkcs11, 1030, xtest_pkcs11_test_1030,
 		 "PKCS11: Test AES-GCM Encryption/Decryption");
+
+static int test_rsa_raw_operations(ADBG_Case_t *c,
+				    CK_SESSION_HANDLE session,
+				    uint32_t rsa_bits)
+{
+	CK_RV rv = CKR_GENERAL_ERROR;
+	CK_OBJECT_HANDLE public_key = CK_INVALID_HANDLE;
+	CK_OBJECT_HANDLE private_key = CK_INVALID_HANDLE;
+	CK_MECHANISM mechanism = {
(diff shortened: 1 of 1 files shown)


host/xtest/pkcs11_1000.c
```diff
@@ -10017,3 +10017,390 @@ static void xtest_pkcs11_test_1030(ADBG_Case_t *c)
...
+	 *
+	 * If first byte is zero and 2nd byte is low, ensure the 2 next byte
+	 * are big (0xff). This will ensure the generated signature has the
+	 * same size as the key modulus.
+	 */
+	in_data[0] &= 0x7f;
+	if (!in_data[0] & !(in_data[1] & 0xf0)) {
+		in_data[2] = 0xff;
+		in_data[3] = 0xff;
+	}
```
jforissier, line 10159: This black magic is somewhat confusing. Why not use a kown good hardcoded `in_data`?

jforissier (commented):
One comment below.

`Tested-by: Jerome Forissier <jerome.forissier@linaro.org> (vexpress-qemu_armv8a)`

host/xtest/pkcs11_1000.c
etienne-lms: I have not found good hardcoded input data for raw RSA cases. The algorithm leave it up to the caller to "pad" input data but some "padding" schemes are not reliable. I tried to test flexible padding scheme yet ensuring the the first byte do not lead to an invalid signature scheme where the size of the signature is not of the size of the RSA key. I agree my implementation here is quite empiric.

jforissier: > I have not found good hardcoded input data for raw RSA cases.

My idea is to run this code once and print `in_data`. That's your known good input ;)

PR #761, opened 2024-11-14
gp: change from xalan to xsltproc

jbech-linaro:
We have been using an outdated version of xalan, which is what is installed using apt in Ubuntu. If you run Fedora, you will receive the latest version. The executable name has changed from "xalan" to "Xalan" and the supported command line arguments are also different. Therefore supporting different versions of xalan across distros can be challenging.

One alternative is xsltproc. It looks like it works on both Ubuntu and Fedora, so let's swap out xalan for xsltproc.

Tested-by: Joakim Bech <joakim.bech@linaro.org> (Fedora 41)
Tested-by: Joakim Bech <joakim.bech@linaro.org> (Ubuntu 22.04)

### DIFF
--- a/host/xtest/CMakeLists.txt
+++ b/host/xtest/CMakeLists.txt
@@ -74,8 +74,8 @@ if (WITH_GP_TESTS)
 	macro(__GEN_GP_FILE outfile xmldir name basedir)
 		add_custom_command(
 			OUTPUT	${outfile}
-			COMMAND	xalan -in ${basedir}/${name}.xml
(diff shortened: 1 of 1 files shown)


jforissier:
`Tested-by: Jerome Forissier <jerome.forissier@linaro.org> (Ubuntu 20.04)`

jbech-linaro:
Thanks for quick turnaround and testing. Patch squashed, rebased and tag(s) applied (should be) ready for merge! Thanks!

PR #762, opened 2024-11-21
xtest: add pkcs11_1031 for CKM_RSA_X_509 sign/verify (take 2)

etienne-lms:
Basic test on signing/verifying with raw RSA mechanism CKM_RSA_X_509.

Replaces https://github.com/OP-TEE/optee_test/pull/757.

### DIFF
--- a/host/xtest/pkcs11_1000.c
+++ b/host/xtest/pkcs11_1000.c
@@ -10021,3 +10021,390 @@ out_close_lib:
 }
 ADBG_CASE_DEFINE(pkcs11, 1030, xtest_pkcs11_test_1030,
 		 "PKCS11: Test AES-GCM Encryption/Decryption");
+
+static int test_rsa_raw_operations(ADBG_Case_t *c,
+				    CK_SESSION_HANDLE session,
+				    uint32_t rsa_bits)
+{
+	CK_RV rv = CKR_GENERAL_ERROR;
+	CK_OBJECT_HANDLE public_key = CK_INVALID_HANDLE;
+	CK_OBJECT_HANDLE private_key = CK_INVALID_HANDLE;
+	CK_MECHANISM mechanism = {
+		CKM_RSA_PKCS_KEY_PAIR_GEN, NULL, 0
+	};
+	CK_MECHANISM sign_mechanism = {
+		CKM_RSA_X_509, NULL, 0
+	};
+	CK_ULONG modulus_bits = 0;
+	CK_BYTE id[] = { 123 };
+
+	CK_ATTRIBUTE public_key_template[] = {
+		{ CKA_ENCRYPT, &(CK_BBOOL){ CK_FALSE }, sizeof(CK_BBOOL) },
+		{ CKA_VERIFY, &(CK_BBOOL){ CK_TRUE }, sizeof(CK_BBOOL) },
+		{ CKA_WRAP, &(CK_BBOOL){ CK_FALSE }, sizeof(CK_BBOOL) },
+		{ CKA_MODULUS_BITS, &modulus_bits, sizeof(CK_ULONG) },
+	};
+
+	CK_ATTRIBUTE private_key_template[] = {
+		{ CKA_TOKEN, &(CK_BBOOL){ CK_FALSE }, sizeof(CK_BBOOL) },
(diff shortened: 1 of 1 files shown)


etienne-lms:
Updated with a more robust message padding scheme.

host/xtest/pkcs11_1000.c
```diff
@@ -10138,29 +10139,26 @@ static int test_rsa_raw_operations(ADBG_Case_t *c,
 	if (!ADBG_EXPECT_CK_OK(c, rv))
 		goto err;
 
+	/*
+	 * Size of the message to sign must be at most the size of the private
+	 * key. If smaller, it is strongly recommended to inserrt padding
```
jforissier, line 10144: insert

host/xtest/pkcs11_1000.c
```diff
@@ -10138,29 +10139,26 @@ static int test_rsa_raw_operations(ADBG_Case_t *c,
...
+	 * bytes to reach to key size. Lets's use random data and use PKCS v1.5
+	 * padding scheme to ensure input data to be signed will generate well
+	 * szied signature.
```
jforissier, line 10147: sized

etienne-lms:
Updated removing the padding scheme. Strong padding is not needed as far as PKCS#11 regression tests are concerned.
The issue I faced with the generated signature size are due to in issue in the PKCS#11 TA implementation, fixed with https://github.com/OP-TEE/optee_os/pull/7179.

etienne-lms:
Fixed an omission in the previous fixup commit. CI test should be good now.

host/xtest/pkcs11_1000.c
```diff
@@ -10021,3 +10021,386 @@ static void xtest_pkcs11_test_1030(ADBG_Case_t *c)
...
+			       private_key_template,
+			       ARRAY_SIZE(private_key_template),
+			       &public_key, &private_key);
+	if (!ADBG_EXPECT_CK_OK(c, rv))
+		goto err;
+
+	/*
+	 * Current implementation of the PKCS#11 TA requires that the
+	 * message to sign has to size of the private key. There is
+	 * no constraint regarding the padding scheme. Lets's use a
```
etienne-lms, line 10144: This is wrong: there is 1 constraint: the message shall represent a value smaller than the key modulus. We need to at least clear the leading bit (`in_data[0] &= 0x7f;`) top achieve that.

By the way, typo above ("has **~to~ the** size of the private key.")

etienne-lms:
Fixed message leading bit + remove some useless tests on key attributes.

host/xtest/pkcs11_1000.c
```diff
@@ -10021,3 +10021,307 @@ static void xtest_pkcs11_test_1030(ADBG_Case_t *c)
...
+	rv = C_Sign(session, in_data, in_data_size, signature, &signature_len);
+	if (!ADBG_EXPECT_CK_RESULT(c, CKR_OPERATION_NOT_INITIALIZED, rv))
+		goto err_destr_obj;
+
+	Do_ADBG_EndSubCase(c, "Sign with invalid data");
+
+	Do_ADBG_BeginSubCase(c, "Sign/Verify with valid data/signature");
+
+	/*
+	 * Test C_Sign() with buffer too short, the C_Sign() operaiton should remain
```
jforissier, line 10132: operation

host/xtest/pkcs11_1000.c
```diff
@@ -10021,3 +10021,307 @@ static void xtest_pkcs11_test_1030(ADBG_Case_t *c)
...
+
+	rv = C_Verify(session, in_data, in_data_size, signature, signature_len);
+	if (!ADBG_EXPECT_CK_OK(c, rv))
+		goto err_destr_obj;
+
+	Do_ADBG_EndSubCase(c, "Sign/Verify with valid data/signature");
+
+	Do_ADBG_BeginSubCase(c, "Verify with altered signature/message");
+
+	/* Test signature wider than expect */
```
jforissier, line 10182: expected

jforissier (commented):
LGTM, just two misspelled words, see below.

etienne-lms:
Thanks for the review. However, this P-R cannot be merged before https://github.com/OP-TEE/optee_os/pull/7179. Without this OP-TEE core fix, we way have failures with this added `pkcs11_1031` test case because it can happen the generated signature it not well sized.

PR #763, opened 2024-11-26
xtest: pkcs11_1021: skip CKM_RSA_PKCS test is not supported

etienne-lms:
Skip CKM_RSA_PKCS test in pkcs11_1021 is the mechanism is not supported. This is likely to append when OP-TEE core is built with CFG_CRYPTO_RSASSA_NA1 disabled as when built with MbedTLS as core crypto library.


jforissier (commented):
With s/append/happen/:

jforissier:
...and "s/is not supported/if not supported/"

etienne-lms:
Commit message fixed and review tag applied.

PR #766, opened 2024-12-10
regression: add case 1041 for fTPM

jenswi-linaro:
Add regression case 1041 to test if the system has a fTPM TA and if so check that it has been probed.

### DIFF
--- a/host/xtest/ftpm_ta.h
+++ b/host/xtest/ftpm_ta.h
@@ -1,0 +1,51 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2024 Linaro Limited
+ *
+ * The copyright in this software is being made available under the BSD License,
+ * included below. This software may be subject to other third party and
+ * contributor rights, including patent rights, and no such rights are granted
+ * under this license.
+ *
+ * Copyright (c) 2018 Microsoft Corporation
+ *
+ * All rights reserved.
+ *
+ * BSD License
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ * Redistributions of source code must retain the above copyright notice,
+ * this list of conditions and the following disclaimer.
+ *
+ * Redistributions in binary form must reproduce the above copyright
+ * notice, this list of conditions and the following disclaimer in the
+ * documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS
+ * IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED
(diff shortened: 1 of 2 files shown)


host/xtest/ftpm_ta.h
```diff
@@ -0,0 +1,51 @@
...
+ * TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A
+ * PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+ * PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+ * LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
```
jforissier, line 38: The whole license text is not really welcome, but I suppose you prefer to have an exact copy of the header file from the fTPM TA tree?

jenswi-linaro: Yeah, the problem is that the license text isn't identical to the boilerplate at https://spdx.org/licenses/BSD-2-Clause.html since they have [1] at the top. We could perhaps keep [1] and drop the rest since the rest only differs by omitting the numbers if I'm not mistaken.

[1]:
```
 * The copyright in this software is being made available under the BSD License,
 * included below. This software may be subject to other third party and
 * contributor rights, including patent rights, and no such rights are granted
 * under this license.
```

jenswi-linaro:
Removed the license part as proposed above, and tag applied.

jforissier:
The CI error should be addressed by https://github.com/OP-TEE/optee_test/pull/767.

host/xtest/ftpm_ta.h
```diff
@@ -0,0 +1,26 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2024 Linaro Limited
+ *
+ * The copyright in this software is being made available under the BSD License,
+ * included below. This software may be subject to other third party and
```
etienne-lms, line 6: "included below" should be rephrased :(
Maybe:
```
 * The copyright in this software is being made available under the BSD-2-Clause
 * License. This software may be subject to other third party and
 * ...

jenswi-linaro: Argh, that's changing the license text. I'll restore the removed portion.

etienne-lms:

jenswi-linaro:
Tag applied.

PR #767, opened 2024-12-17
ci: update container

jforissier:
Update the CI container to jforissier/optee_os_ci:qemu_check [1] which is the latest one, usable for both 32-bit and 64-bit ARM, and which is used by the optee_os CI [2]. This one should be able to build QEMU 9.2.0 without any issue contrary to the current one which causes:

 *** Ouch! ***

 found no usable tomli, please install it

Link: https://github.com/jforissier/docker_optee_os_ci/tree/qemu_check [1]
Link: https://github.com/OP-TEE/optee_os/blob/4.4.0/.github/workflows/ci.yml#L319 [2]

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ jobs:
   QEMUv8_check:
     name: make check (QEMUv8)
     runs-on: ubuntu-latest
-    container: jforissier/optee_os_ci:qemuv8_check2
+    container: jforissier/optee_os_ci:qemu_check
     steps:
       - name: Remove /__t/*
         run: rm -rf /__t/*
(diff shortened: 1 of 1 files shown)


etienne-lms:
`Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>` FWIW.
That said, it seems both optee_os CI still have the issue and the P-R also failed to build Qemu.

jforissier:
> `Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>` FWIW. That said, it seems both optee_os CI still have the issue and the P-R also failed to build Qemu.

Correct. I also need to add the `python3-tomli` package to the image. Using the same image for CI here and in optee_os is a good thing anyways. Thanks for the tag.

etienne-lms:
Do you mean we should restart the CI jobs that failed on optee_os repo?

jforissier:
Yes, but I need to push the updated image to the Docker hub first. The upload is in progress.

jforissier:
Docker image updated on dockerhub (with the missing python3-tomli package added). CI should hopefully pass. I will restart the failed jobs in optee_os.

jforissier:
Needs https://github.com/OP-TEE/build/pull/799

PR #768, opened 2025-01-08
xtest: remove CFG_SECSTOR_TA_MGMT_PTA dependency

etienne-lms:
Makes xtest tool more flexible regarding OP-TEE configuration of `CFG_SECSTOR_TA_MGMT_PTA`.

Embed TA install interface in xtest even if devkit configuration says the secure storage TA management PTA service is not available. If the service is not found (secstor TA management PTA) print an explicit error message instead.

This change does not change xtest regression suite behavior and will still fails if one uses `--install-ta` option when the effective embedded TEE does not provide this service. This change adds an explicit error message when so.

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -53,6 +53,7 @@ srcs +=	adbg/src/adbg_case.c \
 	regression_8000.c \
 	regression_8100.c \
 	hash_perf.c \
+	install_ta.c \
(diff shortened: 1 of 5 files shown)


jenswi-linaro (commented):

PR #769, opened 2025-01-15
regression_1041: give 5 seconds for /dev/tpm0 to appear

jforissier:
It does take some time for the kernel to probe the fTPM device, and therefore if regression_1041 is run immediately after boot it may fail. Add a 5 second timeout to make the test reliable.

### DIFF
--- a/host/xtest/regression_1000.c
+++ b/host/xtest/regression_1000.c
@@ -3371,6 +3371,8 @@ static void xtest_tee_test_1041(ADBG_Case_t *c)
 	TEEC_Session sess = { };
 	uint32_t ret_orig = 0;
 	struct stat sb = { };
+	bool found = false;
+	int i = 0;
 
 	res = xtest_teec_open_session(&sess, &(const TEEC_UUID)TA_FTPM_UUID,
 				      NULL, &ret_orig);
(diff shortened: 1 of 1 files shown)


etienne-lms (commented):

host/xtest/regression_1000.c
```diff
@@ -3381,7 +3383,19 @@ static void xtest_tee_test_1041(ADBG_Case_t *c)
...
+	 * Check for the presence of the TPM device. Give the kernel some time
+	 * (5 seconds) otherwise the test may fail if run immediately after
+	 * boot.
+	 */
+	for (i = 0; i < 5; i++) {
+		if (stat(fname, &sb) == 0) {
+			found = true;
+			break;
+		}
+		(void)usleep(1000000);
```
jenswi-linaro, line 3396: `sleep(1)` without a void cast?

host/xtest/regression_1000.c
```diff
@@ -3381,7 +3383,19 @@ static void xtest_tee_test_1041(ADBG_Case_t *c)
...
+	}
+	if (!found) {
```
jenswi-linaro, line 3398: Shouldn't we record an error?

host/xtest/regression_1000.c
jforissier: Yes

host/xtest/regression_1000.c
jforissier: Indeed. Updated.

jenswi-linaro (commented):

PR #770, opened 2025-01-24
regression_4000: split test cases per algo family

etienne-lms:
Split regression_4006 test cases into specific test cases for each of RSA, DSA, ECDSA and SM2. This eases testing only a target algo family without needing to run all the cases series, especially useful when testing all algo flavors (max test level with option "-l 15").

By the way, fix the test cases description that mention "Asymm Cipher" while both ciphering and authentication operation are tested.


jenswi-linaro (commented):

jforissier (commented):
That's convenient indeed.

`Tested-by: Jerome Forissier <jerome.forissier@linaro.org> (vexpress-qemu_armv8a)`

PR #771, opened 2025-01-28
ta: os_test: fix TA time wrap test

jenswi-linaro:
In the test for wrapped TA time in test_time() we check that the time has wrapped and that not too much time has passed. The limit for this is currently 2 seconds of which 1 second is passed in TEE_Wait(). This is currently not enough time with the recent introduction of fTPM being probed in parallel with the test. So in order to avoid occasional false negatives increase the limit to 60 seconds.

Fixes: a2c1ce3a8c31 ("ta: os_test: fix TA time wrap test")

### DIFF
--- a/ta/os_test/os_test.c
+++ b/ta/os_test/os_test.c
@@ -600,7 +600,7 @@ static TEE_Result test_time(void)
 	 * TEE_GetTAPersistentTime() should be much less than 1 second, in fact
 	 * it's not even a millisecond.
 	 */
-	if (t.seconds > 1 || t.millis >= 1000) {
+	if (t.seconds > 60 || t.millis >= 1000) {
 		EMSG("Unexpected stored TA time %"PRIu32".%03"PRIu32, t.seconds,
 		     t.millis);
 		return TEE_ERROR_BAD_STATE;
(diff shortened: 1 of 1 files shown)


jforissier (commented):

It might indicate that fTPM is holding resources for too long though...

jenswi-linaro:
Yeah, I don't know. Maybe it's only starving the TA for CPU time.

jenswi-linaro:
Tag applied

ta/os_test/os_test.c
```diff
@@ -623,7 +623,7 @@ static TEE_Result test_time(void)
 	}
 	MSG("TA time %"PRIu32".%03"PRIu32, t.seconds, t.millis);
 
-	if (t.seconds > 1) {
+	if (t.seconds > 60) {
```
etienne-lms, line 626: Test at line 603 is not impacted by the same CPU overload issue?

jenswi-linaro: I don't know, it's at least far less likely since we're not exiting to the normal world in between.

etienne-lms: Maybe fragile, depends on the scheduler. Won't this trigger rare CI test failures?

jenswi-linaro: Sure, I'll increase the time there too.

jenswi-linaro:
Update

etienne-lms (commented):

jenswi-linaro:
Tag applied

PR #772, opened 2025-01-30
regression_4006: explicit ECDSA tests are SHA1 based

etienne-lms:
Explicitly use algorithm ID TEE_ALG_TEE_ECDSA_SHA1 for current ECDSA tests in regression_4006 instead of old API ID TEE_ALG_ECDSA_P* that are deprecated in GP TEE Internal Core API v1.3.1. This change is required to allow changes in OP-TEE OS where TEE_ALG_ECDSA_P* will no more straight alias IDs TEE_ALG_ECDSA_SHA* because the GP TEE specification says ECDSA key size is not related to the hash algorithm.

Duplicate however few tests using legacy _OPTEE_ALG_ECDSA_P* for testing these OP-TEE specific still supported IDs.

The key size for the related tests is found from the NIST key data material used in the test instead of the crypto algorithm ID when it does not provide this information.

By the way, add an explicit inline comment telling why we currently force SHA1 hashing for TEE_MAIN_ALGO_ECDSA regression_4006 tests.


host/xtest/regression_4000.c
```diff
@@ -4186,6 +4211,11 @@ static void xtest_tee_test_4006(ADBG_Case_t *c,
 		 * the payload.
 		 */
 		if (tv->mode == TEE_MODE_VERIFY || tv->mode == TEE_MODE_SIGN) {
+			/*
+			 * Current ECDSA tests are based on SHA1 hashed payload.
+			 * Current RSASSA_PKCS1_V1_5 are based on SHA25 hashed
```
jforissier, line 4216: SHA256

host/xtest/regression_4000.c
```diff
@@ -4351,27 +4381,51 @@ static void xtest_tee_test_4006(ADBG_Case_t *c,
...
+					break;
+				case 32:
+					curve = TEE_ECC_CURVE_NIST_P192;
+					break;
+				case 48:
+					curve = TEE_ECC_CURVE_NIST_P192;
+					break;
+				case 66:
+					curve = TEE_ECC_CURVE_NIST_P192;
+					break;
```
jforissier, line 4404: Wrong curve values

jforissier, line 4404: Wrong curve values

host/xtest/regression_4000.c
```diff
@@ -4186,6 +4211,11 @@ static void xtest_tee_test_4006(ADBG_Case_t *c,
 		 * the payload.
 		 */
 		if (tv->mode == TEE_MODE_VERIFY || tv->mode == TEE_MODE_SIGN) {
+			/*
+			 * Current ECDSA tests are based on SHA1 hashed payload.
+			 * Current RSASSA_PKCS1_V1_5 tests are based on SHA256
+			 * hashed payload.
+			 */
 			if (TEE_ALG_GET_MAIN_ALG(tv->algo) == TEE_MAIN_ALGO_ECDSA)
 				hash_algo = TEE_ALG_SHA1;
```
jenswi-linaro, line 4220: How about deriving this from the algorithm in the case of TEE_ALG_ECDSA_SHA*?

etienne-lms:
Comments (bugs) addressed.
This change depends on https://github.com/OP-TEE/optee_os/pull/7249 to allow `TEE_ALG_ECDSA_SHA1` to be used with key bigger than 192bits. CI test will likely pass but it would fail with test level 1 or above where `TEE_ALG_ECDSA_SHA1` is tested with keys larger than 192bit.

host/xtest/regression_4000.c
etienne-lms: I have the change ready to `__tee_alg_get_digest_hash()` to support ECDSA algo IDs. I create a P-R.

etienne-lms: Actually, such change requires 1-to-1 relation ship between `TEE_ALG_ECDSA_P*` and `TEE_ALG_ECDSA_SHA*` is cut. That what's proposed in https://github.com/OP-TEE/optee_os/pull/7251. I'll append a commit on that P-R.

etienne-lms:
Please keep alive. I'll work back on this topic soon...

PR #773, opened 2025-01-30
regression_4006: remove CFG_CRYPTO_RSASSA_NA1 dependency

etienne-lms:
Make xtest tool more flexible regarding OP-TEE configuration of CFG_CRYPTO_RSASSA_NA1. If algorithm TEE_ALG_RSASSA_PKCS1_V1_5 is not supported then it means either OP-TEE OS libraries or OP-TEE core were built with CFG_CRYPTO_RSASSA_NA1=n in which case simply don not test this algorithm.


jenswi-linaro (commented):

etienne-lms:
Review tags applied.

PR #775, opened 2025-02-04
regression_1041: increase ftpm timeout to 20 seconds

jenswi-linaro:
The timeout for /dev/tpm0 to appear was previously 5 seconds, but that's not always enough so increase it to 20 seconds to have a larger margin.


jenswi-linaro:
Needed by https://github.com/OP-TEE/optee_ftpm/pull/10

jforissier (commented):

jenswi-linaro:
Tag applied.

PR #781, opened 2025-03-07
xtest: stats: report heap fragmentation metric

jenswi-linaro:
A new field is added to struct pta_stats_alloc, free2_sum, the sum of size^2 of each free chunk. This can be used to calculate a fragmentation metric for the heap [1]. Some examples with xtest --stats --alloc:

When booted:
Pool:                Heap
Bytes allocated:                       21376
Fragmentation:			       3 %
Max bytes allocated:                   21488
Size of pool:                          202608

At some point during xtest:
Pool:                Heap
Bytes allocated:                       31616
Fragmentation:			       54 %
Max bytes allocated:                   101312
Size of pool:                          202608

When xtest has completed:
Pool:                Heap
Bytes allocated:                       25824
Fragmentation:			       33 %
Max bytes allocated:                   101312
Size of pool:                          202608

Link: [1] https://asawicki.info/news_1757_a_metric_for_memory_fragmentation

### DIFF
--- a/host/xtest/stats.c
+++ b/host/xtest/stats.c
@@ -8,6 +8,7 @@
 #include <errno.h>
 #include <fnmatch.h>
 #include <inttypes.h>
+#include <math.h>
 #include <pta_stats.h>
 #include <stdbool.h>
 #include <stdint.h>
@@ -154,6 +155,15 @@ static int stat_alloc(int argc, char *argv[])
 		       stats[n].desc);
 		printf("Bytes allocated:                       %"PRId32"\n",
 		       stats[n].allocated);
+		if (stats[n].free2_sum && stats[n].size != stats[n].allocated) {
+			double free2_sum = stats[n].free2_sum;
+			double free_sum = stats[n].size - stats[n].allocated;
+			double free_quote = sqrt(free2_sum) / free_sum;
+			double fragmentation = 1.0 - free_quote * free_quote;
+
+			printf("Fragmentation:			       %u %%\n",
+				(unsigned int)(fragmentation * 100.0));
+		}
 		printf("Max bytes allocated:                   %"PRId32"\n",
 		       stats[n].max_allocated);
 		printf("Size of pool:                          %"PRId32"\n",


jenswi-linaro:
Depends on https://github.com/OP-TEE/optee_os/pull/7302

jenswi-linaro:
Am I the only one thinking this is useful?

jforissier:
> Am I the only one thinking this is useful?

How can these numbers be used in practice? When I read the problem statement in the linked article, I don't think they apply, at least for now: IIRC heap cannot be expanded by requesting more memory from the OS, and there is no defragmentation algorithm in place either. Do you have something in mind?

jenswi-linaro:
> > Am I the only one thinking this is useful?
> 
> How can these numbers be used in practice? 

The numbers can be used for the same purpose as all the other fields in `struct pta_stats_alloc`. They give a clue about what's going on. When xtest has completed, we see that not all memory has been returned. The memory might be used for caching or keeping track of something, but the 33% value tells it wasn't allocated optimally since it now leaves the heap divided.

> When I read the problem statement in the linked article, I don't think they apply, at least for now: IIRC heap cannot be expanded by requesting more memory from the OS, and there is no defragmentation algorithm in place either. Do you have something in mind?

I think the linked article gives a nice background to how the reported value can be interpreted. 

By the way, I have patches in the pipe for growing the core heap on demand, but those have nothing to do with this.

jforissier:
> > > Am I the only one thinking this is useful?
> > 
> > 
> > How can these numbers be used in practice?
> 
> The numbers can be used for the same purpose as all the other fields in `struct pta_stats_alloc`. They give a clue about what's going on. 

Sure.

> When xtest has completed, we see that not all memory has been returned. The memory might be used for caching or keeping track of something, but the 33% value tells it wasn't allocated optimally since it now leaves the heap divided.
> 
> > When I read the problem statement in the linked article, I don't think they apply, at least for now: IIRC heap cannot be expanded by requesting more memory from the OS, and there is no defragmentation algorithm in place either. Do you have something in mind?
> 
> I think the linked article gives a nice background to how the reported value can be interpreted.
> 
> By the way, I have patches in the pipe for growing the core heap on demand, but those have nothing to do with this.

OK. I was thinking more about growing the TA heap, in which case higher fragmentation could lead to a large waste of memory which would otherwise have been available for other uses (core, other TAs).

Anyways the code looks good, so:

jenswi-linaro:
Update

host/xtest/stats.c
```diff
@@ -154,6 +155,15 @@ static int stat_alloc(int argc, char *argv[])
 		       stats[n].desc);
 		printf("Bytes allocated:                       %"PRId32"\n",
 		       stats[n].allocated);
+		if (stats[n].free2_sum) {
```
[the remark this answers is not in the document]
etienne-lms: Right! I missed that.

host/xtest/stats.c
```diff
@@ -154,6 +155,15 @@ static int stat_alloc(int argc, char *argv[])
 		       stats[n].desc);
 		printf("Bytes allocated:                       %"PRId32"\n",
 		       stats[n].allocated);
+		if (stats[n].free2_sum && stats[n].size != stats[n].allocated) {
+			double free2_sum = stats[n].free2_sum;
+			double free_sum = stats[n].size - stats[n].allocated;
+			double free_quote = sqrt(free2_sum) / free_sum;
+			double fragmentation = 1.0 - free_quote;
```
etienne-lms, line 162: From [metric_for_memory_fragmentation post](https://asawicki.info/news_1757_a_metric_for_memory_fragmentation), fragmentation metric should be `1.0 - free_quote * free_quote`.

jenswi-linaro: You're right, I'll update.

etienne-lms (commented):

jenswi-linaro:
Tag applied

PR #787, opened 2025-03-31
regression_1006: add test for malloc_add_pool()

jenswi-linaro:
Add a test for malloc_add_pool() to see that pools are merged.

Depends on https://github.com/OP-TEE/optee_os/pull/7342

### DIFF
--- a/ta/os_test/os_test.c
+++ b/ta/os_test/os_test.c
@@ -7,10 +7,12 @@
 #include <compiler.h>
 #include <dlfcn.h>
 #include <link.h>
+#include <malloc.h>
 #include <memtag.h>
 #include <setjmp.h>
 #include <stdint.h>
 #include <string.h>
+#include <sys/queue.h>
 #include <ta_crypt.h>
 #include <ta_os_test.h>
 #include <tee_internal_api_extensions.h>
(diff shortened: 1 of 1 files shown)


ta/os_test/os_test.c
```diff
@@ -758,20 +760,112 @@ static bool optee_pager_with_small_pool(void)
 	return rc;
 }
 
+struct consume_elem {
+	SLIST_ENTRY(consume_elem) link;
+};
+SLIST_HEAD(consume_elem_head, consume_elem);
+
+static void consume_all_heap(struct consume_elem_head *h)
```
jbech-linaro, line 768: Function does what it should do, but if 4096 isn't an upper limit and we just want to make sure to consume all heap, then an alternative and likely faster approach could be to start doubling the size until we get an error, and then after that half the size like you've done. I believe this could save a few loops and keep the linked list shorter. But, the question is whether it's worth doing something like that here? Perhaps just add an  `else { p_sz *= 2 }` to the innermost loop 
and take care of  `SLIST_INSERT_HEAD` accordingly would achieve this?

jbech-linaro (commented):
With or without suggestion:

jenswi-linaro:
Updated and tag applied.

PR #788, opened 2025-04-04
xtest: add test entry for transfer list

raymo200915:
Add test entry for transfer list in xtest 1001.

### DIFF
--- a/host/xtest/regression_1000.c
+++ b/host/xtest/regression_1000.c
@@ -303,18 +303,31 @@ static void xtest_tee_test_1001(ADBG_Case_t *c)
 
 	Do_ADBG_BeginSubCase(c, "Core self tests");
 
-	(void)ADBG_EXPECT_TEEC_SUCCESS(c, TEEC_InvokeCommand(
-		&session, PTA_INVOKE_TESTS_CMD_SELF_TESTS, NULL, &ret_orig));
+	res = TEEC_InvokeCommand(&session, PTA_INVOKE_TESTS_CMD_SELF_TESTS,
+				 NULL, &ret_orig);
+	ADBG_EXPECT_TEEC_SUCCESS(c, res);
 
 	Do_ADBG_EndSubCase(c, "Core self tests");
 
 	Do_ADBG_BeginSubCase(c, "Core dt_driver self tests");
 
-	(void)ADBG_EXPECT_TEEC_SUCCESS(c, TEEC_InvokeCommand(
-		&session, PTA_INVOKE_TESTS_CMD_DT_DRIVER_TESTS, NULL,
-		&ret_orig));
+	res = TEEC_InvokeCommand(&session, PTA_INVOKE_TESTS_CMD_DT_DRIVER_TESTS,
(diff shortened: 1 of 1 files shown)


raymo200915:
Add macros and helper function to avoid returning error for not supported tests.

host/xtest/regression_1000.c
```diff
@@ -318,7 +318,7 @@ static void xtest_tee_test_1001(ADBG_Case_t *c)
 
 	Do_ADBG_BeginSubCase(c, "Core transfer list self tests");
 
-	(void)ADBG_EXPECT_TEEC_SUCCESS(c, TEEC_InvokeCommand(
+	(void)ADBG_EXPECT_TEEC_SUCCESS_OR_NOTSUPPORT(c, TEEC_InvokeCommand(
```
jenswi-linaro, line 321: Please use a helper variable to store the result from TEEC_InvokeCommand() to improve the line breaks, and drop the `(void)`.
I know the rest of this function is different, but we need to start somewhere to improve the pattern.

host/xtest/xtest_test.h
```diff
@@ -34,6 +34,10 @@ ADBG_ENUM_TABLE_DECLARE(TEEC_Result);
 #define ADBG_EXPECT_TEEC_SUCCESS(c, got) \
 	ADBG_EXPECT_ENUM(c, TEEC_SUCCESS, got, ADBG_EnumTable_TEEC_Result)
 
+#define ADBG_EXPECT_TEEC_SUCCESS_OR_NOTSUPPORT(c, got) \
```
jforissier, line 37: If we are to truncate `NOT_SUPPORTED` (which I think is a good idea), I'd rather have `ADBG_EXPECT_TEEC_SUCCESS_OR_NOTSUPP`. I think it reads better, and also think about `EOPNOTSUPP`.

jenswi-linaro: I have my doubts about the macro. Is this hard to read or write?
``` C
res = TEEC_InvokeCommand(...);
if (res == TEE_ERROR_NOT_SUPPORTED)
        Do_ADBG_Log("PTA_INVOKE_TESTS_CMD_TRANSFER_LIST_TESTS not supported, skipping");
else
```

jforissier: I'm fine with that too, it's what we have been doing for ages. With the advantage that the message can be more detailed.

etienne-lms: I would also be in favor of Jens' explicit implementation.

raymo200915: The idea of my patch is just to respect and follow the existing pattern, if all of you agree I can change it to what @jenswi-linaro proposed.

host/xtest/regression_1000.c
raymo200915: Done.

jenswi-linaro:

jforissier (commented):

etienne-lms (commented):

raymo200915:
Added tags and rebased.

Issue #721, opened 2024-01-11
Queries related to PKCS11 based tests


anvisriv:
Hi Team

If repos ([optee_test](https://github.com/OP-TEE/optee_test/tree/4.0.0),[optee_client](https://github.com/OP-TEE/optee_client/tree/4.0.0),[optee_os](https://github.com/OP-TEE/optee_os/tree/4.0.0))with tag **4.0.0** is used then,

1. Can you please confirm whether all the PKCS11 related test cases are passing in OPTEE? If not, can you please highlight those testcases and the plan to fix them?
2. Some of the test cases(like 1008,1009.1020) require NULL input to be handled by returning the output size, although the GP specification(for TEE_CipherUpdate, TEE_MACComputeFinal) does not mention this requirement. These test cases are bound to fail where GP APIs are not implemented as the test case expects. Any reason why these test cases were designed in such a way?

etienne-lms:
Hi @anvisriv,

> 1. Can you please confirm whether all the PKCS11 related test cases are passing in OPTEE? 

All pkcs11 test from xtest are expected to pass.
OP-TEE CI tests check that using qemu_armv8a platform setup.

> 2. Some of the test cases(like 1008,1009.1020) require NULL input to be handled by returning the output size

Pkcs11 API functions differ from GP TEE API functions in how to get an output buffer size. With Pkcs11, to query an output buffer size, you can either pass a NULL pointer (in which case you get an CKR_OK status with the size output argument set) or you can pass a too small buffer (in which case you get a CKR_BUFFER_TOO_SMALL status and the size output argument set). pkcs11 xtest test are implemented using Pkcs11 cryptoki API functions.

anvisriv:
> Pkcs11 API functions differ from GP TEE API functions in how to get an output buffer size. With Pkcs11, to query an output buffer size, you can either pass a NULL pointer (in which case you get an CKR_OK status with the size output argument set) or you can pass a too small buffer (in which case you get a CKR_BUFFER_TOO_SMALL status and the size output argument set). pkcs11 xtest test are implemented using Pkcs11 cryptoki API functions.

Can you please tell me where the output size is calculated for the ‘NULL buffer case with size as 0’ [subtest](https://github.com/OP-TEE/optee_test/blob/4.0.0/host/xtest/pkcs11_1000.c#L1985) of PKCS11 Test 1008? 
Is it calculated inside the PKC11 cryptoki Client library or inside the PKCS11 TA? 
[...]

PR #789, opened 2025-04-06
xtest: pkcs11: Update test_digest for test 1019 to use SHA-512

anvisriv:
Modified test 1019 to use digest_test_pattern_sha512 instead of digest_test_pattern_sha256.

Addressed [comment](https://github.com/OP-TEE/optee_test/issues/721#issuecomment-1952822410) in the issue#721.

### DIFF
--- a/host/xtest/pkcs11_1000.c
+++ b/host/xtest/pkcs11_1000.c
@@ -5767,8 +5767,8 @@ static int test_ec_operations(ADBG_Case_t *c, CK_SESSION_HANDLE session,
 	if (!ADBG_EXPECT_CK_OK(c, rv))
 		goto err_destr_obj;
 
-	rv = C_Sign(session, (void *)digest_test_pattern_sha256,
-		    sizeof(digest_test_pattern_sha256), (void *)signature,
+	rv = C_Sign(session, (void *)digest_test_pattern_sha512,
+		    sizeof(digest_test_pattern_sha512), (void *)signature,
 		    &signature_len);
 	if (!ADBG_EXPECT_CK_OK(c, rv))
 		goto err_destr_obj;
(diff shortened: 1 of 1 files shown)


jforissier:
Why?

anvisriv:
Hi @jforissier 

The GP TEE_AsymmetricSignDigest API requires that when a hash algorithm is specified, the digestLen must match the length of the digest for that algorithm. However, for the mechanisms CKM_ECDSA_SHA384 and CKM_ECDSA_SHA512, the buffer(digest_test_pattern_sha256) will be shorter. This change aims to address that discrepancy.

jforissier:
@anvisriv OK so this should be explained in the commit description. Is there a bug in the PKCS#11 TA? I would expect `C_Sign()` to return an error if the length is invalid. What does the PKCS#11 specification say?

anvisriv:
Hi @jforissier 

It appears that the PKCS11 specification does not impose this type of restriction.
https://docs.oasis-open.org/pkcs11/pkcs11-curr/v3.0/os/pkcs11-curr-v3.0-os.html#_Toc30061189

anvisriv:
Hi @etienne-lms Can you please review?

anvisriv:
Hi @etienne-lms 
Any update on this?

etienne-lms (commented):
I'm not sure this change addresses the issue. The pkcs11 TA implementation for `CKM_ECDSA` does currently uses GP TEE IDs `TEE_ALG_ECDSA_SHAxxx` (through legacy IDs `TEE_ALG_ECDSA_Pxxx`) depending on key size whereas it would need an extension `TEE_ALG_ECDSA_RAW` or like, to not bound digest size to a non-existent hash size. See https://github.com/OP-TEE/optee_os/issues/7235.

Whereas this change induce no regression, I think we would rather need changes in the pkcs11 TA implementation and change this test, consider several precomputed hash test vectors: `digest_test_pattern_sha256`, `digest_test_pattern_sha384` and `digest_test_pattern_sha512`.
 stated in the commit message.

anvisriv:
Commenting it so that issue is not marked stale

anvisriv:
Commenting it so that issue is not marked stale

PR #792, opened 2025-04-16
subkey: Regenerate the subkey signatures

jbech-linaro:
We have increased the size of the default key in optee_os, necessitating re-signing the keys used in the subkey tests (xtest 1039).

This depends on https://github.com/OP-TEE/optee_os/pull/7369. These should be merged at the same time.


jenswi-linaro:
While the steps needed to generate these files are documented in `ta/subkey_notes.txt`, it would be nice to have the actual commands in the commit message, akin to what we do when updating a third-party library in optee_os.git.

jbech-linaro:
> While the steps needed to generate these files are documented in `ta/subkey_notes.txt`, it would be nice to have the actual commands in the commit message, akin to what we do when updating a third-party library in optee_os.git.

Yes, I was thinking about it and I also had an idea of writing a simple script running the commands in the `ta/subkey_notes.txt`. So, commit message, yes I'll add. Script, would that be worth considering?

jenswi-linaro:
A script might be a bit much, but if you like it, I don't mind.

jbech-linaro:
@jenswi-linaro , commit message has been updated.

jenswi-linaro:

jbech-linaro:
Squashed, rebased and tag(s) applied (should be) ready to be merged together with https://github.com/OP-TEE/optee_os/pull/7369  after the ongoing release! Thanks!

PR #793, opened 2025-04-24
Test a logical SP for ffa_spmc suite + a fix

jenswi-linaro:
Depends on https://github.com/OP-TEE/optee_os/pull/7378 and https://github.com/OP-TEE/build/pull/827


jenswi-linaro:
Comment addressed, tag applied.

PR #794, opened 2025-04-25
regression_4017: Test few one-by-one byte cases in level 0

etienne-lms:
Change default value of `middle_count` in regression test 4017 when test level is below 12. Instead of 0, use 2. This allows to pass few tests feeding data byte per byte without increasing the test duration as when test level is 12 or above.


jenswi-linaro:

jforissier (commented):
- For "regression_4017: Test few one-by-one byte case in level 0":
- For "regression_4017: remove dead code":

PR #797, opened 2025-05-07
build: make, cmake: add -Werror based on CFG_WERROR

jforissier:
Update the build files that currently set -Werror unconditionally to use set it based on the value of CFG_WERROR instead. CFG_WERROR is set to y by default, but may be unset (or set to n) from the build command line or the environment.


mikkorapeli-linaro:
Thanks! This is useful in distro integration when toolchains are updated etc.

jenswi-linaro:
I think it should be disabled by default. That's what we have in optee_os.
Sigh, I see now that we have it enabled by default in optee_client.

jforissier:
> I think it should be disabled by default. That's what we have in optee_os. Sigh, I see now that we have it enabled by default in optee_client.

Hmmm yes let's make it disabled by default.

jenswi-linaro:

PR #815, opened 2026-05-21
xtest: Add Application Secrets TA testsuite

tusal-vaisala:
Related optee_os PR: https://github.com/OP-TEE/optee_os/pull/7769

Addresses issue: https://github.com/OP-TEE/optee_os/issues/7768

### DIFF
--- a/host/xtest/CMakeLists.txt
+++ b/host/xtest/CMakeLists.txt
@@ -111,6 +111,11 @@ if (CFG_PKCS11_TA)
 	list (APPEND SRC pkcs11_1000.c)
 endif()
 
+if (CFG_APP_SECRETS_TA)
+	add_compile_options(-DCFG_APP_SECRETS_TA)
+	list (APPEND SRC asteec_1000.c)
+endif()
+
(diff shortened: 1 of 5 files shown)


tusal-vaisala:
keep-alive

tusal-vaisala:
keep-alive

vesajaaskelainen:
Seal path is kinda easy to test as only input is given from user space but unseal path could have some extra tests.

vesajaaskelainen:
Please added unseal test that magic is invalid.

vesajaaskelainen:
Please add unseal test that version is invalid... I suppose 0 and 2 could be tested and why not (unsiged)-1.

vesajaaskelainen:
Please add unseal test that has extra bytes added and over the range test.

vesajaaskelainen:
Please add unseal test for valid input but too small output buffer

tusal-vaisala:
> Seal path is kinda easy to test as only input is given from user space but unseal path could have some extra tests.

Added suggested unseal path cases (and a little bit refactored to reduce code duplication)

PR #816, opened 2026-06-01
test: add ECDH invalid curve attack regression test (Wycheproof)

sahilnxp:
Add regression_4023 to verify that ECDH key derivation rejects peer public keys that do not lie on the negotiated curve (secp256r1). This protects against invalid curve attacks as documented in Wycheproof test vector tcId 335.

The test supplies an off-curve peer public key to TEE_DeriveKey() and expects the TA to panic (TEEC_ERROR_TARGET_DEAD), confirming that the underlying public key validation in OP-TEE core correctly detects and rejects the invalid point before any shared secret is computed.

Also update the ret_orig check in ta_crypt_cmd_derive_key() to accept both TEEC_ORIGIN_TRUSTED_APP and TEEC_ORIGIN_TEE, since a validation failure in the core crypto layer causes TEE_DeriveKey() to panic, resulting in TEEC_ORIGIN_TEE rather than TEEC_ORIGIN_TRUSTED_APP.

### DIFF
--- a/host/xtest/regression_4000.c
+++ b/host/xtest/regression_4000.c
@@ -2,7 +2,7 @@
 /*
  * Copyright (c) 2014, STMicroelectronics International N.V.
  * Copyright (c) 2021, SumUp Services GmbH
- * Copyright 2023 NXP
+ * Copyright 2023, 2026 NXP
  */
 
 #include <stdio.h>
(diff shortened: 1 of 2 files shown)


sahilnxp:
Related to https://github.com/OP-TEE/optee_os/pull/7828

PR #817, opened 2026-06-04
xtest: regression 4005: test aes-gcm overflow on A64

MingyenHung:
Set the input size of the new test case to 17 bytes (1 byte + 1 block). This forces internal_aes_gcm_update_payload_blocks() to bypass update_payload_2block() and fall back to decrypt_pl() instead, to properly trigger and verify the overflow boundary condition.

This test case helps to reproduce the counter overflow issue caused by ce_aes_gcm_ctr_encrypt() on ARM64. 
Note that for the ARM32 case, the same issue can already be reproduced simply by running `xtest -t regression 4005`.


host/xtest/aes_gcm_counter_overflow_test_decrypt.h
```diff
@@ -687,3 +687,23 @@
 	.line = __LINE__,
 	.id = 251
 },
+{
```
jforissier: This header file was generated from https://github.com/C2SP/wycheproof/blob/4f5e05f71e6b724c20e2c1b6934c7bd7ef6d89e7/testvectors_v1/aes_gcm_test.json (see this commit: https://github.com/OP-TEE/optee_test/commit/9d566212b3cee7ea362515ccce573c64aaa54fbe) so I dont think it is a good idea to change it (for example `.id = 252` does not reflect test 252 from the upstream test vector). I would prefer you add a separate header file.

MingyenHung: Fixed by adding a new header, aes_gcm_counter_overflow_test_corner.h, to include this corner case.

jforissier: Thanks

jforissier:

MingyenHung:
> `Reviewed-by: Jerome Forissier <jerome.forissier@arm.com>`

PR #818, opened 2026-06-09
xtest: add test for ECDH with invalid points

jenswikl:
Adds four subtests checking that ECDH fails when the public key is a point that is not on the correct curve, which need to be rejected to avoid invalid curve attacks. Since TEE_DeriveKey() doesn't allow returning errors the TA will panic in the correct case.

The tests use four test vectors from Wycheproof, selected in an attempt to avoid failures due to other checks that are independent of the curve check.

[JW: fixup to use ta_crypt_cmd_derive_key()]


jenswikl:
Based on the attached xtest patch in https://github.com/OP-TEE/optee_os/security/advisories/GHSA-g6qf-hwf7-mg9h

jforissier:
`Acked-by: Jerome Forissier <jerome.forissier@arm.com>`

jenswikl:
Squashed and tag applied.

PR #821, opened 2026-06-23
pkcs11: fix undefined behavior in attribute initialization (compound literals)

sahesaha:
**Problem:**
Incorrect attribute values are observed in test_create_destroy_session_objects,
CKA_TOKEN = 0xe0 (expected CK_FALSE / 0)
The issue appears with default toolchain optimizations, even though no explicit -O2 is set. It disappears when forcing -O0, indicating optimization-dependent undefined behavior.

**Root cause**
Attributes are initialized using compound literals, e.g.:
C&(CK_BBOOL){CK_FALSE}
These create temporary objects whose addresses are stored and reused.
Their lifetime and storage are not guaranteed, and under optimization the compiler may reuse or overlap them, leading to invalid values.

**Why not fix via -O0**
Disabling optimization is not a valid solution:

Hides undefined behavior instead of fixing it
Not portable across compilers/toolchains
Impacts performance in security-sensitive code

**Fix**
Replace compound literals with static variables having stable storage:
e.g. static const CK_BBOOL ck_false = CK_FALSE;

### DIFF
--- a/host/xtest/pkcs11_1000.c
+++ b/host/xtest/pkcs11_1000.c
@@ -1201,14 +1201,18 @@ static CK_ATTRIBUTE cktest_token_object[] = {
 	{ CKA_VALUE,	(void *)cktest_aes128_key, sizeof(cktest_aes128_key) },
 };
 
+static const CK_BBOOL ck_true = CK_TRUE;
+static const CK_BBOOL ck_false = CK_FALSE;
+static const CK_OBJECT_CLASS secret_key_class = CKO_SECRET_KEY;
+static const CK_KEY_TYPE aes_key_type = CKK_AES;
+
 static CK_ATTRIBUTE cktest_session_object[] = {
-	{ CKA_DECRYPT,	&(CK_BBOOL){CK_TRUE}, sizeof(CK_BBOOL) },
-	{ CKA_TOKEN,	&(CK_BBOOL){CK_FALSE}, sizeof(CK_BBOOL) },
-	{ CKA_MODIFIABLE, &(CK_BBOOL){CK_TRUE}, sizeof(CK_BBOOL) },
-	{ CKA_KEY_TYPE,	&(CK_KEY_TYPE){CKK_AES}, sizeof(CK_KEY_TYPE) },
-	{ CKA_CLASS,	&(CK_OBJECT_CLASS){CKO_SECRET_KEY},
-						sizeof(CK_OBJECT_CLASS) },
-	{ CKA_VALUE,	(void *)cktest_aes128_key, sizeof(cktest_aes128_key) },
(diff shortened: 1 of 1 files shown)


sahesaha:
Hi @etienne-lms , could you review this PR?

jenswikl:
Please add a commit message and a Signed-off-by as described at https://optee.readthedocs.io/en/latest/general/contribute.html

jenswikl:
Regarding:
> Attributes are initialized using compound literals, e.g.:
> C&(CK_BBOOL){CK_FALSE}
> These create temporary objects whose addresses are stored and reused.
> Their lifetime and storage are not guaranteed, and under optimization the compiler > may reuse or overlap them, leading to invalid values.

Do you have any reference to back up this claim? To me, it seems like as long as there's a reference to the compound literal, it can't be reused.

sahesaha:
Even though the reference to the compound literal exists, we observe clear optimization-dependent behavior in practice.

**Evidence**

1. Reproducible on fresh runs
After a clean flash and first execution of test 1004, the issue is consistently observed:
CKA_TOKEN = 0xe0
→ This rules out any persistent storage or prior state corruption.

2. Occurs before any API call
A pre-call dump of the attribute array shows:
[1] type=CKA_TOKEN ... val=0xe0
→ The incorrect value is already present before the template is passed to any PKCS#11 function.
→ This indicates the issue originates from how the attribute values are stored, not from downstream processing.

3. volatile does not change behavior
Using:
Volatile CK_ATTRIBUTE attrs[];Show more lines
still results in the same corrupted value.
→ This suggests the problem is not due to simple compiler reordering or caching, but related to how the underlying storage for the compound literals is handled under optimization.

**Interpretation**
These observations indicate that the storage backing the compound literals is not providing stable, reusable memory under optimization, even though a reference exists.

jenswikl:
You're right. So the problem is static initialization with a compound literal address. Apparently, C23 is supposed to address that by using the `static` keyword in the "cast" part. But it's a bit early to require C23 support yet. So your fix is the cleanest.

Please mention that you're fixing a problem with static initialization with a compound literal address in the commit message.

sahesaha:
Hi @jenswikl, please check now. What is the next step for this PR?

jenswikl:
Static initialization using addresses of compound literals is used quite extensively in this file. It would make sense to fix all of it in this file. @etienne-lms, what do you think?

sahesaha:
Hi @jenswikl @etienne-lms, could you please confirm whether we should fix all the Static initialization using addresses of compound literals in this file or we can proceed to get this PR merged?

jenswikl:
Since this fixes a problem for you, let's take this first to avoid further delays. Please apply:
`Reviewed-by: Jens Wiklander <jens.wiklander@oss.qualcomm.com>`

sahesaha:
Addressed. Please help to proceed @jenswikl .

PR #828, opened 2026-08-27
FP context switch xtest

dave-patel-riscstar:
xtest for FP 

A TA runs with the floating-point unit disabled and is handed a context on
the first floating-point instruction it executes, so every sub-test starts
by taking that trap. What they then cover is:

```
  SYSCALL      a syscall serviced entirely inside the TEE, where the TA
               context is saved on the way in and has to be given back on
               the first floating-point instruction after the return
  RPC          TEE_Wait(), which suspends the thread, runs the normal
               world and resumes through thread_resume_from_rpc()
  CRYPTO       a digest, which the TEE may compute in a secure kernel
               floating-point section of its own and so has to take the
               registers from the TA and hand them back
  TAINT        leaves a pattern in the registers and returns
  CHECK_TAINT  reads the registers before writing any of them and fails
               if an earlier TA's pattern is still there
```

### DIFF
--- a/ta/os_test/include/riscv_fp_ctx.h
+++ b/ta/os_test/include/riscv_fp_ctx.h
@@ -1,0 +1,243 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2026, RISCStar Solutions Limited
+ */
+
+#ifndef RISCV_FP_CTX_H
+#define RISCV_FP_CTX_H
+
+#if defined(__riscv) && defined(__riscv_flen) && __riscv_flen == 64
+#define RISCV_FP_CTX_SUPPORTED	1
+
+#include <stddef.h>
+#include <stdint.h>
+
+/*
+ * The floating-point state the RISC-V calling convention requires a callee
+ * to preserve. Everything else, ft0..ft11 and fa0..fa7, may legitimately be
+ * clobbered by a call, so a caller cannot tell a context switching bug from
+ * a compiler doing what the ABI allows and those registers are left out of
+ * these checks.
+ *
+ * The layout is shared with riscv_fp_rv64.S.
+ */
+struct riscv_fp_ctx {
+	uint64_t fs[12];	/* fs0..fs11 */
+	uint32_t fcsr;
+};
+
+/*
+ * Installs @in, calls fn(arg), stores what is left of the context in @out
+ * and returns what fn returned. The caller's own context is preserved.
+ */
+unsigned long riscv_fp_ctx_roundtrip(const struct riscv_fp_ctx *in,
+				     struct riscv_fp_ctx *out,
+				     unsigned long (*fn)(void *), void *arg);
+
+/*
+ * Installs @in and returns with it still in the registers. This breaks the
+ * calling convention on purpose, the caller must have no live
+ * floating-point values of its own.
+ */
+void riscv_fp_ctx_load(const struct riscv_fp_ctx *in);
+
+/* Stores the current context to @out without changing it */
+void riscv_fp_ctx_store(struct riscv_fp_ctx *out);
+
+/*
+ * Fills @ctx with a value per register that is distinct, derived from
+ * @seed, and a finite double rather than a NaN or an infinity so that
+ * nothing traps should a value reach an arithmetic instruction. fcsr gets a
+ * rounding mode and accrued exception flags that differ from the reset
+ * value, so that a save or restore which forgets fcsr is caught too.
+ */
+static inline void riscv_fp_ctx_pattern(struct riscv_fp_ctx *ctx,
+					uint32_t seed)
+{
+	size_t n = 0;
+
+	for (n = 0; n < sizeof(ctx->fs) / sizeof(ctx->fs[0]); n++)
+		ctx->fs[n] = ((uint64_t)(0x3fd0 + n) << 48) |
+			     ((uint64_t)seed << 16) | (n + 1);
+
+	/*
+	 * frm = 2, round down, which differs from the reset value, and all
+	 * five accrued exception flags set. The flags are sticky and are
+	 * only ever cleared explicitly, so starting from all ones keeps the
+	 * comparison from tripping over a called function that happens to
+	 * do some arithmetic of its own, while a save or restore that drops
+	 * fcsr altogether still shows up.
+	 */
+	ctx->fcsr = (2 << 5) | 0x1f;
+}
+
+/*
+ * Returns the index of the first field of @a that differs from @b, 12 if
+ * only fcsr differs, and -1 if the two contexts are identical.
+ */
+static inline int riscv_fp_ctx_diff(const struct riscv_fp_ctx *a,
+				    const struct riscv_fp_ctx *b)
+{
+	size_t n = 0;
+
+	for (n = 0; n < sizeof(a->fs) / sizeof(a->fs[0]); n++)
+		if (a->fs[n] != b->fs[n])
+			return (int)n;
+
+	if (a->fcsr != b->fcsr)
+		return 12;
+
+	return -1;
+}
+
+/*
+ * The three routines above are written in assembly because the whole point
+ * is to have the values actually sitting in fs0..fs11 across a call, and
+ * from C the compiler would be free to keep them on the stack instead,
+ * which would test nothing.
+ *
+ * They are emitted here, rather than from a .S file, so that the same
+ * implementation serves the TA and xtest without either build system
+ * needing to grow an assembler rule. Exactly one translation unit per
+ * binary must define RISCV_FP_CTX_IMPLEMENTATION before including this.
+ *
+ * Comments inside the block use '#' rather than C syntax: these strings do
+ * not go through the preprocessor the way a .S file does, so the assembler
+ * sees them verbatim, and Clang's integrated assembler rejects a comment
+ * in the C form.
+ *
+ * Stack frame of riscv_fp_ctx_roundtrip():
+ *	   0	ra
+ *	   8	s0
+ *	  16	caller's fcsr
+ *	  24	caller's fs0..fs11
+ */
+#ifdef RISCV_FP_CTX_IMPLEMENTATION
+_Static_assert(offsetof(struct riscv_fp_ctx, fs) == 0,
+	       "riscv_fp_ctx layout out of sync with the assembly below");
+_Static_assert(offsetof(struct riscv_fp_ctx, fcsr) == 96,
+	       "riscv_fp_ctx layout out of sync with the assembly below");
+
+__asm__(
+"	.text\n"
+"\n"
+"	.globl	riscv_fp_ctx_roundtrip\n"
+"	.type	riscv_fp_ctx_roundtrip, @function\n"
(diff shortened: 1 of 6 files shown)


dave-patel-riscstar:
Add TA_OS_TEST_CMD_RISCV_FP_CONTEXT, which runs a set of sub-tests that
each install a known floating-point context, which runs different sets of test
out of the TA, and check that the context comes back intact.

A TA runs with the floating-point unit disabled and is handed a context on
the first floating-point instruction it executes, so every sub-test starts
by taking that trap. What they then cover is:

  SYSCALL      a syscall serviced entirely inside the TEE, where the TA
                      context is saved on the way in and has to be given back on
                      the first floating-point instruction after the return
  RPC             TEE_Wait(), which suspends the thread, runs the normal
                     world and resumes through thread_resume_from_rpc()
  CRYPTO      a digest, which the TEE may compute in a secure kernel
                     floating-point section of its own and so has to take the
                     registers from the TA and hand them back
  TAINT         leaves a pattern in the registers and returns
  CHECK_TAINT  reads the registers before writing any of them and fails
                     if an earlier TA's pattern is still there

dave-patel-riscstar:
@jenswikl 
Like #7951, this is also ready for merging.

ta/os_test/include/riscv_fp_ctx.h
```diff
@@ -0,0 +1,243 @@
...
+void riscv_fp_ctx_store(struct riscv_fp_ctx *out);
+
+/*
+ * Fills @ctx with a value per register that is distinct, derived from
+ * @seed, and a finite double rather than a NaN or an infinity so that
+ * nothing traps should a value reach an arithmetic instruction. fcsr gets a
+ * rounding mode and accrued exception flags that differ from the reset
+ * value, so that a save or restore which forgets fcsr is caught too.
+ */
+static inline void riscv_fp_ctx_pattern(struct riscv_fp_ctx *ctx,
```
jenswikl, line 54: Do these functions need to be `static inline`?

ta/os_test/include/riscv_fp_ctx.h
```diff
@@ -0,0 +1,243 @@
...
+					uint32_t seed)
+{
+	size_t n = 0;
+
+	for (n = 0; n < sizeof(ctx->fs) / sizeof(ctx->fs[0]); n++)
+		ctx->fs[n] = ((uint64_t)(0x3fd0 + n) << 48) |
```
jenswikl, line 60: Please use `SHIFT_U64()`

ta/os_test/include/riscv_fp_ctx.h
```diff
@@ -0,0 +1,243 @@
...
+
+	/*
+	 * frm = 2, round down, which differs from the reset value, and all
+	 * five accrued exception flags set. The flags are sticky and are
+	 * only ever cleared explicitly, so starting from all ones keeps the
+	 * comparison from tripping over a called function that happens to
+	 * do some arithmetic of its own, while a save or restore that drops
+	 * fcsr altogether still shows up.
+	 */
+	ctx->fcsr = (2 << 5) | 0x1f;
```
jenswikl, line 71: Magic numbers. Please use `SHIFT_*()`

ta/os_test/include/riscv_fp_ctx.h
```diff
@@ -0,0 +1,243 @@
...
+ * only fcsr differs, and -1 if the two contexts are identical.
+ */
+static inline int riscv_fp_ctx_diff(const struct riscv_fp_ctx *a,
+				    const struct riscv_fp_ctx *b)
+{
+	size_t n = 0;
+
+	for (n = 0; n < sizeof(a->fs) / sizeof(a->fs[0]); n++)
+		if (a->fs[n] != b->fs[n])
+			return (int)n;
```
jenswikl, line 85: Cast not needed.

ta/os_test/include/riscv_fp_ctx.h
```diff
@@ -0,0 +1,243 @@
...
+
+	if (a->fcsr != b->fcsr)
+		return 12;
```
jenswikl, line 88: Magic number

ta/os_test/include/riscv_fp_ctx.h
```diff
@@ -0,0 +1,243 @@
...
+ * in the C form.
+ *
+ * Stack frame of riscv_fp_ctx_roundtrip():
+ *	   0	ra
+ *	   8	s0
+ *	  16	caller's fcsr
+ *	  24	caller's fs0..fs11
+ */
+#ifdef RISCV_FP_CTX_IMPLEMENTATION
+_Static_assert(offsetof(struct riscv_fp_ctx, fs) == 0,
```
jenswikl, line 116: static_assert()
Can asm-defines be used to avoid hardcoding offsets?

ta/os_test/include/riscv_fp_ctx.h
```diff
@@ -0,0 +1,243 @@
...
+	       "riscv_fp_ctx layout out of sync with the assembly below");
+_Static_assert(offsetof(struct riscv_fp_ctx, fcsr) == 96,
+	       "riscv_fp_ctx layout out of sync with the assembly below");
+
+__asm__(
```
jenswikl, line 121: Please move this into a proper .S file.
For assembly files, we try to use a naming convention. riscv assembly files should end with `_rv.S`, `_rv32.S`, or `_rv64.S`, depending on what is accurate.

ta/os_test/include/riscv_fp_ctx.h
```diff
@@ -0,0 +1,243 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2026, RISCStar Solutions Limited
+ */
+
+#ifndef RISCV_FP_CTX_H
+#define RISCV_FP_CTX_H
+
+#if defined(__riscv) && defined(__riscv_flen) && __riscv_flen == 64
+#define RISCV_FP_CTX_SUPPORTED	1
```
jenswikl, line 10: Isn't this available in some `CFG_` variable already?

ta/os_test/os_test.c
```diff
@@ -1721,3 +1724,184 @@ TEE_Result ta_entry_asan_uaf(void)
...
+				     unsigned long (*fn)(void *), void *arg)
+{
+	struct riscv_fp_ctx expect = { };
+	struct riscv_fp_ctx got = { };
+	unsigned long res = 0;
+	int diff = 0;
+
+	riscv_fp_ctx_pattern(&expect, seed);
+
+	res = riscv_fp_ctx_roundtrip(&expect, &got, fn, arg);
```
jenswikl, line 1799: Does `riscv_fp_ctx_roundtrip()` actually return a value that is compatible with `TEE_Result`?

ta/os_test/os_test.c
```diff
@@ -1721,3 +1724,184 @@ TEE_Result ta_entry_asan_uaf(void)
...
+		EMSG("FP context: call failed: %#lx", res);
+		return (TEE_Result)res;
+	}
+
+	diff = riscv_fp_ctx_diff(&expect, &got);
+	if (diff < 0)
+		return TEE_SUCCESS;
+
+	params[1].value.a = diff;
+	if (diff == 12)
```
jenswikl, line 1810: Magic number

ta/os_test/include/riscv_fp_ctx.h
dave-patel-riscstar: Please can you share you thought behind your question?

ta/os_test/os_test.c
dave-patel-riscstar: res is TEE_Result.

ta/os_test/include/riscv_fp_ctx.h
dave-patel-riscstar: There's no CFG_ usable from the TA test code here. CFG_RISCV_FPU exists in the core/dev-kit conf.h, but the TA dev kit doesn't force-include conf.h and doesn't expose it as a make variable, so TA code only sees CFG_ flags its own sub.mk -Ds; CFG_TA_FLOAT_SUPPORT is available but architecture-neutral, and there's no CFG_ for the FP register width. So I dropped the RISCV_FP_CTX_SUPPORTED macro and now gate both the header and the TA test directly on #if defined(RV64) && defined(__riscv_flen) && __riscv_flen == 64; the dev-kit RV64 macro (same as #if defined(ARM64) already in this file) plus the D-extension / register-width check.

ta/os_test/include/riscv_fp_ctx.h
dave-patel-riscstar: Switched _Static_assert to static_assert. On asm-defines: the mk support exists in the TA dev kit, but gen-asm-defines.h isn't exported to TAs and no TA uses the mechanism, so wiring it up would mean vendoring optee_os build internals for two offsets. Instead I moved the offsets into riscv_fp_ctx.h (RISCV_FP_CTX_FS_OFF / RISCV_FP_CTX_FCSR_OFF, outside an __ASSEMBLER__ guard) and riscv_fp_ctx_rv64.S now includes the header and uses those macros — so the asm no longer hard-codes any struct offset, and the static_assert()s verify the struct matches them. Same effect as asm-defines, without the extra build machinery.

dave-patel-riscstar:
@jenswikl I have addressed your comments, pls check.

ta/os_test/include/riscv_fp_ctx.h
jenswikl: It was a half-rhetorical question, and I think you just answered that.
Please make them normal, non-inline functions.

ta/os_test/os_test.c
```diff
@@ -1721,3 +1723,185 @@ TEE_Result ta_entry_asan_uaf(void)
...
+ * Floating-point context switching.
+ *
+ * A TA runs with the floating-point unit disabled and is given it on the
+ * first floating-point instruction it executes, so every one of these
+ * sub-tests starts by trapping into the TEE and being handed a context.
+ * What each one then checks is that the context comes back intact across a
+ * different kind of excursion out of the TA.
+ */
+
+static unsigned long fp_call_syscall(void *arg __unused)
```
jenswikl: Please use `TEE_Result` return type for the functions that return values of that type.

ta/os_test/include/riscv_fp_ctx.h
jenswikl: Where's the `#if defined(ARM64)` in this file?

Do what you think is easiest to maintain (you might be expected to maintain this when things blow up). I'm only trying to nudge you in the right direction.

ta/os_test/os_test.c
jenswikl: Yes, but it has an `unsigned long` return type. Please fix.

ta/os_test/include/riscv_fp_ctx.h
dave-patel-riscstar: make sense; so I have dropped RV64 as it is already implied and there is no need to be explicit about it.

dave-patel-riscstar:
@jenswikl please check and confirm.

PR #829, opened 2026-09-01
regression: add 4020 to test discarded output

jenswikl:
Test case for https://github.com/OP-TEE/optee_os/pull/7966


host/xtest/regression_4000.c
```diff
@@ -7232,3 +7302,242 @@ static void xtest_tee_test_4019(ADBG_Case_t *c)
...
+							   &dlen);
+		}
+		if (!ADBG_EXPECT_TEEC_SUCCESS(c, res) ||
+		    !ADBG_EXPECT_COMPARE_UNSIGNED(c, text_size, ==, dlen) ||
+		    !ADBG_EXPECT_BUFFER(c, ciph_text, text_size, tmp_text,
+					dlen))
+			goto out;
+		if (auth_enc &&
+		    !ADBG_EXPECT_BUFFER(c, tag, sizeof(tag), tmp_tag, tlen))
+			goto out;
```
jforissier, line 7560: I would suggest to also check the returned tag length:

```suggestion
		if (auth_enc &&
		    (!ADBG_EXPECT_COMPARE_UNSIGNED(c, sizeof(tag), ==, tlen) ||
		     !ADBG_EXPECT_BUFFER(c, tag, sizeof(tag), tmp_tag, tlen)))
			goto out;
```

jforissier (commented):
With or without my suggestion:

host/xtest/regression_4000.c
jenswikl: That's already checked by `ADBG_EXPECT_BUFFER()`.

jforissier: Ah yes. Then it's fine.

anarchyj:
All subcases here use AES. The discard path in libutee depends on DISCARD_BUF_SIZE being a multiple of op->block_size and on tee_buffer_update() keeping the right residue for 8-byte blocks, and DES, DES3 and SM4 operations reach it too. Please add at least TEE_ALG_DES3_CBC_NOPAD and TEE_ALG_SM4_XTS with the same split points.

jenswikl:
Added TEE_ALG_DES3_CBC_NOPAD. SM4 has the same block size as AES, so I skipped that.

jenswikl:
Adding tests for TEE_AEDrcryptFinal().

PR #830, opened 2026-09-01
xtest: regression_6010: test GetNext after failed start

GSoJC234:
Verify that GetNext returns ITEM_NOT_FOUND after Start fails because the storage directory does not exist.

Add coverage for calling `TEE_GetNextPersistentObject()` after
`TEE_StartPersistentObjectEnumerator()` returns `TEE_ERROR_ITEM_NOT_FOUND`.

The expected result is `TEE_ERROR_ITEM_NOT_FOUND`, without a Secure Core abort.

Depends on: [optee_os PR URL](https://github.com/OP-TEE/optee_os/pull/7975)

Tested with:

    make -j2 CHECK_TESTS=xtest XTEST_ARGS=6010 check


jenswikl:

PR #831, opened 2026-09-02
xtest: add digest final short-buffer regression test

GSoJC234:
Add coverage for TEE_DigestDoFinal() after TEE_DigestExtract() with an insufficient output buffer on non-XOF algorithms.

Verify that it returns TEE_ERROR_SHORT_BUFFER and leaves the operation usable for a subsequent finalization, as required by GP TEE Internal Core API v1.3.1 §6.3.2.

### DIFF
--- a/host/xtest/regression_4000.c
+++ b/host/xtest/regression_4000.c
@@ -1502,6 +1502,15 @@ static void xtest_tee_test_4001(ADBG_Case_t *c)
 			ta_crypt_cmd_copy_operation(c, &session, op2, op1)))
 			goto out;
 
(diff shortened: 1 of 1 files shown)


GSoJC234:
Depends on: https://github.com/OP-TEE/optee_os/pull/7978

jforissier:

jforissier:
We would need your real name in the Author and Signed-off-by though, as per https://optee.readthedocs.io/en/latest/general/contribute.html#developer-certificate-of-origin. Thanks!

GSoJC234:
Updated the Author and Signed-off-by with my real name, and added the Reviewed-by tag. Thanks!

PR #832, opened 2026-09-02
xtest: AES known answer tests against the NIST SP 800-38A vectors

dave-patel-riscstar:
**Draft — companion to the optee_os Zvkned AES series (OP-TEE/optee_os#7980).**

The AES cipher cases in `ciph_cases[]` are all 128-bit, so the 192- and
256-bit key schedules are never exercised. That matters for an accelerated
AES, where the key length selects a different number of rounds and so a
different path through the block function.

Adds one test per mode against the NIST SP 800-38A appendix F vectors, for
all three key lengths, each split both on and off a block boundary so the
update path is covered buffered and unbuffered:

| Test | Coverage |
| --- | --- |
| 4020 | AES-ECB, F.1 |
| 4021 | AES-CBC, F.2 |
| 4022 | AES-CTR, F.5, plus a partial final block and two counter-carry cases |

The case macros emit both directions, so every vector is a known-answer test
encrypting and decrypting rather than a round-trip that could pass with both
halves wrong.

The CTR carry cases are the interesting ones: an accelerator that counts
several blocks ahead in vector registers naturally does the arithmetic on
the low 32 bits alone, and carry above that is a separate path none of the
SP 800-38A vectors reach, since their counters stay inside the low byte. One
case starts the counter at all-ones, one with the low 32 bits all-ones and a
non-trivial value above them.

The first commit is a pure refactor: test 4003's driver loop was written
against `ciph_cases[]` directly, so it is split into `xtest_cipher_cases()`
taking a table, leaving 4003 unchanged in behaviour and its table untouched.

Only `host/xtest/regression_4000.c` is touched, and the series applies
directly to master with no dependency on the RISC-V FP or vector test series.


jenswikl:
Please drop the riscv patches, and this can be handled independently of that other PR.

dave-patel-riscstar:
> Please drop the riscv patches, and this can be handled independently of that other PR.

ok @jenswikl, there are separate PR's for FP and V xtest.

One question do we need this PR in general, as there are existing xtest for AES ECB, CBC and CTR?

jenswikl:
> One question do we need this PR in general, as there are existing xtest for AES ECB, CBC and CTR?

I haven't looked closer at this PR, but I was under the impression that you needed more key sizes, etc.

dave-patel-riscstar:
> > One question do we need this PR in general, as there are existing xtest for AES ECB, CBC and CTR?
> 
> I haven't looked closer at this PR, but I was under the impression that you needed more key sizes, etc.

Yes, it covers different key size. I have updated the PR by removing FP and V as requested.

jenswikl:
How about adding all the new NIST cases to `ciph_cases`?
You can separate the added cases with comments in the array, like we already do there.

PR #834, opened 2026-09-04
xtest: add object usage intersection regression

GSoJC234:
Add a regression test for TEE_CopyObjectAttributes1(). Restrict an AES source object to ENCRYPT|DECRYPT and an uninitialized destination object to ENCRYPT, then copy the attributes and read back the destination usage.

The expected destination usage is ENCRYPT. Affected implementations instead return ENCRYPT|DECRYPT by replacing the destination usage mask.

### DIFF
--- a/ta/crypt/cryp_taf.c
+++ b/ta/crypt/cryp_taf.c
@@ -511,6 +511,39 @@ TEE_Result ta_entry_copy_object_attributes(uint32_t param_type,
 	return TEE_CopyObjectAttributes1(dst, src);
 }
 
+TEE_Result ta_entry_restrict_object_usage(uint32_t param_type,
+					 TEE_Param params[4])
+{
+	TEE_ObjectHandle object = obj_handle_lookup(params[0].value.a);
+
+	ASSERT_PARAM_TYPE(TEE_PARAM_TYPES
+			  (TEE_PARAM_TYPE_VALUE_INPUT, TEE_PARAM_TYPE_NONE,
+			   TEE_PARAM_TYPE_NONE, TEE_PARAM_TYPE_NONE));
+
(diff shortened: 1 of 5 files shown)


GSoJC234:
Depends on: https://github.com/OP-TEE/optee_os/pull/7989

ta/crypt/cryp_taf.c
```diff
@@ -511,6 +511,65 @@ TEE_Result ta_entry_copy_object_attributes(uint32_t param_type,
 	return TEE_CopyObjectAttributes1(dst, src);
 }
 
+TEE_Result ta_entry_copy_object_attributes_usage(uint32_t param_type,
```
jenswikl: Can't `ta_entry_copy_object_attributes()` be used instead?
That is, keeping only thin wrappers here while xtest does the heavy lifting?

GSoJC234: Yes, that makes sense. I’ve updated the test to reuse the existing TA_CRYPT_CMD_COPY_OBJECT_ATTRIBUTES command and its ta_entry_copy_object_attributes() handler, and removed the test-specific ta_entry_copy_object_attributes_usage() function.

I added only two thin, generic TA wrappers for restricting and inspecting object usage, keeping all test-specific setup and assertions in xtest. Since these wrappers are not tied to this regression, they can also be reused by future object-usage tests.

I pushed this as a follow-up commit for easier review and will squash it before merge. Thanks.

jenswikl:
Looks good. Please squash and add:

GSoJC234:
Thanks. I've squashed the commits and added the Acked-by tag as requested.

jenswikl:
Please rebase to resolve the conflict.

PR #835, opened 2026-09-09
xtest: add truncate no-space test

GSoJC234:
Add an opt-in regression test that fills the REE file system before extending a persistent object. Verify that TEE_TruncateObjectData() returns TEE_ERROR_STORAGE_NO_SPACE and that the TA session remains alive.

Use XTEST_REE_FS_ENOSPC_DIR to specify the file system parent path used by tee-supplicant.

### DIFF
--- a/host/xtest/regression_6000.c
+++ b/host/xtest/regression_6000.c
@@ -6,6 +6,10 @@
 #include <string.h>
 #include <stdio.h>
 #include <pthread.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <stdlib.h>
+#include <unistd.h>
 
 #include <adbg.h>
 #include <xtest_test.h>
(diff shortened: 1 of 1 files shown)


GSoJC234:
Depends on: https://github.com/OP-TEE/optee_os/pull/7997

host/xtest/regression_6000.c
```diff
@@ -2318,3 +2322,84 @@ static void xtest_tee_test_6021_single(ADBG_Case_t *c, uint32_t storage_id)
 DEFINE_TEST_MULTIPLE_STORAGE_IDS(xtest_tee_test_6021)
 ADBG_CASE_DEFINE(regression, 6021, xtest_tee_test_6021,
 		 "Modify and check persistent object usage");
+
+static bool fill_ree_fs(const char *dir, char *path, size_t path_size)
```
jenswikl, line 2326: No, we can't fill REE FS. That's too hostile. I think we can skip this test altogether. It seems too hard to implement compared to the upside.

host/xtest/regression_6000.c
```diff
@@ -2318,3 +2322,84 @@ static void xtest_tee_test_6021_single(ADBG_Case_t *c, uint32_t storage_id)
...
+{
+	static const uint8_t object_id[] = "truncate-enospc";
+	const char *ree_fs_dir = getenv("XTEST_REE_FS_ENOSPC_DIR");
+	TEEC_Result res = TEEC_ERROR_GENERIC;
+	char filler_path[512] = { };
+	TEEC_Session sess = { };
+	uint32_t orig = 0;
+	uint32_t obj = 0;
+
+	if (!ree_fs_dir || !*ree_fs_dir) {
```
anarchyj: Without `XTEST_REE_FS_ENOSPC_DIR` the case returns early and counts as passed, and it covers `TEE_STORAGE_PRIVATE_REE` only. Please skip instead of pass when the variable is absent, and add the RPMB storage id once the RPMB truncate reports no space.

anarchyj: Without `XTEST_REE_FS_ENOSPC_DIR` the case returns early and counts as passed, and it covers `TEE_STORAGE_PRIVATE_REE` only. Please skip instead of pass when the variable is absent, and add the RPMB storage id once the RPMB truncate reports no space.

PR #837, opened 2026-09-09
Test/storage write position limit

GSoJC234:
regression: test writes beyond TEE_DATA_MAX_POSITION

Move an empty RPMB object's data position to
TEE_DATA_MAX_POSITION - 1 and attempt a two-byte write. Verify that
TEE_ERROR_OVERFLOW is returned without changing the position or size.

Use RPMB to keep the failure path bounded. Reopen the storage TA for
cleanup if the affected implementation panics.

### DIFF
--- a/host/xtest/regression_6000.c
+++ b/host/xtest/regression_6000.c
@@ -2318,3 +2318,87 @@ exit:
 DEFINE_TEST_MULTIPLE_STORAGE_IDS(xtest_tee_test_6021)
 ADBG_CASE_DEFINE(regression, 6021, xtest_tee_test_6021,
 		 "Modify and check persistent object usage");
+
+static void xtest_tee_test_6022(ADBG_Case_t *c)
+{
+	const uint32_t flags = TEE_DATA_FLAG_ACCESS_WRITE |
+			       TEE_DATA_FLAG_ACCESS_WRITE_META;
+	char object_id[] = "xtest_write_position_limit";
+	uint8_t data[2] = { };
+	TEE_ObjectInfo obj_info = { };
+	TEEC_Result res = TEEC_ERROR_GENERIC;
+	TEEC_Session sess = { };
+	uint32_t orig = 0;
+	uint32_t obj = 0;
+
+	if (!is_storage_available(TEE_STORAGE_PRIVATE_RPMB)) {
+		Do_ADBG_Log("RPMB storage not available, skipping test 6022");
+		return;
+	}
+
+	res = xtest_teec_open_session(&sess, &storage_ta_uuid, NULL, &orig);
+	if (!ADBG_EXPECT_TEEC_SUCCESS(c, res))
+		return;
+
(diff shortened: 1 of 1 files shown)


GSoJC234:
Depends on: https://github.com/OP-TEE/optee_os/pull/7999

jenswikl:
It's good to add regression tests when fixing certain bugs. But I think we need to organize them in some way or piggyback on existing tests. This is a bit much to add for such a trivial case, and why wouldn't we want to test it on TEE FS?

It's fine to add fixes without an updated test. Perhaps it's better to make a list of all the things we want to improve the tests with later. So when we have the big picture, we can design the tests properly. Some things certainly need new tests immediately, so it's a bit of a judgement call.

By the way, this test fails for me even with OP-TEE/optee_os#7999 added.

host/xtest/regression_6000.c
```diff
@@ -2318,3 +2318,87 @@ static void xtest_tee_test_6021_single(ADBG_Case_t *c, uint32_t storage_id)
...
+	res = fs_create(&sess, object_id, sizeof(object_id),
+			flags | TEE_DATA_FLAG_OVERWRITE, 0, NULL, 0, &obj,
+			TEE_STORAGE_PRIVATE_RPMB);
+	if (!ADBG_EXPECT_TEEC_SUCCESS(c, res))
+		goto exit;
+
+	if (!ADBG_EXPECT_TEEC_SUCCESS(c, res))
+		goto exit;
+
+	res = fs_seek(&sess, obj, INT32_MAX, TEE_DATA_SEEK_CUR);
```
anarchyj: One `TEE_DATA_SEEK_CUR` of `INT32_MAX` from 0 lands at `0x7FFFFFFF`, so the comparison with `TEE_DATA_MAX_POSITION - 1` fails and the case exits before `fs_write()`, with and without optee_os#7999. It takes two seeks to reach `0xFFFFFFFE`, and the duplicated success check after `fs_create()` looks like where the second one went.

anarchyj: One `TEE_DATA_SEEK_CUR` of `INT32_MAX` from position 0 lands at `0x7FFFFFFF`, so the comparison with `TEE_DATA_MAX_POSITION - 1` fails and the case exits before `fs_write()`, with and without optee_os#7999. Two seeks are needed to reach `0xFFFFFFFE`, and the repeated success check after `fs_create()` suggests the second one was dropped.

anarchyj, line 2352: One `TEE_DATA_SEEK_CUR` of `INT32_MAX` from position 0 lands at `0x7FFFFFFF`, so the comparison with `TEE_DATA_MAX_POSITION - 1` fails and the case exits before `fs_write()`, with and without optee_os#7999. Two seeks are needed to reach `0xFFFFFFFE`, and the repeated success check after `fs_create()` suggests the second one was dropped.

PR #838, opened 2026-09-09
xtest: pkcs11: keep the X.509 subject and issuer names const

ricardosalveti:
OpenSSL 4.0 changed X509_get_subject_name() and X509_get_issuer_name() to return a const X509_NAME pointer. The certificate test assigns the results to non-const pointers, and since xtest builds with -Werror the whole host build fails against OpenSSL 4.0:

  pkcs11_1000.c:7879:27: error: assignment discards 'const' qualifier
  from pointer target type [-Werror=discarded-qualifiers]

Declare both names const. They are only passed to i2d_X509_NAME(), which has taken a const pointer since OpenSSL 1.1.0, and to ADBG_EXPECT_NOT_NULL(), which takes a const void pointer, so the change builds unchanged against OpenSSL 3.x as well.


ricardosalveti:
Failed tests to be fixed by https://github.com/OP-TEE/optee_test/pull/839

jenswikl:

jenswikl:
Please apply the tag.

ricardosalveti:
Tag applied!

